PR #38, opened 2020-09-25
Merge with Ron's working-directory lispcore

masinter:


masinter:
i copied from a newer dropbox into master, instead of following this "keep-old-versions"

PR #44, opened 2020-10-12
Add `-screensize` arg to run-medley

### LINKED ISSUE #42: run-medley doesn't set -sc for the screen resolution
Without setting -sc, the screen defaults to the full size of my display on Kubuntu 20.04. This is not good since I have multiple monitors and it sets the size if the Medley screen to the *overall* size of *all* monitors put together. There doesn't seem to be a way to pass in -sc to the script.

AbeJellinek:
Resolves #42


masinter (approved):

masinter:
I agree

PR #45, opened 2020-10-12
Add existing fg and bg options to run-medley

AbeJellinek:
Enables Interlisp/maiko#16 by pulling these two options into the run script.


masinter:
i think this is handled by passthru. I updated the Maiko corresponding PR but it doesn't seem to have any effect.

PR #118, opened 2020-12-30
Ignore versions + macinterface + some cleanup of internal/test

masinter:
this was an experiment (unsuccessful) trying to work on the same branch.
I'm not sure how to tease out the separate commits though, so let's just not ever do this again.


rmkaplan (approved):

masinter:
just this one last time -- until we have better git tools

PR #123, opened 2021-01-01
Remove Maiko submodule.

AbeJellinek:
I was under the impression that Git submodules had improved; they have not. Submodules still require manual updates, effort on the part of the user to initialize them, and constant maintenance. There's such a thing as a "subtree," which might better fit our needs, but I'm going to scrap the whole idea for now because the advantages over cloning both repositories separately (or having an init script that does it for us) aren't clear.


nbriggs (approved):
OK

PR #128, opened 2021-01-04
Larry teditmenu

### LINKED ISSUE #120: odd black blobs in TEdit expanded menu
image

masinter:
fixes #120


nbriggs:
You didn't try writing this with the EOL convention set to LF?  The older version of the file looks as though that was the case, while the new one comes out as one big long line in all the replaced parts.

masinter:
@nbriggs I don't know any way currently to set the EOL convention; I think that was something Ron was imagining, e.g., as a simpler case of "output to Unicode". I suppose I could have used tr in Unix instead of editing the function, but then everything would appear as a blob.

rmkaplan:
The EOL convention is a property of the stream, so it can be set on opening (with PARAMETER EOL).  It’s also a property of EXTERNALFORMAT, which I think also changes the stream if given as an opening paramater.

But I think the first thing to do is to make the character reading primitives ignore that property and convert lf and crl to cr.  So if files happen to be written out or get important from different environments, the internal behavior would be correct.

We could also do it incrementally, with a safety net, by adding a 4th value to the 2-bit field with an interpretation of ANY, and then do the normalization only in that case.

That property may still be useful for writing, for users who tend to operate soley within the confines of a local operating system.  But for our Unix/Mac/git interactions we would eventually convert the defaults to LF.

One hurdle (if the relevant stuff is on LLREAD and not FILEIO) is that the last time I tried to patch LLREAD, a year ago, I was not able to produce a compiled version (either LCOM or DFASL) that could be loaded without a bad-compiled-code error.  I’ll try again, if it fails I'll stick in a bug report.

nbriggs:
I believe that the current PFCOPYBYTES (I think that's what it is called) carefully converts any incoming EOL convention to CR, but then on output I would have expected either the default EOL character(s) or the one(s) specified as a parameter to OPENSTREAM to have been printed.  I believe it calls (TERPRI) doesn't it?

rmkaplan:
PFCOPYBYTES has other issues (e.g. the multiple-value return).  Also, PFCOPYBYTES is overwritten by another version in a library package that I think does reformatting for files that were not prettily dumped.  

The primitive macro is probably \NSIN…which we would also eventually like to get rid of.

nbriggs:
I tried replacing OPENSTREAM with a version that cons'd (EOL LF) onto the parameters and then called the original openstream, and then did a (LOAD 'TEDITMENU 'PROP), CONN {dsk}<tmp>,  (MAKEFILE 'TEDITMENU 'NEW) and it made it with the LF convention.  That wouldn't have gone through PFCOPYBYTES so would have avoided any issues there.

rmkaplan:
That’s the expected behavior, which is good.  But the input side needs some work.  If I did the ANY extension, then we could experiment by having (EOL ANY) on an input stream and (EOL LF) on output streams.

nbriggs:
I was looking at PFCOPYBYTES on PRINTFN.  There are packages that do things like
(CHANGENAME 'FB.FASTSEE.ONEFILE 'PFCOPYBYTES 'PFI.MAYBE.SEE.PRETTY) but I haven't seen anyone that redefines PFCOPYBYTES yet.

masinter:
This PR is blocked pending approval. The discussion of how to do things better is interesting but this fixes an annoying bug.

masinter:
see Issue #2

rmkaplan (approved):

rmkaplan:
That sounds familiar, maybe that’s what I saw.  Something that BVM had done.  (I tried to find where that was, but FINDCALLERS got into its unknown-package problems and after a sufficient number of RETFROM’s I got an arrayblock GC URAID, haven’t see one of those in quite awhile.)

The reason that PFCOPYBYTES has all that opencoded stuff is that it has to keep track of the byte positions while it is trafficking in characters that may be multi-byte.  (The variable is named #CHARS and the comment talks about chars, but I think it is really bytes.)

nbriggs:
I don't have all my site/personal init files set up properly, so when I want to find something I frequently just open sources/system.hash in Emacs and search for the symbol... the next bit of text will tell me where it is.

PR #129, opened 2021-01-05
add MEDLEY-FIX-DIRS to fix dates of (most of) medley repo

masinter:
MEDLY-FIX-DIRS can and should be done on any checkout / clone / pull

Right now it just fixes file dates, but fixing top-two-versions-equal (and "if only version 1 then rename version 1 to version of filecreated as") can be added later.


masinter:
https://github.com/Interlisp/medley/discussions/102 
has part of the rationale

rmkaplan (changes requested):
You have a typo in the MEDLEY-FIX-DIRS directory list:  lispuseers → lispusers

I see that all of the functions on the file have been changed, and it seems that the reason is a difference in capitalization in the clisp words (if/IF etc.).  So how did you do the edit?  Did you load all of the symbols and then do a makefile of the whole thing?  As opposed to a LOAD of the compile file and a loadfrom the source?

Also, do we have an issue about different settings of whatever flag controls case-changes for clisp?

masinter:
I don't have a personal init, so whatever flg that turns off CLisp lowercasing  of iteration keywords is likely set in your personal greeting. I did `. ANALYZE ON ANY  IN FILELST or some such; Masterscope calls to dwimify to get the translation is likely what changed it. I did use COMPARESOURCES to check the file changes.

nbriggs (commented):
Why does MEDLEY-FIX-DIRS have a mixture of syntaxes for directories?

(RPAQQ MEDLEY-FIX-DIRS ("sources" "library" "lispusers" "internal/library" "greetfiles"
                                  "docs>Documentation Tools" "cltl2" "clos" "makesysout"))

nbriggs:
IL:LCASEFLG is set in all our sysouts, including the Venue lisp.sysout and full.sysout, so it's not surprising if things transition from uppercase to lowercase if touched by CLISPIFY, is it?

masinter:
As long as we're stepping through all of the files in the repo, i want to refactor:
if given a directory, step through all the files matching `*.*;` (highest version)
if given a file, enumerate `file.exe;*` (all versions)
1. fix the dates on all file versions that are lisp files
2. if there is more than one version and highest version == the next highest version (same length and date and content), then delete the highest version
3. otherwise (there was only one version) if there was only one version, and it was version one, and the file name extension and last part of path (library/lispusers etc basically strip MEDLEYDIR) match the CREATED-AS info, then rename file.ext;1 to file.ext;N where N is taken from the CREATED-AS.   

I think we can set up GitHub to automatically write a .NEEDS-FIXING (gitignored) file whenever anything is fetched.
Maybe run-medley could check it?

masinter:
> IL:LCASEFLG is set in all our sysouts, including the Venue lisp.sysout and full.sysout, so it's not surprising if things transition from uppercase to lowercase if touched by CLISPIFY, is it?
technically CLISPIFY also turns (PLUS A B) into A+B, it's DWIMIFY that lower cases keywords.
The results aren't as pleasing if you then print with the xcl readable, which turns `(for X in Y do Z)` into `(|for| X |in| Y |do| Z)`.
As long as you stay in Interlisp it's not bad.
Maybe raise a new issue? I think handling this kind of stuff might go beyond fixing bugs.

rmkaplan:
I think the controlling variable is LCASEFLG.  It comes in as T,  it seems to control whether DWIMIFY lower cases.  I set it to NIL in my INIT, since I don’t want gratuitous changes (even if that propagates nonuniformity) on functions that have come in from different authors.

If its value is T,  the effect would be to create spurious diffs both in git and COMPARESOURCES, if a function previously existed with upper case and anything happened (like running masterscope as you did in a way that marked the function as changed so it got needlesssly dumped).

LCASEFLG is not indexed in the manual, as near as I can tell.

There are other variables that might also introduce troublesome changes:  CLISPIFYPRETTYFLG and CLISPIFTRANFLG.  Those are set iin SYSEDIT, while LCASEFLG is not included there.  But SYSEDIT doesn’t need to be loaded for typical work on library and lispusers files.

I propose that we set up our site greeting file to turn off gratuitous transformations.

masinter:
> Why does MEDLEY-FIX-DIRS have a mixture of syntaxes for directories?
> 
> (RPAQQ MEDLEY-FIX-DIRS ("sources" "library" "lispusers" "internal/library" "greetfiles"
> "docs>Documentation Tools" "cltl2" "clos" "makesysout"))
Because I was focused on the space and trying to come up with a reason why it bothered me. I'd move it to internal/doctools or even merge into internal/library if I had a credible reason. 

More to the point, though, I'm hoping next pass to get rid of the explicit list and just look in every file in the medley repo (excluding fonts)

rmkaplan:
Actually, I think there may be a few historical lurkers from earlier file transfers that I did not notice during my cleanup passes, and those may not be the latest versions.  So it may be worth comparing all earlier sequential pairs of versions for byte equality and fixing those up too.  That would clean up the directory if things accidentally get out of step at some point in the future.

rmkaplan:
An intermediate position would be to have Masterscope ANALYZE rebind the variables to turn off all of the gratuitous change controllers. For sure the act of analyzing should not change functions, no matter what other policy we want to enforce.

PR #139, opened 2021-01-08
update MAKE-EXPORTS-ALL and run it

masinter:


rmkaplan (approved):

nbriggs (changes requested):
The deletion of the ISSUE_TEMPLATE and other files don't look as though they belong here.

Issue #140, opened 2021-01-09
DIRECTORY problems


masinter:
I'm not quite sure what the problem is, but I'm having
trouble calling IL:DIRECTORY or CL:DIRECTORY to get
a list of source files only in MEDLEYDIR/library

When I do (DIRECTORY (CONCAT (MEDLEYDIR "library") "*.;"))
i get all versions and all extensions in subdirectories.

This seems to also happen in lisp.venuesysout so lll

nbriggs:
That doesn't happen for me in the full.venuesysout (using IL:DIRECTORY).

nbriggs:
It *does* include the directories, though, with no files underneath them (e.g., LAFITE>) which I think is correct since the directory has no extension and no version.

masinter:
LISP.VENUE has the problem
![Screenshot 2021-01-08 193801](https://user-images.githubusercontent.com/1116587/104082205-27174b80-51e9-11eb-83b0-02a8be112bd7.png)
full.venue doesn't
![FULL VENUE](https://user-images.githubusercontent.com/1116587/104082486-65156f00-51eb-11eb-9d99-9d8111671f3a.png)

masinter:
I've narrowed this down to the difference in code (do `INSPECTCODE(\UFS.NEXTFILEFN)`  between lisp.venuesysout dated 10 April 2000 (which has the problem) and full.venuesysout dated 4 Nov 2003 (which doesn't).
\UFS.NEXTFILEFN is in the file UFS.  `pl UFS` shows the same filedates of 29-MAR-95, 

`find . -type f -name "UFS*" -mtime +7400 -exec ls -l {} \;`

nbriggs:
I suspect it's the difference between UFS.~3~ and UFS.~4~

masinter:
the file  UFSNFS (which is not in the medley repo, but *is* in /history/2020s/202011-lisp/library/UFSNFS)
contains an earlier version of \UFS.NEXTFILEFN (and a few other  \UFS functions).
It looks like UFS was edited by jds in 93 to pay attention to the SUBGENERATORS field of GENFILESTATE.

nbriggs:
UFSNFS is in the venue full.sysout but not in the venue lisp.sysout.   This means the process of vetting files that were removed from the main repository was incomplete/incorrect and needs to be done again to ensure that we do not lose source for code that we are running.

masinter:
After diving into the code and getting lost, I've come to some conclusions:
[...]

PR #141, opened 2021-01-09
Added ~= selector for COMPAREDIRECTORIES to exclude byte-equal files.…

rmkaplan:
…  Added DIR argument to MEDLEY-FIX-DIRS


masinter (changes requested):
I can't quite tell but it looks like you didn't do a 'git pull' before editing? At least when I look at the COMPAREDIRECTORIES file in the Comparedirectories branch, you started with previous date 31-Oct-2020 instead of mine of 4-Jan-2021

masinter:
also, I want to hold off on changes to MEDLEY-FIX-DIRS until we understand why DIRECTORY("library/*.;") from MEDLEYDIR sometimes picks up all the files in the lafite sub-directory (including all LCOMs and DFASLs all versions. See #140

rmkaplan:
This has nothing to do with MEDLEY-FIX-DIRS, it’s a problem with the directory enumeration that I thought had been reported as another issue.  It’s a serious nonfeature for DIR, FILDIR, FB, etc.  I looked at it a year or so ago but couldn’t figure out how to fix it.  I think there is a DEPTH parameter that the lower-level code doesn’t respect.  I didn’t notice that maybe the behavior was different in different sysouts, which maybe would point to a place to look.

rmkaplan:
That may well be.  The strategy that we talked about on the phone call is really pretty much unworkable, at least as I understood it.

What I understood was that I should continue to work in my local-dev environment, as before, but at least try to keep that synchronized with the clone of the Git Master, which gets updated from the respository by pulling.

With changed files copied from Master to local-dev, changes gete made in local-dev.  Then to get those changes back into the repository, I have to create a new branch, switch to that branch, copy the changed files to the branch, push the branch, ask for a pull request,wait for somebody to approve, eventually do a merge, switch back to master.  And update the local-dev in case something else changed in the meantime.

That’s a lot of overhead for simple updates, like adding an extra selector in COMPAREDIRECTORIES, and even for more intricate updates.

I have been trying to write some automation tools for this (which motivated the new selector), but it is still a disaster. The wrong files are undoubtedly going to be moved (or not moved) to the wrong place at the wrong time.  And it obviously takes the joy out of development.

Maybe I misunderstood the proposal.  If so, please tell me what to do to simplify this.  Maybe fixing the version problem will make the need for some of this to go away.

If not, I think I will stop trying to maintain all of these dependencies in a timely way.  I will at least try to make sure that I copy Master updates into my local-dev environment, so that I don’t get too out of step with whatever else might be happening.  But other than that I will just work in my own world, and not bother with trying to get all the branching/pushing/requesting/merging to do the right thing with the right files in an incremental process.

From time to time (once a week, once a month?) I can invoke this or another heavy-weight procedure to make my changes available for others to examine and decide whether to merge them in.

masinter:
I thought we were engaged in an experiment on how to integrate GitHub into the ordinary Interlisp way.
I know things are awkward now and the edges are rough. I'm sorry to be the source of some of the confusion, but I'm just learning my way around it myself. I've been looking for a github consultant
    https://support.github.com/ticket/personal/0/971243 
"My open source project needs a github consultant #971243" scroll to the bottom:
> We have several problems with GitHub and could use some brief consulting (perhaps a couple of hours total) from someone FLUENT in github and its various features and extensions, custom diff tools, automating releases. (Some of the topics are in Discussions under https://github.com/Interlisp/medley/discussions/categories/github-use).

> I looked in Fiverr and CodeMentor but no success. Do you have any suggestions about where to look?

but I got a pretty useless (ai?) reply


There's a lot of controversy in other populations (or at least in the IETF)

masinter (approved):

PR #149, opened 2021-01-14
Add Larry's CR->LF converter and an install script

AbeJellinek:
Users can run scripts/install-diff-filter.sh to automatically update their Git config with a rule that translates CR to LF when diffing a file with an all-uppercase name. The conversion is only made in memory - this won't cause Git to start messing with our files on disk - and the rule is only set for the Medley repo.


nbriggs (commented):
I'd prefer to not tie it to /bin/bash (I'm happier with /bin/sh).  This is what I've used for locating the path...
```
SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
```

masinter:
I don't know how to propose updates using git so I'll send them to you by email.

masinter:
Fixed per @nbriggs review comments in #153

PR #151, opened 2021-01-15
dir and fb behave badly when filing.enumeration.depth is T; change it

masinter:
simplest fix to avoid surprises (me, @rmkaplan @pmcjones ). See #140 for the wild goose chase.


rmkaplan (approved):
Seems like the change is just in the simple-init, which is good.  But, should the default setting be changed in the system source file?

masinter:
I dont know what system file defines it to initially. I suspect was originally added as part of the XNS Filing support and later added as a more general feature in UFS (Unix File System).  the addition was ugly and incomplete. The right "default" depends on guessing correctly what "most" users would prefer based on how the kinds of file structures they are likely to encounter.

PR #153, opened 2021-01-17
Medley utilities

masinter:
one file (MEDLEY-FIX-DIRS) has utilities to make EXPORTS.ALL, WHEREIS.HASH, fix repository hard links and file dates.


masinter:
although this PR seems to have multiple changes, they seemed to interlink

AbeJellinek (approved):
This looks good to me. Thanks for making the `/bin/sh` change.

rmkaplan (approved):

PR #156, opened 2021-01-19
move palatino fonts down into c0

### LINKED ISSUE #144: FONTCREATE fails on font in FONTSAVAILABLE
even on case-insensitive file systems.

The weird problem I’m having with them is that 
(FONTSAVAILABLE ‘PALATINO 9 ‘MRR 0 ‘DISPLAY T) returns the expected
((PALATINO 9 0 (MEDIUM REGULAR REGULAR) DISPLAY))
but
(FONTCREATE ‘PALATINO 9 ‘MRR 0 ‘DISPLAY ) returns the NIL instead of a font descriptor, and therefore none of the TEDIT files with that font will load.

_Originally posted by @Anzus in https://github.com/Interlisp/medley/issues/142#issuecomment-757409414_

masinter:
fixes #144


nbriggs (approved):
Looks good to me.

PR #161, opened 2021-01-21
Update instructions for building and running & coordinate with Wiki

masinter:
I centralized and updated the various places where it described how to build and run Medley (and Blake's medley-intro) to the Wiki.
I'll update the pointers from the web site.


waywardmonkeys (commented):
I'm unsure here. If and as things change, then having the directions in the wiki means that they're out of sync with the source code and versioned differently / separately.

This is an issue for Maiko as well, but more so of an issue here, I think.

masinter:
I moved the actual build instructions back to the readme, left the wiki pages which now point back to the READMEs.

PR #171, opened 2021-02-02
UTF-8 wasn't interpreting EOLCONVENTION

rmkaplan:
Fixed UNICODE character fns to interpret the stream's EOLCONVENTION


rmkaplan:
Another attempt at doing "the right thing" in the git desktop.

masinter (approved):
I just looked at the source code change, and verified that UNICODE.LCOM seemed to be compiled on the same version of UNICODE.    Made me think about how to automate testing of "is this patch installed?"  More in the "testing" discussion.

PR #174, opened 2021-02-07
move loadup scripts + part of initial phase of cleanup

masinter:
@rmkaplan I'm moving things around to clean up and normalize. There are two Lisp files loadup-full and loadup-lisp which go with the scripts in the scripts folder. no more makesysout folder. I didn't get rid of patches yet but I'm close.

From looking at the diff of the dribble, it seems like you have a newer MACINTERFACE that isn't checked in.


rmkaplan:
If you get the makeinit going, then you shouldn’t need the SYNCLISPFILES.  But you still will need a place for the MAKEFULLSYSOUT.  Patches shouldn’t be necessary if we can always install updates in the main files and redo (automatically?) the loadups (including makeinit for updates to early-loadup files).

I’ll check on the status of the MACINTERFACE.  I was working on the problem of identifying clicks in the corners and the title bar with collections of attached windows, starting with the filebrowser.  I made some progress in that version, so that it is safe to upload, even though it is still incomplete.  I’m surprised that it wasn’t checked in as part of the Unicode check in.  But, I still have trouble coordinating the Medley vs. git models of development.

Issue #162, opened 2021-01-21
commit 6f7ec059bc (Remove dead stores) breaks -DSTACKCHECK


nbriggs:
```
clang -m32 -std=gnu99 -fno-strict-aliasing -fwrapv -c -O2 -g  -DMACOSX -DOLD_CURSOR -DBYTESWAP -DLOGINT  -I/opt/X11/include -DXWINDOW -DNOPIXRECT -DRELEASE=351 -DSTACKCHECK -DFSBCHECK -DPCTRACE  -I../inc/ -I../include/ ../src/llstk.c  -o ../darwin.386-x/llstk.o
../src/llstk.c:340:7: error: expected expression
      StackWord *orig68k = scanptr68k;
      ^
../src/llstk.c:349:27: error: use of undeclared identifier 'orig68k'
        if (scanptr68k != orig68k) {
                          ^
../src/llstk.c:354:64: error: use of undeclared identifier 'orig68k'
        if (((Bframe *)scanptr68k)->ivar != StkOffset_from_68K(orig68k)) {
                                                               ^
3 errors generated.
make: *** [../darwin.386-x/llstk.o] Error 1
```

nbriggs:
It leaves a declaration in the middle of code, not at the start of a `{ }` block.  Wrapping the code as `default: { code }` is a potential fix for this.

Issue #36, opened 2020-09-21
get MAKEINIT to run again (so we can build system from sources)


masinter:
The Interlisp tool set helped with some tools for allowing you to build systems from lisp sources, but it didn't enforce it, and we've seen instances #31 where it isn't true. 

I imagine in many cases it was set up ok, but in the end we have to check it out file by file.

@rmkaplan  has some cases where compiled and source files don't match.

 think you have to weigh the risk between
-- the .dfasl is from the wrong version of the file, and it's broken
-- the dfasl is fine but recompiling from the right source will make things worse, because
- a bug in the compiler (due to emulator or needing some declarations or other compile-time environment setup problems)

I don't think those things can be resolved quickly. We have sysouts that work and are pretty stable.
we have compiled files that work. let's put out a release of what we have and warn people and ask for help testing.
I'd like to get to the point where a full sysout can be rebuilt from sources (compilling every file from sources automatically (except having to have a 'starter' sysout to make an init to compile the compiler) but we're not there.

masinter:
idea: port MAKEINIT to Common Lisp. Then you'd just need a valid common lisp to remake the system from sources

rmkaplan:
I don’t think the problem is not having an environment to make the init, the problem is that the code we have is making an init that isn’t quite right.

masinter:
what is the process for making an init? I only vaguely remember.

Kirtai:
> idea: port MAKEINIT to Common Lisp. Then you'd just need a valid common lisp to remake the system from sources

This would also make Medley [bootstrappable](http://bootstrappable.org/) which would be a nice feature in itself.

rmkaplan:
I think porting MAKEINIT would be a very big deal.

As long as we have a running Medley sysout, we can use it to make makeinits for new systems (if we can figure out why what we have now or the procedure that we tried when we last looked at this doesn’t produce a runnable sysout).

In principle, Medley bootstraps itself.

Kirtai:
It would also help separate the runtime tooling from the build time tooling. IIRC, I believe this was part of the motivation to make SBCL buildable from other lisp implementations.

masinter:
Run time *is* build time in Interlisp, More or less on purpose.
If you run into a program error, you can stop and fix it and keep going.
Separating them isn't a project goal.

I changed the issue title to reflect the original intent

Kirtai:
[...]

PR #187, opened 2021-02-14
Making loadups from "scratch"

masinter:
Well, you need a working medley and possibly more. I tried to do it in stages except for the final push.
You also need a Linux x86 ldeinit which you make with `makeright init`.
It will reuse your `~/lisp.virtualmem`.
* scripts/loadup-init.sh   -- makes a init.dlinit  
* scripts/loadup-lisp-from-init.sh  -- makes a lisp.sysout
* scripts/loadup-full.sh -- makes a full sysout
* run-medley points a the new names


masinter:
this goes a long way toward fixing #36,  which I think I will close and open a new issue for the cleanup work

nbriggs:
What happens if you take the new lisp.sysout you've built with this process and use that as the starting point for building another new lisp.sysout?

masinter:
I’m about 12 generations deep

--
LarryMasinter.net

masinter:
so i'd like some advice on strategy for how to break this apart.
What I was trying to do was to untangle the nest of changes such that each commit was as small as I could make and still leave the state after the commit was "better" in some way.
So if I wanted to move or rename a file, I had to drag along the changes that referenced that file. Moving a definition required committing a change that affects lots of files.

Git Merge and comparison -- the whole git philosophy -- doesn't work with lisp.
At all, independent of Common Lisp vs. Interlisp. It works OK with C if you're really careful.
A set of git extensions that worked better with Lisp (in general, as well as with Medley Interlisp) would be a worthy project.

masinter:
I split loadup-lisp-from-init into two parts
* `./scripts/loadup-mid-from-init.sh`
* `./scripts/loadup-lisp-from-mid.sh`

Only the first needs ldeinit.

`./scripts/loadup-all.sh` will do the 4 steps in order. Timing:

* Raspberry Pi4: 1 minute 6 seconds
* Intel (5 years old): 12.4 seconds
* mac mini M1: 7.9 seconds

nbriggs:
These would work better if you just require/assume that "ldeinit" is on the search path...
```
prog=../maiko/linux.x86_64/ldeinit
if [ ! -x $prog ] ; then
    prog=../maiko/darwin.aarch64/ldeinit
fi
if [ ! -x $prog ] ; then
    prog=../maiko/linux.armv7l/ldeinit
fi
$prog -sc 1000x900 -g 1000x900 loadups/init.dlinit
```
should just be `ldeinit -sc 1024x768 -g 1042x790 loadups/init.dlinit`, and every other use of lde/ldex in the loadup scripts should use the same `-sc` and `-g` values so that the screen is not scrambled during the early stages of the loadup.

nbriggs:
I don't understand why all the scrolling commits are on this branch, they're not part of cleaning up the makeinit, are they?

masinter:
I did a rebase to pick up Ron's changes to lispusers following https://stackoverflow.com/questions/3876977/update-git-branches-from-master

nbriggs:
Ah, OK, I guess -- you didn't need to do that if his changes were independent of your changes.  If a merge (squash is not appropriate for the number of changes you're making here) showed no conflicts it would have just happened without your having to do anything extra.  When I'm working with Bruce, we seldom need to rebase.

nbriggs:
I'm looking through the changes.  I'm surprised that the first MDS data page moved up 16384 pages, and I don't know why.
```
-|\\FirstMDSPage| 16382) (|\\MaxMDSPage| 524285) (|\\DefaultSecondMDSPage| 65532) (|\\MDSIncrement| 512
+|\\FirstMDSPage| 32766) (|\\MaxMDSPage| 524285) (|\\DefaultSecondMDSPage| 65532) (|\\MDSIncrement| 512
```

masinter:
LLPARAMS from sunloadup moved to sources

masinter:
That is, I tried doing `git rm sources/LLPARAMS` followed by `git mv sunloadup/LLPARAMS sources` but it came out as removing the sunloadup version and updating the one in sources. I suppose I could have done it with two commits but I thought the diff was interesting.

masinter:
I'd appreciate some feedback on this PR. There's quite a bit of cleanup still #162

rmkaplan:
What kind of feedback are you looking for?

masinter:
I was waiting for a review before merging with master.  As it wasn't forthcoming, I just did the merge tonight.

There were some git problems I didn't quite understand at the time, and I had to dig out your changes. 
Could you update (git pull) and check that all your recent changes are in full.sysout?

MACINTERFACE ==> MODERNIZE, FILEBROWSER changes.

rmkaplan:
In master or makeinit7?

masinter:
I merged makeinit7 into master, so do
```
cd medley
git checkout master
git pull
./run-medley -full
```

rmkaplan:
The filebrowser, modernize, scrolling stuff seems to work (at least as well as I had it--still some glitches with reshaping Sedit when its menu is attached and with the goofy window arrangement of the CL array inspector, plus a buried Tedit grabbing the wheel scroll).

It also feels faster, at least in the way it starts up.

Looks good.

rmkaplan:
However, the change to run-medley to set -m 256 means that some existing sysouts don’t run.

I get a message:

sysout_loader: You can't specify the process size.
Because, secondary space is already used.
(size is 64, you specified 256.)

I tried  putting -m 64 as a run-medley parameter, thinking it would get passed through and override, but that didn’t help.

I suggest setting it up as a variable that defaults to 256, at least as a transition (although I just made a local copy with -m 64 built in).

rmkaplan:
Also I note that the function MEDLEYVERSION has disappeared.  An anachronism, to be sure, but my LFG loadup calls it to decide whether it needs to load some sort of version-dependent patches.  I’ll rip out that dependency.

masinter:
Can you do a loadup?  
```
cd medley
time ./scripts/loadup-all.sh
```
total time is still less than a minute on a 2009 iMac

rmkaplan:
Here’s what I get:

9.385u 2.938s 0:16.76 73.4%	0+0k 0+0io 1020pf+0w

This is on a 2017 13” macbook pro with 3.5 Ghz Intel Core i7, 16 GB memory.

Anzus:
I did get an error reported at the midpoint though, is this normal?  It still made the sysout.

start /Users/arunwelch/Interlisp/Medley/sources/YREM.CM
sysout_loader: can't open sysout file: tmp/init-mid.sysout: No such file or directory

masinter:
There are four steps:  loadup-init, loadup-mid-from-init, loadup-lisp-from-mid and loadup-full.
All .sh files in scripts.   loadup-mid-from-init needs a different maiko `makeright init` instead of `makeright x` will build ldeinit in the same directory as lde and ldex.  The scripts probably don't handle errors correctly.

rmkaplan:
Are you creating dribble files for the different stages where that makes sense?

masinter:
I couldn't figure out what good the dribble files were, so i wasn't religious about making them. What are they for?

rmkaplan:
I make them for all my loadups, as a logfile to reference when you are trying to understanding why things might have gone wrong.  From the sysout you know what files were loaded, but you don’t know, for example, what order they came in, how dwim might have plowed through potential errors, etc.

masinter:
I fixed run-medley so that it doesn't ignore a -m parameter but just defaults to 256.
About dribble files, I'd rather have 1 than 4.

Issue #177, opened 2021-02-08
Want better loadup / release method with Docker integration


masinter:
**Is your feature request related to a problem? Please describe.**

The Docker integration is a good demo of the idea, but the build stopped working with maiko not a subproject.

**Describe the solution you'd like**
1. Separate out 'building linux.x86_64 maiko container (which just needs a dev environment) from a minimal 'run medley' container
2. The run medley container should have more run-time files like fonts and dinfo
3. instructions on how to mount a folder in your file system if you're docker desktop host
4. ability to use a local X server as well as vnc

**Describe alternatives you've considered**
Maybe we'll be able to run in a browser rather than in the cloud

**Additional context**

waywardmonkeys:
I've been learning Docker recently for other things and have some ideas on how we can improve upon this.

For item 1, we can do a multi-stage build where the first stage is a full image with dev tools, but then we copy the required build products to the second stage which is the actual image that gets deployed.

Items 2, 3, 4 are things I don't know about yet.

The alternative ... I'm working on it slowly.

AbeJellinek:
Trying to decide how to proceed on this. Our current build file does use a multi-stage build, and it works (worked) well. The main issue is that a Dockerfile can only copy files from the directory that it is in into the container, i.e. `COPY ./foo.txt /` works, but `COPY ../other-dir/foo.txt /` doesn't. Now that Maiko is no longer a submodule, we're left with two options for how to run the build:

1. Have two Dockerfiles, one in `maiko`, one in `medley`. `maiko` will build with full dev tools, `medley` will copy only the relevant files out of the `maiko` image and build a final (slim) release image. Anyone wanting to build a new version of the Docker image will need to run two separate Docker builds in the right order - inconvenient and fragile.
2. Have one multi-stage Dockerfile [like we have now](https://github.com/Interlisp/medley/blob/master/Dockerfile). First stage builds the `maiko` image, second stage copies the built artifacts out of the `maiko` image and builds the `medley` image. To get around the Dockerfile directory issue, create a symlink in `medley` pointing to `maiko` that we clean up at the end. Better because we only need one Dockerfile and we don't have to tag the large intermediate (`maiko`) image, worse because it requires a wrapper script to create/dispose of the symlink, and because the symlink approach itself seems suboptimal. (But what is optimal here?)

@waywardmonkeys, curious to hear your thoughts about this.

masinter:
[...]

PR #196, opened 2021-02-18
Docker: run on non-x64, use ../maiko

AbeJellinek:
run-medley assumes that maiko/ will be organized neatly with the
executables in the correct (platform-specific) folder, but we don't know
what platform we're on when we're building via Docker (and I don't think
that we should need to). To deal with this, we just add the entire maiko
folder to the PATH rather than only the subdirectory corresponding to
the current platform.

Build is still two-stage.

Fixes part of #177.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
 FROM ubuntu:focal
 ENV DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Closing because I moved these changes to a branch out of better judgment; will reopen.

PR #197, opened 2021-02-18
Docker: run on non-x64, use ../maiko

AbeJellinek:
run-medley assumes that maiko/ will be organized neatly with the
executables in the correct (platform-specific) folder, but we don't know
what platform we're on when we're building via Docker (and I don't think
that we should need to). To deal with this, we just add the entire maiko
folder to the PATH rather than only the subdirectory corresponding to
the current platform.

Build is still two-stage.

Fixes part of #177.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
 FROM ubuntu:focal
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN apt-get update && apt-get install -y build-essential clang libx11-dev
-COPY maiko /build/
+RUN apt-get update && apt-get install -y make gcc libx11-dev
+COPY ../maiko/ /build/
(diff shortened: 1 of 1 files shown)


masinter:
i had thought that the docker image would be x86_64 with a particular (small) linux distro.

AbeJellinek:
It is. Sorry, should've been more clear. It still can (and does) run on x86, but I was testing on ARM and couldn't get it to build (because Docker on ARM runs an ARM VM by default). Now the Dockerfile no longer assumes x86_64, so it'll run fine on whatever architecture you happen to be using. We should still stick with x86_64 as the default.

masinter (changes requested):
I think we went over this on Monday and there were still some problems

PR #200, opened 2021-02-19
Restore fontprofile

masinter:
Not sure how these files FONTPROFILE and FONTPROFILEPATCH got deleted.
Seems to be missing the 2018 source file for FONTPROFILE.LCOM;2

This PR just restores the files into GitHub, we still need to set up reasonable font profiles for high res displays


rmkaplan:
A little scary, but go for it!

masinter:
I cant find the 2018 fontprofile source

rmkaplan:
The LCOM there was compiled in 2018, but it is compiled on the symbolic from 1999, which is there as FONTPROFILE.~2~.

So I think the problem is that an unversioned file FONTPROFILE is not properly linked to the versioned file.

masinter:
@rmkaplan I think you "closed" the Pull REquest without squash merging.

rmkaplan:
Probably, I find the git user interface, particularly on the pull page, very confusing.

PR #217, opened 2021-02-28
Fix run-medley -dimensions processing to make sensible choices

nbriggs:
Given a "-dimension WxH" argument, round up the W to a multiple of 32 for the
Lisp window width and use an X window geometry with an additional
22 pixels for both W and H to account for the current scrollbar size so that
the resulting X window will not require scrolling.

### DIFF
--- a/run-medley
+++ b/run-medley
@@ -59,8 +59,15 @@ while [ "$#" -ne 0 ]; do
             shift
             ;;
         "--dimensions" | "-dimensions")
-            geometry="-g $2"
-            screensize="-sc $2"
+            sw=`expr "$2" : "\([0-9]*\)x[0-9]*$"`
+            sh=`expr "$2" : "[0-9]*x\([0-9]*\)$"`
+            if [ -n "$sw" -a -n "$sh" ] ; then
+                 sw=$(( (31+$sw)/32*32 ))
+                 gw=$(( 22+$sw ))
+                 gh=$(( 22+$sh ))
+                 geometry="-g ${gw}x${gh}"
+                 screensize="-sc ${sw}x${sh}"
+            fi
             shift
             ;;
         "--geometry" | "-geometry" | "-g")


masinter (approved):
This is OK but I really wish we could do more for the case where no dimensions were specified.

I've been experimenting with different algorithms. The latest is to take the screen height in actual pixels (1080 in my case), and subtract 66 (1016 in my case). That is "screen height".
66 is approximate 22 pixels for scroll-bar + window title + taskbar. 
Then to get screen width, take 2^21 /height. For me, this gives the largest possible window in terms of screen real estate. 
This algorithm doesn't take into account what the X server reports and the kinds of bugs to expect, though.

blakemcbride:
I have been trying ldex without the shell script.  There are three sizes:

1. the monitor size
2. the Medley window size
3. the Medley window size when maximized

The -g option appears to set #2.  None of the options affect #3.  That's where the problem is.  I understand two things:

A. As designed, maiko can only handle a maximum size of  X (some number)
B. There may also be problems with the runtime system if #3 changes from one run to another

I think correcting this is a big win.  While in the process, addressing the font issue and multiple color issues would fit in well.

blakemcbride (approved):

nbriggs:
`-sc` sets the size of the underlying Lisp screen, its width is rounded up from what you give it to a multiple of 32.
`-g` sets the size of the X window within which the Lisp screen is displayed.  This includes the 22 pixels of scroll bar and borders (in both width and height)
The Lisp screen can be at most 65536*32 pixels (width*height with width rounded up to a multiple of 32) -- under that size it can be anything you set with `-sc`.
The Lisp screen can be changed from run to run (different `-sc` values) -- the system will re-layout the windows and force a redisplay.  The `-g` can change as well, all it will do is change the amount of the Lisp screen that's visible at once.

PR #233, opened 2021-03-08
Docker: build with side-by-side medley and maiko dirs

AbeJellinek:
I'd love for others to test these changes on their machines, because it seems that Docker is a bit more temperamental than I thought. The gist of this PR is that it makes Docker work with our new directory structure: to run a build, `medley/` and `maiko/` should be side-by-side in a directory, ideally by themselves, and you build with:

```bash
$ cd interlisp   # whichever directory contains medley/ and maiko/
$ docker build -f medley/Dockerfile . -t "interlisp/medley"
```

If you're on ARM, add `--platform linux/amd64` to the build command to build an x86_64 image.

Commit history is a little messy because of branch/merge problems, but I'll squash it before merging.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
 FROM ubuntu:focal
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN apt-get update && apt-get install -y build-essential clang libx11-dev
-COPY maiko /build/
+RUN apt-get update && apt-get install -y make gcc libx11-dev
+COPY ../maiko/ /build/
 WORKDIR /build/bin
 RUN rm -rf /build/linux*
 RUN ./makeright x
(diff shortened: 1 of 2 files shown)


masinter:
check out my changes

Dockerfile
```diff
@@ -18,7 +18,7 @@ RUN mkdir -p /app/maiko
 WORKDIR /app
 COPY ./* ./
```
stumbo, line 19: Why copy the directory?  Wouldn't it be better to mount the medley directory when Docker starts?  Copying Medley into the Docker image bloats it to 2.9 GB.  Without the directory the Docker image is only 376 MB.  By keeping it separate you could have one Docker container that runs multiple instances of Medley,  One could easily imagine the Maiko engine being on a different release cadence then Medley.

masinter:
@hjellinek was going to work on Docker configuration -- any progress?

hjellinek:
I literally just sat down to do some more work on it.  I have nothing to 
show yet, but hope to soon.

         Herb

masinter:
great. I thought I should walk you through the 'release' process: the scripts loadup-all.sh (and the five loadup-*.sh steps), copy-all.sh, and start-release.sh and release-one.sh. Those are all for building and releasing "by hand".

PR #234, opened 2021-03-08
Move all of sunloadup directory to obsolete

masinter:
these files may be of historical interest but not needed for loadup. Perhaps there should be a README for "obsolete" that explains what we know about them.


nbriggs (approved):
Looks OK.  You're not using the `FILESET` mechanism any more?  We probably still want to create a few different sysout configurations, but presumably that doesn't depend on the INIT.xxx files anymore.

masinter:
I was following Ron's loadup proccesses; in hindsight I might have followed filesets (and could reconstruct one) but for now just edit the file LOADUP-LISP/-FULL.

PR #235, opened 2021-03-08
Move unused package READ-PRINT-PROFILE from sources to LispUsers

masinter:
Not referenced or loaded up afaik


nbriggs (commented):
This is not completely unreferenced -- it's mentioned in internal/library/MULTI-COMPILE.
This exposes the fact that MULTI-COMPILE is not up to date with respect to the files in the sysouts -- it has lists that include various things you're no longer building in.  I'm not sure whether this (read-print-profile) should be in library or lispusers.

masinter:
i had originally moved it to internal/library, but I couldn't remember why.

masinter:
I'll move it to internal/library in a new PR

PR #236, opened 2021-03-08
Move unused X-window package sources to obsolete

masinter:
Not referenced or loaded up. Sources only (not compiled for 3.5). Corresponding code in Maiko abandoned.


nbriggs (approved):
LGTM

PR #238, opened 2021-03-08
remove note files

masinter:
Two files of file names corresponding to ?? looking for uncompiled filesS?
Anyway, not of historical interest even


nbriggs (approved):
LGTM.

PR #239, opened 2021-03-09
Move all networking surces to net directory -- no longer part of loadups.

masinter:
All of these files require working ethernet and someone to talk to. Of mainly historical interest, but moved to "net" subdirectory instead of "obsolete". This PR may not be independent of changes to loadups that used to reference them but no longer do.


nbriggs:
I'd like to still be able to build an ethernet-aware sysout.   I have the ethernet code working on Solaris 11.4 using the DLPI application interface, and as soon as I get a chance I'll be checking on using the BPF application interface that both macOS and the *BSD systems provide.

masinter:
this won't be done until I can show building a sysout with networking enabled

PR #240, opened 2021-03-09
Of historical interest: BCPL and Dorado microcode, moved

masinter:
no need for these now; maybe Dorado Microcode compared to Maiko size? Not sure but .DM was a kind of zip file?


nbriggs:
They're not compressed but do look as though there is some perhaps disk block information?   They seem to be all source code, with interesting comments such as, for bitblt:
```
%*-----------------------------------------------------------
Case A:  dest _ f(gray)

BBOp is taken from the following table:

    Function (class)    gray    srcFunc         dstFunc         ALU operation
        10 (A)          TRUE    normal          null            NOT A
        14 (A)          TRUE    complement      null            A

Entry point to this code is at GrayDestSetup.
It dispatches here after setup.

Timing, per scan line, including vertica<FE>^A^@^V<E1>l loop overhead:
        19 cycles minimum in the normal case
        +1 cycle per full word (excluding first and last partial words)
        +4 + (2 * # of munches) cycles if item is wider than 20B words

        12 cycles total in the "thin" case
```

and so on.

masinter:
I think BITBLT was originally done for smalltalk. Mesa did "PilotBitblt" for Star needs.

nbriggs (approved):
Seems fine.

PR #243, opened 2021-03-09
Main cleanup of new init process, scripts, etc

masinter:
I should have done this one first, before deleting things


nbriggs (approved):
I'll approve this, but it shows up a bug in the building of the full.sysout such that if you start the new full.sysout it reports a missing directory...
Screen Shot 2021-03-08 at 10 33 50 PM

PR #245, opened 2021-03-09
remove nits

masinter:


nbriggs (approved):
P4A.scn looks like data for a logistics planning package.  I agree it probably doesn't belong here.

PR #247, opened 2021-03-09
keep until issues issues raised, was loadup cleanup

masinter:
This is the target of all the little PRs


masinter:
These commits need to be sorted out and reasons written:
* b44b1e5 1894df8 1889678 1889678 dc03666 12ff4b7 43cef0b
* (just submit) 35d9f42 3c8b1f2 0670b57
* (need small tweak) 17e97c2 56361d5
* 
These have been handled by other commits
* 7bd0973 f7f25a6 650fc1e

rest needs to be sorted

PR #249, opened 2021-03-10
Make exports.all and whereis.hash in tmp directory. Add script to release.

masinter:
the move of the eval-when-patch (garbage) was because it gave a HELP when making whereis.hash


nbriggs (approved):
OK.

PR #250, opened 2021-03-10
removed files duplicated or older versions of library/lafite

masinter:
except the one tedit file which I moved in.
Wonder if there's a lafite demo that could use Linux mail?


nbriggs:
See the library/unixmail package.

masinter:
please approve or suggest changes :)

nbriggs (approved):
LGTM

PR #251, opened 2021-03-10
Save RDSYS for reading sysouts. Made by loadup-init

masinter:


nbriggs (approved):
Yes It's too bad there is no way to associate the RDSYS with the sysouts that it's capable of reading (other than it being in the git repository) -- fortunately we're probably not going to be in a position where anyone **needs** to try reading a broken sysout where the corresponding RDSYS isn't available in the repo.

PR #252, opened 2021-03-10
move older "new" TEDIT

masinter:
looks like the "new" files was a project spring of 95.
Some of the TEDIT* files are older, but most or newer.
So perhaps these were abandoned?


nbriggs (approved):
OK.

PR #253, opened 2021-03-10
Testing the loadup-all and release-all scripts

masinter:
but we should do loadup-all each push to master like with maiko

We can add more testing after, but it will be good to know if changes "break the build"


nbriggs (commented):
I don't understand why this shows checking in empty sysouts... it looks like a git oddity, though, as the sysouts are there when I look.

masinter:
conflicting loadups are newer

PR #255, opened 2021-03-10
COLOR on LispUsers older than copy in library

masinter:


nbriggs:
COLOR (lispusers and library) and DICOLOR (internal/library and lispusers) share functions, and should probably be dealt with together.  I suspect that DICOLOR functions (though perhaps not all of them) were incorporated into the COLOR code.

masinter:
how about approving this PR or suggest changes, and adding details of what else should happen to issue #6 (color support)

nbriggs (approved):
LGTM.

PR #259, opened 2021-03-11
add options to run-medley for new sysouts

masinter:


rmkaplan:
The -newfull option worked in my repository.  The -newlisp did not, but I don't haven't made a local lisp.

The -full option still fails.

I approve the merge

nbriggs (approved):
OK.  Not sure why it needs options for this when it takes a sysout name as a pass-through argument.

rmkaplan:
Well, Larry has set it up so that without a special tag your locally experimental sysout would be a little less convenient to reference than the normal (released?) versions.  But presumably those are the ones that you will want to test, without the (admittedly minor) nuisance of specifying a long path name for the file if connected somewhere else or first doing a cd.  Just another thing to forget or mistype.

Issue #23, opened 2020-08-11
PRINTCODE shows negative numbers in nametable


masinter:
(LLSH 8388808 8) returns a negative number. it should just print the bytes. in ACODE

masinter:
fixed by #261  changed default to be hex to match uraid display, and shortened name table lines

PR #261, opened 2021-03-12
new loadup full

masinter:
done with 4 commits:
 1.  sources and lcoms
 2.  putting back a  tuning in MAKE-EXPORTS-ALL
 3. new lisp and full with problems fixed
 4. other files that might not need to be updated


masinter:
ry loading this PR and see if it fixes the problems.
I was mv -ing my medley directory to test, but I’m not sure I tested all the cases.
* Loadup-all
* Loadup-full (from which lisp? It’s set up to start with the just-built lisp)
* Run-medley startup
* Lde call with no MEDLEYDIR set ?

nbriggs:
Still complains about LOADUP-FULL.LCOM having disappeared.

Screen Shot 2021-03-11 at 5 07 12 PM

rmkaplan:
full sysout still fails for me too

masinter:
this is  awful to debug. still working on it... loadup is an elaborate sequence of dependencies

nbriggs (commented):
Still can't find `LOADUP-FULL.LCOM`

nbriggs:
If you write output to a stream open on "{unix}/dev/stdout" you'll get output in the terminal window where you're running the loadup -- that may make it easier to see what's going on because the output won't disappear.  In the end, you won't be able to leave that in, not *all* systems support the name /dev/stdout, but it may help now.

masinter:
One mess is around LOADUPDIRECTORIES and DIRECTORIES and how they get initialized when and to what value.
This has to respond to your unique configuration. I'm not happy with the system, it will confuse anyone. It's very fragile -- minor changes have large ripple effects. Right now, there are too many phases.  I'd like to focus on getting rid of dlfixinit and ldeinit .... why can't makeinit's sysout be run without 'fix'ing it first? Then we'd have: makeinit + loadup as the path.

masinter:
The only idea i could come up with was making a {medley} device as an alias. You could do it as a shortcut, and still be compatible with code that packs on "{DSK}".  But maybe we could just fix those (they're all recent) and make it a new device.

I think we should try to document the constraints on changing code depending on what files are used where in the loadup process.

rmkaplan:
For me at least I still can’t run the full sysout, after a pull.

And I can’t locally make my own full sysout because I don’t have the tmp/lisp.sysout that the loadup-full.sh wants.  (unless I first copy to tmp the lisp sysout that I happen to have in loadups, or I change the load-full script to use loadups instead of tmp)

Is the idea that in order for me to make a full after a fresh clone, I also have to locally make a lisp, starting with doing the makeinit?  Which I can’t do because I don’t have a runnable full?

I think I’m getting lost in the dependencies.

masinter:
There's another problem which I can't figure out immediately. The whereis.hash file is bad. I get setfileptr errors.
I had to split making a new one into another phase `loadup-aux`

masinter:
to get started, you need a (not broken) full.sysout
loadup-init.sh looks in loadups/full.sysout

in this branch `git checkout new-loadup-full` `git pull`
before you starrt `rm tmp/*`
`./scripts/loadup-all`

rmkaplan:
I was able to pull this and use it to make everything from scratch, all the way to a full.

nbriggs:
@rmkaplan -- I'll reopen this so that we can merge it.

nbriggs (approved):
LGTM

nbriggs:
This set of commits included a change for the INSPECTCODE output from octal to hexadecimal, which was not documented anywhere...?

masinter:
Fixes #23 changed the default from octal to hex because uraid is hex. people are more used to hex now

masinter:
i had this as a draft PR. Ron probably meant to Approve the PR rather than Close it.
When you reopened it, it didn't remember it was a Draft. Might be a problem with git.

PR #272, opened 2021-03-15
Change make-whereis-hash to only index loaded files

masinter:
until we've straightened out the whereis.hash problem, this would simplify. But it's a good idea to separate the system wherein from things that might be arbitrary (maybe a whereis-other.hash?)


masinter:
NOW I understand what system.hash was for. I'll put it back so that system.hash is loaded files (in lisp.sysout) and whereis.hash has everything (or everything else?)

masinter:
alternate plans

Issue #267, opened 2021-03-13
SIMPLE-INIT is not closed after greeting


rmkaplan:
**Describe the bug**
Starting up with Lisp or Full, the file SIMPLE-INIT remains open when the greeting has completed.
**To Reproduce**
1. Files loaded '...'
2. Form to eval and/or menus to click '....'
3. What happened? '....'

**Expected behavior**

**Screenshots**

**Context (please complete the following information):**
 - OS: Mac
 - OS Version: Big Sur
 - Host arch: [e.g. x86_64, arm7l, arm64, sparc]
 - Maiko version: [e.g. commit ID from `git log | head`]
 - IL:MAKESYSDATE: 12-mar-2021 12:24:04 (full.sysout)

**Additional context**

nbriggs:
Given that the loadup process was implemented as a GREETFILE... this is probably the same fundamental problem.  I'd look at the GREET implementation to see what got broken.

nbriggs:
Fixed by PR #273

PR #273, opened 2021-03-15
Changes to interpret.rem.cm and sysout et all

### LINKED ISSUE #227: SYSOUT returns wrong value when resuming
**Describe the bug**
SYSOUT seems to return a (non LISTP) non-existant version when resuming

**To Reproduce**
make a sysout. LOGOUT and restart it
The SYSOUT call returns the name of a new (non-existent) version

**Expected behavior**
return a list of the name of the restarted version

**Screenshots**
image

image

masinter:
this fixes the problem of leaving init file open.
While I was there, I went through and fixed #227 , 2017 patches to ADIR weren't correct


nbriggs:
Also closes #267

nbriggs (approved):
LGTM

rmkaplan:
I made a new tmp/full from scratch, and the init files are now no longer open when it starts up.

However, I now get a stream-not-open error when I run the LFG loader as a greetfile, in the standard way.  The sysout does seem to be constructed before the error happens.  

Also:  it appears that the retrofitting changes to ADIR undid the intended effect of the previous patches, which was to make version-numbered sysouts and makesys's.  Now it seems like the new sysouts smash on top of the previous one.  \COPYSYS doesn't do version numbering, which is why the patch was to make the sysout under a tmp name and then rename it.

masinter:
\COPYSYS is a terrible mess.  A clue is its location (MOD44IO). It's a good idea to remove all the complexity of sysout to non-maiko-{DSK}.

PR #274, opened 2021-03-15
Remove awkward duplicate function CHECKTRAN

masinter:
The function CHECKTRAN had two definitions, one in DWIM and one in DWIMIFY with (in one case) different behavior.
The definitions didn't collide because the version on DWIM was blockcompiled away.
I renamed it to avoid the problem.


nbriggs (approved):
LGTM.

PR #275, opened 2021-03-15
SYSEDIT was resetting MSMACROPROPS for no good reason

masinter:


nbriggs:
In a full.sysout, MSMACROPROPS was already set to COMPILERMACROPROPS, however in a lisp.sysout, MSMACROPROPS is unbound.  Why is it not correct for SYSEDIT to set up MSMACROPROPS if it should happen to be loaded into a lisp.sysout?

masinter:
need to figure out why MACRO-FN wasn't on it, and when MACRO-FN is used.

rmkaplan (commented):
Seems like this is still under discussion?

masinter:
yes, it was under discussion. The question was why setting the value in a sysout that didn't contain Masterscope was wrong.
MSMACROPROPS is only used by functions on MSANALYZE and is set to a different value there (now). So the resetting of MSMACROPROPS is wrong in sysouts that already had a value set by loading MSANALYZE, and not useful in sysouts that don't have MSANALYZE set.

Issue #271, opened 2021-03-15
Loadup-all should stop when one phase fails


rmkaplan:
The loadup-all scripts invokes a sequence of scripts for the different phases.  The script doesn't notice and stop when a phhase  fails and doesn't produce its intended output to feed into the next phase.  If there is a preexisting artifact in the tmp/ directory that the next phase depends on, it just goes ahead and uses that.  And if it can't get all the way through, the final artifacts (lisp.sysout and full.sysout) are not created, and previously existing ones masquerade as if they were new ones that were just built.

For example, my loadup failed just after the makeinit because I apparently don't have an ldeinit (./makeright x apparently doesn't make it).  But the loadup went on, scrolling an enormous amount of stuff in various sysouts (a lot of it having to do with the whereis), and all of it goes away, can't be scrolled.

When I went to test my new features (just in the full, didn't depend on anything else, but the scripts unfortunately at this point are oriented towards making everything from scratch), it took me quite awhile to figure out that I was still looking in an old sysout, the loadup had failed.  Scolling back just in the terminal window is where I found a message about the absent ldeinit.

If the loadups are chained in this way, the different phases should have a way of posting error codes.  And it shouldn't be required that a complete loadup is necessary to work on just the lisp and full sysouts, even though speed is not an issue on current hardware.

nbriggs:
I think this should be a Makefile encoding the dependencies so you don't have to build everything every time.
I'm not sure that we currently have a way for Lisp to exit while returning a normal (Unix) process exit code that scripts can easily interrogate, but I see no reason we couldn't implement it.

I'm looking at whether we can dispense with a separate ldeinit, or at least build it without compiling everything twice and always build it if lde/ldex is built.

masinter:
The problem originally reported has been fixed by touching a 'loadup.timestamp' file. A makefile would be better if there were some reasonable way to create dependencies

PR #276, opened 2021-03-17
a litte better info and checking

masinter:


scripts/loadup-init.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+o#!/bin/sh
```
nbriggs, line 1: Looks like a typo included here

run-medley
```diff
@@ -172,9 +172,15 @@ export PATH="$MAIKODIR"/`osversion`.`machinetype`:"$oldpath"
 
 cd "$OLDPWD"
 
-echo "sysout is $LDESRCESYSOUT"
-echo "running `which $prog` $geometry $screensize $mem"
-echo "start $LDEINIT"
+whichprog=`which $prog`
+
+if [ -z "$whichprog" ] ; then
```
nbriggs, line 177: `which` isn't the best choice here -- it's a c-shell-ism. One alternative is
```
if command -v $prog > /dev/null 2>&1; then
    # we're good
else
   # we're not
fi
```

masinter: trying to fix #271

masinter: I don't know what you mean by command? Not part of my Linux.

nbriggs: It's literally the word "command"

nbriggs:
Are you going to roll this branch back to the 0ba2ce (before the mistaken commit of the GETHOSTINFO change) and force push it so that we can approve/merge it?

masinter:
I am confused by the git status; not sure what 'force push' means in this context

nbriggs:
You have a branch... named script-tweaks that needs to be adjusted before the pull request should be done.  You check out the branch, then checkout the commit before the last, then "git push --force" which will update github with the branch rolled back to the place you had checked out.   Then the PR can be done and it will only include the branch in its new fixed state.   If you don't use "--force" it will not let you push the branch because it would not be simply adding commits.  Never use "--force" on "master" or on any branch that has already been merged anywhere.

PR #278, opened 2021-03-17
Checktran called from outside block, remove one blockcompile

masinter:


nbriggs:
I'm guessing you meant CHECKTRAN+ (not Checktran) is called from outside NEWFAULT1... where it's called from VARSBOUNDINFORM?

If so, this looks good to me.

masinter:
i should redo from original DWIM and DWIMIFY.

Issue #213, opened 2021-02-25
inconsistency between subr.h and LLSUBRs list of subr numbers


nbriggs:
The checked-in subr.h is not consistent with the collection of subr numbers generated by LLSUBR's `WRITECALLSUBRS` based on the `\INITSUBRS` list.  The same issue occurs in the same subr.h file for the miscn_XXX values.

These need to be consistent and reconciled with the code that implements the subrs and miscn opcodes. 

Additionally, the output of `WRITECALLSUBRS` needs to be updated to include an include guard.

nbriggs:
The subr numbers consistency part of this was fixed by commit 850a0ce3cec3462cc7ce8cececcd8b3dad62f900.

PR #279, opened 2021-03-17
Add writing guard to subrs.h and running again

masinter:
the result doesn't match maiko/subrs.h but that's another issue
#213 
waiting for https://github.com/Interlisp/maiko/pull/343


masinter:
it's awkward to split out independent branches while running a sysout without reloading all the time. I had two PR's open: one for DWIM and a separate one for LLSUBRS.  I didn't have any way of running with both. Switching branches and keeping the image coordinated with what's happening in the file system is awkward. (Thus the MAKEFILE(LLSBS NEW)

nbriggs:
This doesn't (shouldn't) depend on Interlisp/maiko#343 -- it only needs to write the list of subrs (and misc opcodes) in the newer format (with a guard).  The medley repo shouldn't contain a subrs.h -- that will be generated from this code and checked in to the maiko repository.

A separable task (which is trivial) is to check that all subrs and misc opcodes currently defined on the lists are implemented in maiko.

nbriggs:
Also need to check for SUBR codes that are passed numerically rather than by name as in
```
(\BLTCHAR
  [LAMBDA (CHARCODE DISPLAYSTREAM DISPLAYDATA)
    ((OPCODES SUBRCALL 135 3)
     CHARCODE DISPLAYSTREAM DISPLAYDATA])
```
(this one is actually NEW-BLTCHAR (=135))

PR #282, opened 2021-03-19
fix whereis making

masinter:
two patches, one to increase the estimated size and the second to change the build target back from {CORE} to {DSK}

I'm not sure why that stopped working but this makes whereis.hash buildable again.


masinter:
i used cherry pick 6967e97 and resubmitted
will fix 6eb69eb, pr independently

PR #283, opened 2021-03-19
move staging file for whereis.hash (back?) to {DSK}

masinter:
(resubmitted unbundeled)


nbriggs (approved):
Seems OK.  It avoids a copy from memory back to disk in the rename step.

PR #277, opened 2021-03-17
Fix sysout makesys to make new versions

masinter:
wound up pushing into \COPYSYS. Move generic \COPYSYS into ADIR (and also VMEMSIZE) to remove MOD44IO from loadup.

PR #293, opened 2021-03-22
Revert "Fix sysout makesys to make new versions"

masinter:
Reverts Interlisp/medley#277

too eager


nbriggs:
Can't you test before merging things to master?

If master has advanced (with tested working fixes) past where you started with a particular branch then you can rebase your branch on the current master to pick up changes you need from that.

PR #296, opened 2021-03-22
sysout logout version

masinter:
tested writing new sysouts and doing logouts etc. Not able to fully test sysout to a device other than {DSK} or {UNIX} (for which it DOESN'T make a new version)


nbriggs:
Seems to work OK.  It looks as though you're "losing edits" because you're not rebasing onto master and then making the additional changes to files that were previously fixed.   Try `git help rebase` to see how to avoid working with/committing old files.

nbriggs (approved):
LGTM.

PR #305, opened 2021-04-17
no need for PSETF (file nowhere to be found). Guess it was psetf macr…

masinter:
…o which just expands to CL:PSETF


nbriggs:
PSETF is in CMLSETF where it is defined with a defmacro.

masinter:
In less telegraphic: There apparently was a file called PSETF somewhere, but it isn't in any of the dumps I have from envos, from the 2016 copy that Ron got from John. There was a definition of (lower-case) il:psetf as a macro that expanded to CL:PSETF. The only thing I did was remove a FILESLOAD of PSETF.

nbriggs:
Is there any reason that it should have a definition of il:psetf when there's a cl:psetf macro?  That is, why doesn't it just use cl:psetf directly in the code?

Anzus:
It’s not in any of my copies of the sources either.

…arun

masinter:
it's a Lispusers package (used by AIREGIONS) that had 5 different authors over time. "Why?" is not a question easily answered.

nbriggs:
OK, "should it now use `CL:PSETF` directly and not define `il:psetf` to expand to `CL:PSETF`"?

masinter:
My take when patching someone else's Lispusers package would be to minimize changes. It would be different if it were in sources or library.

masinter:
Willl redo and resubmit

PR #306, opened 2021-04-17
loadup cleanups

masinter:
the new loadup process doesn't use these files


masinter:
This removes one file (a dribble spelled differently) and moves files that were part of old loadup processes that are no longer used.
The moved files are still in obsolete/sources (which can remain in the repo but not be part of the release).

PR #308, opened 2021-04-17
New loadups

masinter:
i'm still not sure when we should do loadups... Maybe after we have nightly releases going it will be clearer?


masinter:
these are obsolete. When we have nightly builds, maybe we won't need commits like this.

PR #312, opened 2021-04-22
Change default EOL to LF on UFS

rmkaplan:
UFS changes the the default in \UFSeol to LF.EOLC.  LLREAD changes \rprint2 to convert EOL to LF if escaped in a string.  NSPROTECTION eliminates literal EOL

Note: this pull request is just for the code changes to change the default EOL convention.  Round-trip testing has been done on all current and translated source files, but  translated source files have not yet been included.    (See notes at issue #2)


masinter (approved):
I looked over the source changes, but didn't test it. LGTM

PR #314, opened 2021-04-27
Add a brief sanity check to loadup process

masinter:
Touch tmp/loadup.timestamp before starting, and make sure the loadup produces newer files.


rmkaplan (approved):
Seems reasonable

PR #315, opened 2021-04-27
All source files converted to LF

rmkaplan:
Also, HPRINT:  EQUALALL knows about CL arrays
FILEIO:  STREAM record with fields for external format functions

No LCOMS other than FILEIO.LCOM and HPRINT.LCOM


masinter (changes requested):
you have a new file loadups/makeinit.dribble
which I renamed to init.dribble months ago.
I wish the FILEIO had been done separately
you did sources, lispusers, library but not internal/library, docs/Document Tools, clos, cltl2 etc.
it's OK, but we'll get more files and there should be some documentation about how to import these files into Medley.

rmkaplan:
I don’t track those directories.  Is that the complete list?

masinter:
I expect we'll add notecards? There are some sources in lispusers/*/.  (i.e., subdirectories) and possibly in library. The list of directories I scan with MEDLEY-FIX-DATES (function on MEDLEY-UTILS) is MEDLEY-FIX-DIRS (I think). 
If the idea in using LF is to make source files work with GREP and DIFF, it's better to have them fixed if they're in the medley repo. It would be really helpful to also add a pointer to what's documented. If I bring in Notecards, what do I need to do?


image

rmkaplan:
The first question is whether it is safe to assume that the files in these directories without extensions are source files that need to be transformed.

Second question is whether they need to be tested—switch back and forth, equal readfile?—before transforming them.

When it is presumed safe, then either a tr can be done, if you don’t care about preserving the previous (CR) versions (other than by git), or my little lisp routine.

Then also presumably files with extention txt or text, although I think for text files we would just assume that the translation can’t hurt.

I will run my little testing program on all the no-extension files in the MEDLEY-FIX-DIRS directories.

rmkaplan:
I tested internal/library/ and docs/Documentation Tools, all OK, so I pushed the converted files.

Those were the only directories on MEDLEY-FIX-DIRS that I hadn’t done.

What are the commonlisp (or other directories) that matter?

masinter (approved):

PR #318, opened 2021-04-29
Remove unused file reference to PSETF and  (lower case) psetf macro

masinter:
The Lispusers package FILLREGION is used only by AIREGIONS. It used to contain a (FILES PSETF) -- which file is not to be found -- and a macro (lower case) il:psetf, which is not used. FILLREGION didn't have a compiled file, likely because of the dangling reference.

AIREGIONS-DEMO is a nice lispusers package to illustrate inventive Lisp programming.

The fix was to remove (FILES PSETF) and (FUNCTIONS psetf), MAKEFILE(FILLREGIONS NEW) and BCOMPL(FILLREGION)


masinter:
low risk Lispusers cleanup

PR #322, opened 2021-04-30
Use .CM to loadup full, to avoid GREET call

### LINKED ISSUE #316: MAKESYS doesn't clear all user state
MAKESYS of the full sysout doesn't clear out user state -- when I make the FULL sysout, Iron sees Larry in the who-line, while it greets me with "Hi, Ron" when I start a full made by Ron.

There's a little protocol that MAKESYS before it makes the sysout tries to clear all the state, and rebuild them when it's started. (Including settings of MEDLEYDIRS).
SYSOUT does the same dance but it's assumed that the GREET was done and just issues a warning. GREET is supposed to be undoable. The last GREET is remembered in a special cache to enable this.
@rmkaplan this is what I think is going on.

masinter:
Fixes  #316 by using the same method for loadup-full as was used by loadup-lisp, using a BKSYSBUF instead of having LOADUP-FULL function do it.


masinter:
this fix is needed to keep the builder's init from being in the sysout

nbriggs (approved):
Seems to work OK for me.

PR #323, opened 2021-04-30
Add a loadups/full.database masterscope database

masinter:
wasn't quite sure where to put this.
Creating it isn't quite automated yet, but it's close.
```
(pushnew DIRECTORIES (MEDLEYDIR "greetfiles")
(for X in SYSFILES do (LOAD X 'PROP))
LOAD(SYSEDIT)
UNSET(MSMACROPROPS)
. ANALYZE ON ANY IN SYSFILES
```
need to `RETURN NIL`  because of the interpreter NOBIND error (separate issue)
Then 
```
(RENAMEFILE (MAKEFILE 'DATABASE.FULL) (MEDLEYDIR "loadups" "full.database" T))
```
  In the meanwhile, this PR just adds the database file.


masinter:
this just adds a single file, as requested by Ron, of a masterscope database of "everything"

PR #324, opened 2021-04-30
Upgrade TEDIT to LF

rmkaplan:
Fixes the black boxes in the Tedit menus.
TEDIT had a built-in bias towards CR in files and in displaying them.  I changed the bias to LF.

I checked several round trips, formatted and unformatted.  But Tedit has lots of internal optimizations at this level, and some pieces are pointers into files and some pieces are internal strings.  So there may still be some unscrambling.  In some cases he did some brute force copychars into {NODIRCORE} to normalize things.


masinter (commented):
Is it a good idea to have the same declarations in two different places? Isn't there some risk of divergence? I've been working to stomp out having the same function defined in two different files.

Is TEDITDCL.LCOM loaded or used?

rmkaplan:
TEDITDCL is loaded whenever the symbolics of any other TEDIT file is loaded.  It appears to act as a TEDIT-specific version of EXPORTS.ALL.

I think the way it is supposed to be created is in the following sequence: 
	LOAD(TEDITDCL)
	LOADFROM all the other TEDIT files (smashing TEDITDCL declarations with new ones)
	CLEANUP(TEDITDCL)
	compile
That would ensure that TEDITDCL contains copies of all the current declarations.  Presumably only needed when one of those changes.  That seems to be how John set it up.

What I discovered was that the current symbolics had only the FILECOMS, not any of the content, even though the LCOM was complete.  So I did that sequence.

We could get rid of it, and put in a little function EDITTEDIT() that did the LOADFROM of all the files, assuming that that wouldn’t smash anything critical.

Or mark all the declarations with EXPORT, and provide a little function that does (GATHEREXPORTS teditfiles ’TEDIT-EXPORTS).

I don’t think you would want all of the TEDIT declarations to end up in EXPORTS.ALL—you might get various kinds of conflicts.

masinter (approved):
I'd guess TEDITDECLS.SAVED is obsolete.
GATHER-EXPORTS (called by MEDLEY-MAKE-EXORTS-ALL on MEDLEY-UTILS) 
```

rmkaplan:
Yes, there were 2 of them hanging around with different names, I kept that one just in case.  Toss it.

Issue #326, opened 2021-05-01
Hardcopy to a file in TEDIT:  Asks for PUP host number


rmkaplan:
**Describe the bug**
When I click hardcopy to file in a Tedit window, there is a long delay before I get a pop up asking me for my pup host number

**To Reproduce**
New full Master sysout on a mac
TEDIT(file)
right click hardcopy in title bar, selected "to a file"

**Expected behavior**
No delay, no pop up.

**Screenshots**

**Context (please complete the following information):**
 - OS:Mac
 - OS Version: [e.g. Big Sur]
 - Host arch: [e.g. x86_64, arm7l, arm64, sparc]
 - Maiko version: [e.g. commit ID from `git log | head`]
 - IL:MAKESYSDATE: [ date ]

**Additional context**

masinter:
any  chance your init sets DEFAULTPRINTINGHOST?

rmkaplan:
Yes, it sets it to the empty string.

It sets DEFAULTPRINTERTYPE to POSTSCRIPT

It also loads lispusers/SETDEFAULTPRINTER (dated June 1993).

rmkaplan:
The request for the Pup host comes from \CANONICALIZE.PUP.HOSTNAME, called from under CANONICALHOSTNAME under PRINTERTYPE

rmkaplan:
I took SETDEFAULTPRINTER out of my INIT, but it seems that it is still included separately in the full loadup.

masinter:
the long delay is also a problem. There is something missing as far as trying to run PUP even if the network isn't turned on (which it isn't in the regular maiko). Type a control-B when you first start

MAKESYS on startup calls \DEVICEEVENT(AFTERMAKESYS) which calls \ETHEREVENT

nbriggs:
Somewhere it should probably be calling (SUBRCALL ETHER-AVAILABLE) before trying to do network related things.

masinter:
The problem reported is fixed b #327 . The problems of PRESS and INTERPRESS printing were noted in a separate issue #328

PR #327, opened 2021-05-02
Better handling of no-ether when starting without ethernet

masinter:
No longer wait for PUP or NS to be ready if ETHERNET-AVAILABLE returns false.
Add checks to \DEVICEEVENT -- if no network, just leave it off.
Moved CANONICAL.HOSTNAME from PUP to LLETHER because we might move to a setup where TCP is on, but PUP and XNS are not loaded.

Partly addresses #326 but doesn't fix it completely -- if you set DEFAULTPRINTINGHOST to an empty string, it calls NSPRINT which errors. The printing code is pretty complex and I'm not sure what is wanted.

I think it should be possible to have a loadup with POSTSCRIPT (and PDF?) but no INTERPRESS and PRESS.


nbriggs:
It should be the case that if you don't ask for PRESS or INTERPRESS that the code doesn't get invoked.
`IL:DEFAULTPRINTERTYPE` should probably be set to `IL:POSTSCRIPT` instead of `IL:INTERPRESS` (as part of the site init file?) and the code that looks at the output stream type for hardcopy should know that an output stream type of `IL:POSTSCRIPT` suggests a file extension of `.PS` (it currently only knows INTERPRESS => IP and one other thing, perhaps PRESS=>PRESS)

sources/LLETHER
```diff
@@ -1471,10 +1498,13 @@ Copyright (c) 1982, 1983, 1984, 1985, 1986, 1987, 1988, 1990, 1991, 1992, 1993 b
                               HOSTNAMEP _ 'NILL])
 
 (\ETHEREVENTFN
-  [LAMBDA (DEV EVENT)                                    (* ; "Edited 15-Jan-88 01:30 by bvm")
-    (SELECTQ EVENT
-        ((NIL AFTERLOGOUT AFTERSYSOUT AFTERMAKESYS AFTERSAVEVM RESTART) 
-             (PROG (NDB TURNOFFNS TIMESET)
+  [LAMBDA (DEV EVENT)                                  (* ; "Edited  1-May-2021 19:23 by larry")
+    (PROG (NDB TURNOFFNS TIMESET)
+          (OR (SUBRCALL ETHER-AVAILABLE)
```
nbriggs, line 1503: I'm not convinced this is the right place to do this -- if you make a sysout on a machine with ethernet enabled, then you need to do the cleanup that follows (e.g., clearing `\PUP.READY` etc., clearing the ETHERPORTS) or you'll get the wrong behavior.

nbriggs (changes requested):
see inline comment

sources/LLETHER
masinter: good catch, updated. I made a separate issue for the printer problem.

masinter:
made suggested changes

PR #333, opened 2021-05-13
utility git-versions.sh which pulls old versions out of git history

masinter:


masinter:
what I'd like to do is get rid of the `file.~NN~` versions in the repository and instead have those versions recoverable.
So when you want to work on FOO you can git-versions.sh FOO. Then we won't need MEDLEY-FIX-LINKS any more.

I imagine it is possible to write this in Lisp instead of sh, with calls to ShellCommand to invoke git, ln (for making hard links)

masinter:
i'll cherry-pick this into two commits.
One to add utilities to remove old versions and one to restore them.
The second to remove old versions being in 'master'; this will be draft until adding the utilities is merged into master.

PR #346, opened 2021-06-24
Some files were checked in with only versioned files (from unix side)

masinter:
move versioned file `file.~nn~` to unversioned `file`.

This was a problem from the beginning first commit of files from Ron's Interlisp/sources commit.


rmkaplan (approved):
lgtm

PR #347, opened 2021-06-26
Add scripts rm-versions.sh and restore-versions.sh which can ne used ...

masinter:
These are scripts which can be used in a repo to remove and restore versions of the medley repo .
restore-versions can be used in a clean newly-cloned repo (or after a rm-versions) to restore all versions, including both from those files originally checked in with explicit `~nn~` versions and also each file we've made since started using git.


nbriggs:
Does it order the created ~nn~ versions by the date of the commit? Or the order of the commits? Or the date within the file?  (e.g., when you cherrypick a commit the commit may have an earlier date but will come in order after more recently committed changes)

masinter:
It's in the order expressed in line 83
```

for commit in `git log --remove-empty --reverse --format="%h" "$file"`
```
so: the order from git log, reverse order, omitting events that don't change this file

If the file had any ur-versions (versions checked in ) then n starts with their max. (that's all this is doing:

image

masinter:
superseded by smaller PR

Issue #211, opened 2021-02-22
CLIPBOARD:  needs testing on non-mac platforms


rmkaplan:
I pushed a new version that uses the mac pbpaste and pbcopy only if (UNIX-GETENV "OSTYPE") includes "darwin", otherwise defaults to "xclip" for both streams.

I don't know if this is correct, please test on all of your favorite platforms.

masinter:
on Linux/wsl there is no "xclip", and I don't understand how it would work.
You need the OS Type of the X-server, not the medley process.

rmkaplan:
Can you copy-paste between linux applications?

If not, then I would just change clipboard so that at least it works between different windows in the same medley, for internal consistency.

Otherwise, we have to figure out what the tie-in should be.

nbriggs:
> ... I don't understand how it would work.
You need the OS Type of the X-server, not the medley process.

That is not correct.  The command is going to be executed on the system where the `lde` process is running.  The OS type of the server where the display is being presented is irrelevant.  The clip/copy/paste command communicates with the (potentially remote) X server to deal with the text.

nbriggs:
On most Linux distributions you can install the `xclip` package.  Check and see if that's available under WSL.

nbriggs:
If WSL is running Ubuntu, try "sudo apt install xclip"

nbriggs:
In `CLIPBOARD.PASTE.STREAM` the call to `xclip` needs to be `xclip -o` (the default is `xclip -i`)

nbriggs:
In fact, the best results may come from using the two commands
```
xclip -i -selection clipboard
xclip -o -selection clipboard
```
because then, if you are running on a remote machine, with display on a Mac, with clipboard/pasteboard sync turned on, the copy selections you make will be available in the Mac paste buffer to paste into any window not just within Medley.

rmkaplan:
So it is still CREATE-PROCESS-STREAM, just with those elaborated arguments?

Is there a way of testing whether xclip has been enabled?
If it isn’t enabled, then I think the clipboard should just simulate an internal Medley stream.  And presumably not bother with Unicode conversion.

I’m unclear about the remote-machine scenarios.
	If I’m on my Mac (my display) but Medley is running on a remote machine:
		If I copy a medley string to the clipboard, would it be available to
			paste in other apps on the remote machine, or
			paste in other apps on my Mac
		If I am pasting into Medley, should I get strings only that were
			copied by another app on the remote machine, or
[...]

Issue #207, opened 2021-02-21
don't add new dates (> 2003) to old copyright


masinter:
When following normal development in Interlisp, it will add a new year (2021) to the existing COPYRIGHT property.
I think the practice with maiko was to leave copyright, it was OK to remove license terms, but not to add any more copyright notices automatically.

nbriggs:
If you don't record any information to the contrary, I believe that [copyright ownership automatically resides with the author](https://www.copyright.gov/title17/92chap2.html), whether the notice is there or not.  It's a lot easier if it is recorded.

nbriggs:
I agree that we shouldn't be adding new years to the Venue/Xerox/other notices.  My comment on #203 was that if the current copyright owner info is not the same as the existing copyright owner info in a file, we should not add the new year to the old list but rather add a new copyright statement with the new owner and the new year.  If the copyright owner is not set, or the mechanism is turned off, then obviously the code shouldn't do this.

masinter:
Copyright dates are still being added -- see #348    we can write copyright Interlisp.org

rmkaplan:
Should we make this condition on seeing Xerox, Venue, or Envos?  That is, if the copy right is one of those and we are making a change, change the copyright to Interlisp.org?  (

masinter:
The guidelines I'd like to follow are:
* Do not **delete** any copyright notice or date
* Shortening ranges is OK
* It is OK (and preferable) to delete license statements or claims "All Rights Reserved" 
* I'm not sure about whether to turn off any modification of copyright properties, but i think it's best
* I suggest we add a NOTICE.md about copyright ownership and link it from LICENSE.md

stumbo:
Apache has a tool, [rat](https://creadur.apache.org/rat/) that is used to validate copyright notices in files.  Maybe we can harness something like this to validate files are correctly copyrighted.  

If we can wire it up, we could have part of the approval of a PR require passing an audit.

nbriggs:
Interlisp is already managing the copyright notices, in a more sophisticated manner than rat does, when one edits the files.

johnwcowan:
Updating copyright dates has no legal effect in practice.  Who cares if our code goes into the public domain in 2073 or 2110?  It may serve a *social* function in that people may assume anything with a copyright date less than a few years ago is abandoned, but that's not our target audience, I trust.

My recommendation: "Add a copyright date when a file is created; leave it alone after that."

Issue #121, opened 2020-12-31
CHAT library needs update (see summary)


masinter:
prints in terminal window "Open lisp side of PTY failed."
(This isn't new behavior, just reporting it now)

nbriggs:
OS and processor architecture?

masinter:
I tried it on a mac darwin.x86_64 and on WSL linux.x86_64 with same result.
A different result in full.venuesysout -- \\ILLEGAL.ARG error.

nbriggs:
Works ok for me in a full.venuesysout, with lde built at commit d717946929b3eda5b1a05fb1d491dd404d88f9ec (current head of master).

Make sure that the ldex that lde is finding is the current one -- if you're running from a directory that isn't on your search path the lde may find an old one that **is** on your search path.

masinter:
i did a pull and makeright starting with full.venuesysout on three different systems (WSL linux, MacOS high sierra, raspbian) 
the problem is in FORK-SHELL 
![Screenshot 2021-01-01 173653](https://user-images.githubusercontent.com/1116587/103449083-fd45ae00-4c57-11eb-8a8b-fc73c722f9c0.png)
prints "Open of lisp side of PTY failed." and fails. Some environment variable you set and I don't?

masinter:
I tried compiling unixcomm.c with -DDEBUG and got this output

Entering FindUnixPipes
NPROCS is 1024; leaving FindUnixPipes

Unix_handlecomm: trying 8

Unix_handlecomm: trying 11
[...]

Issue #39, opened 2020-09-26
odd glitches with huge screen


masinter:
I have a big display. When running through X I get a short window with lots of garbage. Some kind of overflow
    https://gyazo.com/cec6ddc1f7712edf4ba1a24b1cea000e

moon-chilled:
If it's signed overflow, compiling with `-ftrapv` should catch it.  (Unsigned overflow will need to be debugged the old-fashioned way.)

masinter:
https://i.gyazo.com/b1c2f37b7f93b18477990306c1e35ec9.png 
(started with 
 lde -geom 1000x800 -sc 2000x1080  -d mywindows-machine:0
on a Raspberry pi

nbriggs:
It's not overflow, it's because there's a limited amount of memory that Lisp can address for the frame buffer (or backing bitmap for an X window):
```
../src/initdsp.c:  if ((displaywidth * displayheight) > display_max) { displayheight = display_max / displaywidth; }
```
and if X presents Lisp with, say, a 4000 x 2000 pixel window where it's expecting something more like 1440x900 it'll allocate it inappropriately (see above).  I'm not sure whether it would be better to give you a 4:3 aspect ratio that fits within the given (X) screen size, or something with the same aspect ratio as the (X) screen that fits.  XQuartz on macOS with multiple displays has this issue and it doesn't (seem to) offer any way to determine the physical screens that compose the single large virtual one it presents.

blakemcbride:
Here is some analysis of the problem (which is more complex than supporting a large display):

Let me define three sizes:

1.  The size of the monitor
2.  The size of the maiko window on the monitor
3.  The size of the maiko window when maximized

Maiko has some startup parameters that affect #2.  

The size of #3 is fixed or hardcoded.  

This is a problem for several reasons.

A.  Maiko can't handle larger displays - like mine.
B.  Regardless of your command-line settings, you can't change #3.  You should be able to specify #3 and then just get scrollbars when #2 is less than #3.

I have communicated with Nick about this.  He indicated that this was a big job to fix.  Nevertheless, it needs to be fixed.  It is a huge blemish on the system that makes it cumbersome to use.  Imagine trying to release an app in Medley and having to explain why Medly can't use its display properly.

If you add support for larger displays then, if the system continues to function as it is, that'll just cause more scrolling on smaller displays.

What needs to happen is the window size when maximized needs to be determined from the command line.  In other words, it needs to be variable.  Right now it is static.

[...]

PR #348, opened 2021-06-28
Cleanup character io interfaces

rmkaplan:
I think this is ready for more general testing and eventual merging.  It isolates the XCCS/NS external format dependencies into one new file XCCS, and all character access then goes through a few abstract functional interfaces (\INCCODE, \PEEKCCODE...), with the XCCS implementations still set up as the default.

This required touching and recompiling lots of files, but it all seems to hang together in this full sysout and in the LFG system built on top.

Other improvements/fixes to MODERNIZE, WHEELSCROLL, CLIPBOARD, and others.


rmkaplan:
I have updated to fix some glitches that I observed in further testing.  In particular, the interface to MAKEFILE now lets you specify a new format (like :UTF-8), and that propagates through remaking and also through compiling and recompiling.

Those are new capabilities.  The main strategy for reviewing and testing this branch would be just to run in it.  All old functionality is supposed to work as it used to, despite the changes to low-down interfaces. That's the first confirmation.

It may be appropriate to merge when all old behavior appears to be stable and correct.  Then we can inch up on the new encoding capabilities.

masinter (commented):
why is sources/LLREAD.LCOM.~1~ on .gitignore? it already ignores *~

masinter (commented):
there were changes (introductions of functions) to BIGBITMAPS in the library -- not loaded in FULL. Were these not part of the IO functions?

masinter (commented):
CHAT was recompiled with no changes. CHAT(SHELL) is the only useful CHAT target.
But shouldn't it assume :UTF-8 ?   I think CHAT(host) could turn into ssh but for now it is better to leave out?
CHAT(SHELL) currently always opens /bin/csh (C-shell) which isn't in most Linux distros. See #121

rmkaplan:
I think this was something else that got included to make the masterscope analysis more accurate, making sure that all ambiguous field names were qualified.

rmkaplan:
Could be that macros or records changed underneath.

rmkaplan:
Dunno

masinter (commented):
there was a change to SYSEDIT that removed resetting MSMACROPROPS to leave MACRO-FN on the list, leaving the value on MSANALYZE. Macros for all kinds are analyzed by analyzing their expansion. There was also a patch to remove / no longer generate spurious incorrect masterscope "templates" (call to SETTEMPLATE in GETTEMPLATE).
Otherwise,

.gitignore
```diff
@@ -27,3 +27,4 @@ core
 
 # set up by install-diff-filter.sh script
 .gitattributes
+sources/LLREAD.LCOM.~1~
```
masinter, line 30: *~ is already gitignored

library/BIGBITMAPS
```diff
@@ -1,34 +1,38 @@
...
+      |changes| |to:|  (FNS \\DRAWLINE.BIGBM.DASH \\DRAWLINE.BIGBM.NODASH BIGBITMAPP)
+                       (VARS BIGBITMAPSCOMS)
 
-      |previous| |date:| "20-Jan-93 13:40:44" |{PELE:MV:ENVOS}<LISPCORE>LIBRARY>BIGBITMAPS.;10|)
+      |previous| |date:| "10-May-2021 15:37:51" 
+|{DSK}<Users>kaplan>Local>medley3.5>git-medley>library>BIGBITMAPS.;1|)
 
 
-; Copyright (c) 1991, 1993, 1994 by Venue.  All rights reserved.
+; Copyright (c) 1991, 1993-1994, 2021 by Venue.
```
masinter, line 12: new copyright could say Interlisp.org

library/BIGBITMAPS
```diff
@@ -1,34 +1,38 @@
...
-       ((DECLARE\: DONTEVAL@LOAD EVAL@COMPILE DONTCOPY (RECORDS BIGBM)
+       ((DECLARE\: EVAL@COMPILE DONTCOPY (RECORDS BIGBM)
                (CONSTANTS (|\\MaxBitMapHeight| 65535)
                       (|\\MaxBitMapWidth| 65535)
                       (|\\MaxBitMapWords| 131066))
                (MACROS |GetNewFragment|)
                (MACROS |\\SFInvert|))
         (INITRECORDS BIGBM)
-        (FNS BITBLT.BIGBM BITMAPCREATE.BIGBM BITMAPCREATE BITMAPCOPY BLTSHADE.BIGBM BITBLT 
+        (FNS BIGBITMAPP BITBLT.BIGBM BITMAPCREATE.BIGBM BITMAPCREATE BITMAPCOPY BLTSHADE.BIGBM BITBLT
```
masinter, line 24: is there any documentation for these changes or BIGBITMAP itself?

library/BIGBITMAPS
```diff
@@ -65,11 +69,11 @@
 
 (PUTPROPS |\\SFInvert| MACRO ((|BitMap| \y)
 
-         (* |corrects| |for| |the| |fact| |that| |alto| |bitmaps| |are| |stored| |with|
```
masinter, line 68: can we convert these comments to strings? Or use a different readtable than XCL?

library/BIGBITMAPS
```diff
@@ -84,6 +88,10 @@
        '6)
 (DEFINEQ
 
+(BIGBITMAPP
```
masinter, line 91: I suppose these changes are to avoid the cross-file record access. Sorry that just became clear.

rmkaplan:
Is this a change that didn’t get captured in this branch?  I don’t remember doing anything about macroprops or templates.

I did fix masterscope to go to the whereis database as a last resort for CONTAINS queries, so I could find the things I needed to work on.

(Also, I extended DUMPDATABASE to take a FILE argument so that it could dump the database with a define-file-info reader environment that would load in all contexts.)

rmkaplan:
Yes, to export a test that didn’t require the declaration to be present.  (The logic in that code is still a little confusing, I think the bigbitmap extension is wired in at a slightly wrong level.  But I left it.)

lispusers/MODERNIZE
```diff
@@ -1,11 +1,11 @@
-(DEFINE-FILE-INFO READTABLE "INTERLISP" PACKAGE "INTERLISP")
-(FILECREATED "14-Mar-2021 20:33:34" 
-{DSK}<Users>kaplan>Local>medley3.5>git-medley>lispusers>MODERNIZE.;14 20950
```
masinter, line 3: shouldn't this move to Library? Basically anything that is part of FULL.

masinter (changes requested):
The problem with MSMACROPROPS is just an indication that you made your branch and didn't update it from master for months; I wasn't doing a _lot_ of editing but we need to resolve the conflicts (not many).

I've gone through about 50 / 150 of the changed files in about 5 hours. I'll try to go through the rest.

rmkaplan:
Thanks for slogging through this, I know it’s a lot.

In terms of legacy dates, I thought it was set up to do that.  It certainly does that in my LFG environment.  Maybe there is a flag some where that should be set.

In fact, it seems that in LFG I have a file EDITDATEPATCH that has a variable SINGLEEDITDATE that controls that.  Maybe the system edit date isn’t set up that way.

sources/CMLREAD
```diff
@@ -237,38 +240,41 @@ Copyright (c) 1985, 1986, 1987, 1988, 1990, 1993 by Venue & Xerox Corporation.
...
+       '10)
 
 (DEFMACRO WITH-READER-ENVIRONMENT (ENV . BODY)
    `((CL:LAMBDA (E)
             (LET ((*PACKAGE* (ffetch (READER-ENVIRONMENT REPACKAGE) of E))
                   (*READTABLE* (ffetch (READER-ENVIRONMENT REREADTABLE) of E))
                   (*READ-BASE* (ffetch (READER-ENVIRONMENT REBASE) of E))
-                  (*PRINT-BASE* (ffetch (READER-ENVIRONMENT REBASE) of E)))
+                  (*PRINT-BASE* (ffetch (READER-ENVIRONMENT REBASE) of E))
+                  (*EXTERNALFORMAT* (ffetch (READER-ENVIRONMENT REFORMAT) of E)))
```
masinter, line 260: where are the other variables declared SPECIAL? Why just the external format? Is this really at the right level? The other elements don't depend / vary when you change XCCS and UTF-8... they're EXTERNAL-FORMAT independent

sources/CMLREAD
```diff
@@ -237,38 +240,41 @@ Copyright (c) 1985, 1986, 1987, 1988, 1990, 1993 by Venue & Xerox Corporation.
...
-(PUTPROPS WITH-READER-ENVIRONMENT INFO EVAL)
+(PUTPROPS WITH-READER-ENVIRONMENT INFO EVAL)
 (DECLARE%: DOEVAL@COMPILE DONTCOPY
 
 (GLOBALVARS *COMMON-LISP-READ-ENVIRONMENT*)
 )
 
 (RPAQ? *COMMON-LISP-READ-ENVIRONMENT* (create READER-ENVIRONMENT REPACKAGE _ (CL:FIND-PACKAGE
                                                                                   "USER")
-                                                 REREADTABLE _ CMLRDTBL REBASE _ 10))
```
masinter, line 269: same ... common lisp vs. interlisp readtable shouldn't change the "stream" properties.

sources/EXTERNALFORMAT.TXT
```diff
@@ -0,0 +1,70 @@
+New architecture for character input-output and alternative external formats
```
masinter, line 1: we really need to put together a Medley 4 release manual, formatted and fixed up

sources/FASLOAD
```diff
@@ -1,14 +1,14 @@
 (DEFINE-FILE-INFO READTABLE "XCL" PACKAGE "FASL")
-(IL:FILECREATED "17-Apr-2018 07:55:20" 
-IL:|{DSK}<Users>kaplan>Local>medley3.5>lispcore>sources>FASLOAD.;2| 35249  
+(IL:FILECREATED "10-Jun-2021 18:26:43" 
+IL:|{DSK}<Users>kaplan>Local>medley3.5>git-medley>sources>FASLOAD.;8| 35584  
 
-      IL:|changes| IL:|to:|  (IL:FNS CONVERT-FASL-DATE)
```
masinter, line 5: This is really illegible IL:|changes| IL:|to:|

masinter (changes requested):
while there are a lot of "cosmetic" issues (bad pretty printing, edit comments lost, etc) there are a few more serious issues.  I think it's necessary to sort out the "Masterscope gives error on ambiguous record field" changes from the "support external character encoding". The former could be fixed by adding a per-file record priority list, removing the necessity of any source edits. And the READTABLE should be independent of the file encoding. READ works on a stream of characters, which remain the same character independent of whether they are encoded in XCCS, UTF-8, ISO8059-10 or what have you.
I propose using this branch as a model, and bringing forward changes one at a time starting with current master.

sources/FASLOAD
nbriggs: Happens any time the FILECREATED expressions are written/read with a readtable other than an IL based one.  IL:|previous| IL:|date|, etc.

rmkaplan:
With respect to CMLREAD, the other variables are the variables that define the current default readtabe, package, and base, when you call IO functions with those not locally specified.  I presume that there are global declarations somewhere that they are specvars.

I added the external format as another variable, when I thought that it was of the same sort and that was the right thing to do.  But it isn’t, the external format is a property of the stream itself, not of the printing environment.  So that variable is really not interpreted—the external format is installed specifically when the stream is opened or the user changes it by STREAMPROP or SETFILEINFO.

To make it operate implicitly as part of the context set by this macro, a further change would have been to have this reset the format of the stream under a resetsave and restore the previous external format on exit.  But the problem is that this macro doesn’t actually take the stream as an argument, it just sets these variables for any and all printing/reading to any and all streams underneath.

(Note also that setting the format changes all 5 of the primitive character reading/writing functions in the stream. It’s a level below whether or not a given internal character is a sepr in this context, or a package identifier should be printed.)

(Note also:  I have no idea why this was in CMLREAD)

rmkaplan:
Also, the externalformat.txt document reflects an earlier stage of the migration from what I inherited and where I ended up.  I’ll do an update when this settles down.  (After this merge, there are a few minor cleanups (rationalizing a few of the naming conventions, maybe even taking the format variable out of the WITH-READER-ENVIRONMENT) still to do.)

masinter:
Can we at least (a) implement some "unit tests" that makes sure the UTF-8 code is working? Preferably something that can be automated? 
(b) agree to make up "issues" that cover the problems I identified, as a way of getting agreement on what problems you agree need fixing and why? 
I didn't finish reviewing all of the files, and those that diff well (because of pretty printing) are an issue

sources/FASLOAD
masinter: a separate issue, but I think we should change the FILECREATED format to use strings or keywords

   "changes to" IL:READ "previous" "12-DEC-2021"

or

   :changes "IL:READ" :previous-date "12-Dec-2021"

so it isn't so ugly. This is in support of "Modernization" and "better support of Common Lisp".

.gitignore
masinter: now issue #357

library/BIGBITMAPS
masinter: Issues #207 (copyright dates) and (possibly?) #39 (larger screen bitmap)

library/CLIPBOARD
```diff
@@ -203,8 +203,8 @@
 )
 (PUTPROPS CLIPBOARD COPYRIGHT (NONE))
```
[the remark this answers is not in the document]
masinter: See notes in #211

library/POSTSCRIPTSTREAM
```diff
@@ -1113,59 +1111,63 @@ Copyright (c) 1989, 1990, 1991, 1992, 1993, 1994, 1995, 1997, 1998, 2018 by Venu
...
+                                                                   \PS.SCALE0)
+                                                            T))
                 "%%%%Title: "
                 (MKSTRING (OR (LISTGET OPTIONS 'DOCUMENT.NAME)
                               FILE))
                 T "%%%%Creator: PostScript Driver Copyright (C) 1988-1992 Venue and others" T 
                 "%%%%CreationDate: " (DATE)
                 T %# (COND
-                        ((EQ 'LPT (FILENAMEFIELD FP 'HOST))
+                        ((EQ 'LPT (FILENAMEFIELD STREAM 'HOST))
```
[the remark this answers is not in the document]
masinter: #363 moderrnize printing

lispusers/ISO8859IO
```diff
@@ -35,23 +35,10 @@ Copyright (c) 1995-1997, 1999, 2021 by Xerox Corporation.
...
-               (CONSTANTS (NSCHARSETSHIFT 255))
-               
-               (* ;; "From LLREAD")
-
-               (MACROS \XCCSIN \XCCSPEEK \BACKXCCSCHAR)
-               
-               (* ;; "From MODARITH")
-
-               (MACROS UNFOLD))))
+        (DECLARE%: EVAL@COMPILE DONTCOPY (P (OR (GETP 'EXPORTS.ALL 'FILE)
```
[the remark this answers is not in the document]
masinter: #365

lispusers/MODERNIZE
```diff
@@ -92,19 +101,28 @@
...
-        (WINDOWPROP WINDOW 'BUTTONEVENTFN (IF ANYWHERE
-                                              THEN (FUNCTION MODERNWINDOW.BUTTONEVENTFN.ANYWHERE)
-                                            ELSE (FUNCTION MODERNWINDOW.BUTTONEVENTFN))))
+    (CL:WHEN (AND TITLEPROPORTION (GREATERP TITLEPROPORTION 0.5))
+           (ERROR "TITLEPROPORTION cannot be greater than .5"))
+    (CL:WHEN (WINDOWPROP WINDOW 'PREMODERN-BUTTONEVENTFN)
+        (WINDOWPROP WINDOW 'BUTTONEVENTFN (WINDOWPROP WINDOW 'PREMODERN-BUTTONEVENTFN))
+        (WINDOWPROP WINDOW 'PREMODERN-BUTTONEVENTFN NIL))
+    (WINDOWPROP WINDOW 'PREMODERN-BUTTONEVENTFN (WINDOWPROP WINDOW 'BUTTONEVENTFN))
+    (WINDOWPROP WINDOW 'BUTTONEVENTFN (IF (OR ANYWHERE TITLEPROPORTION)
```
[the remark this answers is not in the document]
masinter: issue #367

sources/CMLREAD
masinter: Issue #371

masinter:
no longer needed; was turned to draft after merging to make sure issues raised / tested

PR #349, opened 2021-07-08
Uploaded Xerox Common Lisp Implementation Notes book in chunks

johnwcowan:


masinter:
I'm not sure where this document belongs, and we haven't been consistent, but the documentation available as pdf's seem to be on the web site and perhaps also in the zotero bibliography,

johnwcowan:
I think that everything should be in Medley/docs that is needed to use
Medley as it stands today.  It's parallel (physically and semantically) to
the ReleaseNotes, which gives you the Medley additions to the IRM.

masinter:
[https://interlisp.org/docs/cl-impl-notes.pdf](https://interlisp.org/docs/cl-impl-notes.pdf) as one file, 
with link from [medley wiki](https://github.com/interlisp/medley/wiki).
There's also an HTML export at [https://interlisp.org/docs/cl-impl-notes.html](https://interlisp.org/docs/cl-impl-notes.html), but I'm not sure how usable that is.

masinter:
As confirmed 7/18/21

PR #352, opened 2021-07-22
Add small util for restoring versions from git history

masinter:


rmkaplan (approved):
Checks things out, doesn't delete anything.  Helpful for people who want the help.

PR #374, opened 2021-07-29
Git hub build

stumbo:
GitHub workflow to build Docker image for Medley and store in dockerhub.  Uses Maiko image created by Maiko workflow.


rmkaplan (commented):
I don't think I'm competent to review this, these are technologies that I use but with only peripheral understanding

stumbo:
I've started putting together a document describing how github actions work and detailing exactly what the `yml` file in this commit is supposed to do.  It's still a work in progress, my next step is to document the Dockerfiles.  It resides in the wiki at:
  
- [Building Maiko and Medley with GitHub Actions](https://github.com/Interlisp/medley/wiki/Building-Maiko-and-Medley-with-GitHub-Actions)

Once completed it will provide a basic understanding of the process and be useful when we want to extend or modify it.  Especially since most of us don't do DevOps fulltime -- starting with me.

masinter (commented):
* While we're getting used to this and debugging, it might be better to have manual triggers
 * it would be better to allow one to trigger a build of any branch. 
 * building maiko by `cd maiko/bin running './makeright x' and './makeright init' for the platforms we run on
   * that includes Mac OS 10 on x86 and Mac OS 11 on arm64
 * not sure what the docker maiko images are for 
 * build of medley would do `cd medley` `./scripts/loadup-all.sh` and the `copy-all.sh` `start-release.sh`
   * only need to build medley on ONE platform

nbriggs:
It looks as though it starts from a maiko docker image to build the medley docker image.

nbriggs:
In
> * [Building Maiko and Medley with GitHub Actions](https://github.com/Interlisp/medley/wiki/Building-Maiko-and-Medley-with-GitHub-Actions)

we style it as "Interlisp" (no capital L on lisp).

masinter (commented):
- not sure whether to approve or request changes.
- it looks like you're running X in the container and then tightvncserver, but xvnc contains both and only needs one process.
- it would be good to find a smaller linux distro

nbriggs:
There are certainly smaller Linux distros available as Docker images (Alpine Linux is one such) but I don't know how big they get by the time we add in the packages we need to run (make, C compiler, C development headers, C library, X11 library, X11 development headers, VNC)

masinter:
> packages we need to run (make, C compiler, C development headers, C library, X11 library, X11 development headers, VNC)

You don't need any of those to run, just to build (except xvnc which you need to run in vnc mode)
To run, you could add csh (to run CHAT(SHELL)) and xclip and perhaps a few others

nbriggs:
You don't need make or the C compiler/headers to run, you need the C library and the X11 library.  I wouldn't add csh, I'd just change the CHAT(SHELL) to be my preferred shell (seems to be zsh these days).  For xclip you need which ever package has the X11 utilities.

stumbo:
>     *  While we're getting used to this and debugging, it might be better to have manual triggers
>     *  it would be better to allow one to trigger a build of any branch.

We can do this - or another approach I've used for debugging is having it run on commits to a test branch.  That way the whole process runs but not on the master branch.  Once satisfied it's simple to switch over to the master branch.
 
>     * building maiko by `cd maiko/bin running './makeright x' and './makeright init' for the platforms we run on     
>       * that includes Mac OS 10 on x86 and Mac OS 11 on arm64
>     * not sure what the docker maiko images are for

Building the Maiko Docker image and using it as the base for the Medley Docker image  results in an image that will run on Windows, Linux, Mac and Pi in a *Docker container*.  We can easily extend to any other operating system that support running Docker Linux containers.

The Maiko Dockerfile performs the `./makeright x` but currently doesn't have `./makeright init` as part of the process.  That looks like a shortfall.

>     * build of medley would do `cd medley` `./scripts/loadup-all.sh` and the `copy-all.sh` `start-release.sh`    
>       * only need to build medley on ONE platform

Yes, we've talked about this as being an improvement.  The current approach is non-optimal and dependent on sysouts being stored in git.  Both areas we want to fix.

stumbo:
>     * not sure whether to approve or request changes.

This is one piece of a larger puzzle - and as we build out the whole puzzle will likely change.  As I see it we eventually will have an automated release process that consists of all of the following:

1. Maiko test suite run on each push **Exists**
2. Build Maiko releases for supported environments
3. Medley test suite (this might be part of building the release)
3. Build Medley Release, which consists of building the sysouts and packaging it.  **Script exists to do this from a workstation**
4. Build a Docker image to support Medley on a variety of platforms.
5. Build a Docker image to support Docker in the Cloud (this may be an extension of the above, but that unclear).

Maybe building the Docker image isn't the right place to continue the process, the Maiko test suite exists, but from my perspective it's been useful in helping me understand how Medley is built.  

Personally, I think getting any next piece in place and exercising it is a positive step.  

>     * it looks like you're running X in the container and then tightvncserver, but xvnc contains both and only needs one process.

This is an area that can be improved.  The Dockerfile Abe created used tightvncserver so I kept with it.  I played around with loading other X servers but found nothing that either worked or was supported by the various environments we want to support with Docker.

The whole how to connect to Interlisp running in a Docker Container is something that needs some work.  It's non-obvious and took me a bunch of fiddling to get to work.  I'd like this to be simple and bullet proof, we're no there yet.

>     * it would be good to find a smaller linux distro

I started out using Alpine but found I was adding a lot to it, especially to get VNC working.  I dropped that idea and went with Ubuntu.  I would suggest finding a lighter distro is a good longer term goal.

stumbo:
> You don't need make or the C compiler/headers to run, you need the C library and the X11 library. I wouldn't add csh, I'd just change the CHAT(SHELL) to be my preferred shell (seems to be zsh these days). For xclip you need which ever package has the X11 utilities.

This is one of the reasons to create a Maiko Docker image.  We pull in the tools to do the build, create the Maiko executable and then delete stuff that's no longer needed.  I do a little of that, but there is room for improvement.  When we use it for building the Medley Docker image we get the executable per architecture ideally with the bare essentials needed to run.

masinter:
will create new issue so discussion doesn't get lost when PR is closed.
> > ```
> > * not sure whether to approve or request changes.
> > ```
> 
> This is one piece of a larger puzzle - and as we build out the whole puzzle will likely change. As I see it we eventually will have an automated release process that consists of all of the following:
> 
> 1. Maiko test suite run on each push **Exists**
> 2. Build Maiko releases for supported environments
> 3. Medley test suite (this might be part of building the release)
> 4. Build Medley Release, which consists of building the sysouts and packaging it.  **Script exists to do this from a workstation**
> 5. Build a Docker image to support Medley on a variety of platforms.
> 6. Build a Docker image to support Docker in the Cloud (this may be an extension of the above, but that unclear).
> 
> Maybe building the Docker image isn't the right place to continue the process, the Maiko test suite exists, but from my perspective it's been useful in helping me understand how Medley is built.
> 
> Personally, I think getting any next piece in place and exercising it is a positive step.
> 
> > ```
> > * it looks like you're running X in the container and then tightvncserver, but xvnc contains both and only needs one process.
> > ```
> 
> This is an area that can be improved. The Dockerfile Abe created used tightvncserver so I kept with it. I played around with loading other X servers but found nothing that either worked or was supported by the various environments we want to support with Docker.
> 
> The whole how to connect to Interlisp running in a Docker Container is something that needs some work. It's non-obvious and took me a bunch of fiddling to get to work. I'd like this to be simple and bullet proof, we're no there yet.
> 
> > ```
> > * it would be good to find a smaller linux distro
> > ```
> 
> I started out using Alpine but found I was adding a lot to it, especially to get VNC working. I dropped that idea and went with Ubuntu. I would suggest finding a lighter distro is a good longer term goal.

masinter (approved):
time to get started
the comments were mainly requests for next steps

PR #383, opened 2021-08-01
get exports.all and whereis.hash from medley/tmp when run-medley -new or -newlisp

masinter:
Add option (mainly for Ron) to add tmp to front of directories where exports.all and other newly created sources will come from (when using SIMPLE-INIT). Modified fn MEDLEY-INIT-VARS in LOADUP-LISP.


rmkaplan (approved):
Branching on the location of the source sysout looks right to me, that should give consistency.

PR #394, opened 2021-08-06
Update SYSEDIT to load exports.all when running run-medley -new

masinter:
FILESLOAD has the "feature" of defaulting to the same directory one is loading from.
When DIRECTORIES starts with tmp, sysedit and the 'right' exports.all are on different directories.


rmkaplan (commented):
Looking at the diff, I don't see what changed.

Also, why does SYSEDIT have an LCOM?  It doesn't contain any functions.

masinter:
The change was to use LOAD? instead of FILESLOAD to load exports.all.
Why is there an .LCOM? I don't know. Some kind of "compile everything" consistency?

This patch isn't actually good enough -- if you cd library first and LOAD(SYSEDIT) it will get the old one. Changing DIRECTORIES isn't enough. But for now,be careful.

rmkaplan (approved):

PR #395, opened 2021-08-06
duplicate definition of UNINTERRUPTABLY pp macro

masinter:
There was an old definition of the PRETTYPRINTMACRO for UNINTERRUPTABLY on AINTERRUPT and a new one on NEWPRINTDEF. The old one is bad and the new one is better. This PR just removes the old one (It was just a problem when working with AINTERRUPT loaded).

PR #397, opened 2021-08-06
Revert "duplicate definition of UNINTERRUPTABLY pp macro"

masinter:
Reverts Interlisp/medley#395


rmkaplan (approved):
If this puts us back the way we were, I approve.  We can then figure out how to improve

PR #413, opened 2021-08-16
loadups for latest sources

masinter:
hopefully we can standardize making loadups when pushing to master


nbriggs:
Just ran into a problem trying Steve's issue with the LOOPS loadup -- MAKESYSNAME is expected to be :MEDLEY (or :LYRIC) -- we should probably revert to that.  MAKESYSNAME also goes into the `CL:LISP-IMPLEMENTATION-VERSION` function, which feeds into `CL:SOFTWARE-VERSION` function.   I'm not seeing an obvious place, yet, to hook in the git hash (or tag) that won't possibly break things.   Even the `IL:LISP-RELEASE-VERSION` hasn't changed from 2.0 -- and it's a FLOATP.  I don't know what might depend on that value.

masinter:
do your own loadups for now
```
cd medley
./scripts/loadup-all.sh
./run-medley -new
```

PR #415, opened 2021-08-17
Restore package/reatable eval in define file fino

rmkaplan:
Relatively few changes to restore previous behavior.


masinter (approved):
THINFILES is less useful than before but harmless

are there any test cases for eval in define-file-info?

rmkaplan:
THINFILES is actually quite useful for directories, like tmp/ or my own project directories, that are not synchronized by git.

These minor enhancements were aimed specifically at thinning out tmp/ where things can really stack up when you do as many loadups as I’ve been doing.

The test case for define-file-info is to inspect the environment returned by \DO-DEFINE-FILE-INFO after noticing a file like SEDIT-COMMANDS.  Does the environment have the SEDIT package form?  After LOAD/LOADFROM…, the environment is (CAR (GETHASH (INFILEP ’SEDIT-COMMANDS)  *FILEMAP-HASH*).

PR #424, opened 2021-08-22
add big fonts, move MEDLEYDIR init

masinter:
Once more: add "big" fonts for use with edit files in envos repos
(and move around some 'fontsold' fonts.

While modifying initialization,, moved MEDLEYDIR logic for relative paths to MEDLEYDIR environment variable (and MEDLEY-INIT-VARS) so that it is LOAD?-ed at loadup different times as needed (by SIMPLE-INIT when starting venuesysouts, during loadups)

GREET should no longer error when running lisp.venuesysout


nbriggs:
Is there any particular reason to segregate the larger sizes of some of the common fonts into `big/` and to put the Titan fonts into `other/` -- that just means that you'll need to have extra entries on the DISPLAYFONTDIRECTORIES in order to get at them, when they're really not special at all.

masinter:
I vaguely recall being warned about some data structure/algorithm that was proportional to the square of the number of entries in a single directory (as being one reason for the separation of fonts into 'presentation' 'printwheel', etc in the envos releases).  I didn't want to tackle the 'fontcreate' lookup logic to confirm -- it looked like a mess of patched code when I looked last.

rmkaplan (approved):
Looks reasonable.

PR #439, opened 2021-08-27
use 'less -R' instead of 'more' for better compatibility in lsee

masinter:


nbriggs:
"less -R" only lets ANSI color sequences through, "less -r" lets raw control sequences through.

Could you open a new issue and attach a sample UTF-8 file that gets garbaged by "lsee" ?

PR #440, opened 2021-08-27
Update unicode mapping tables

rmkaplan:
This replaces the Unicode mapping tables for one-byte encodings (ISO8859-x) that John Cowan has indicated are obsolete.  These new tables were downloaded from Unicode.

I also removed the EAST ASIAN tables, which John also reports are obsolete.  The readme file points to the Unicode directory where current versions of those files can be obtained, and also a reference to a document that describes the format of those files.


masinter (changes requested):
i'd rather not include Unicode files in the Medley source tree that say copyright (c) apple (especially ones of recent vintage.) If need be we can make another repo 'unicode-tables'.

And I would rather not have "obsolete" files if there isn't some justification for keeping them.

Are these files needed at run time? Are they used to create some Lisp code? Or tables? How are they used?

rmkaplan:
They would be used if someone wanted to read a file in one of those formats.

The ISO885IO file currently doesn’t look at those files (I think the codes were copied out a long time ago), but on the list is generalizing that to read these codes and pivot through the XCCS to Unicode mapping.  That is, a generic function that creates an ISO8859 external format for any of the variations.

Maybe obsolete isn’t needed in the git world, but I thought we were keeping obsolete versions of other directories.

I don’t know about copyright.  All these files came from Unicode.org

masinter:
I think we should be more careful with copyright -- John was careful to mark copyright owners.
Our LICENSE file is per-repo. I'll want to check with our lawyer (once we have one), but I think anything Venue released. copyright Xerox or Fuji Xerox or Venue or Envos or JDS is OK in the Medley / Maiko repos which are marked MIT License.
I think (but will want to confirm) that LispUsers contributions are OK there.
Other things (especially recent files like outline fonts and unicode tables) may be available for download and distribution but under different license terms. If it is a problem, I'd rather make a new repo called "fonts".

As far as obsolete goes, I'd like to drop the directory after confirming there isn't any use (the GATHER-INFO list of called-but-undefined used global but never set might help.  But you know where you used the Unicode tables.

If you still want to keep them,, I'd prefer  obsolete/pathto/file  rather than pathto/obsolete/file since when I tar up files for release I can leave 'obsolete' behind.
I still don't know from the reply whether / which unicode files need to be part of the release.

rmkaplan:
Move the obsolete to wherever you think it should go, or even just delete it, on the theory that it can always be recovered from git (or my time machine) if anybody ever cares.

Copyright is an issue in general, but is it an issue for things that come from Unicode?  That is, even if there is a file copyright by Apple, by putting it up in Unicode have they made it available for general use?  I assume that everything would get all tied up if that isn’t the case, but I don’t know.

rmkaplan:
Actually, I don’t remember creating “obsolete”, I just assumed from your comments that it was something I must have done on this round.

But in fact that file was created in 1997, we inherited it from someplace. There is no need to retain it.

masinter (approved):
sorry, I misunderstood the origin of the "Obsolete" files.

rmkaplan:
Actually, I think it was just the one file that was historically called obsolete, the obsolete subdirectory is in fact where I put the old stuff now.

I will move those files under the top-level obsolete.

rmkaplan:
But I’ll do that on the next branch, when I also replace the XCCS mappings with Peter’s corrected and augmented data.

Issue #446, opened 2021-09-01
Contents of medley release tar blob isn't sufficient... errors on startup


nbriggs:
If you start from the medley-210831.tgz file, and do ./run-medley you get a couple of errors during startup.

(1) FULL 31-Aug-2021 ...No such medley directory internal/library
(2) it loads medley>greetfiles>SIMPLE-INIT.;1, File created 21-Aug-2021 18:24:59 then reports: "File not found: /home/briggs/medley/sources/MEDLEYDIR.LCOM" and errors out.

nbriggs:
version medley-210901.tgz fails like this:

![Screen Shot 2021-09-01 at 11 06 00 PM](https://user-images.githubusercontent.com/8138841/131791024-a4daf24d-f1aa-4f58-817b-2df31b909896.png)

PR #447, opened 2021-09-02
release more

masinter:
- move badfiles
- fix release to include sources


masinter:
i meant to make the change to the release script separate, but the other change (adding MAKE-PS to internal/library) got added.

It's ready to go

masinter:
unfortunately this PR includes two different changes:
1. fixes to the release process to eliminate errors on GREET
2. initial inclusion (in internal/lilbrary) of a 'MAKE-PS' package to convert all TEdit files in a given directory to postscript

masinter:
will fix #446

nbriggs:
I'll be on this in the AM.

nbriggs (approved):
This looks OK.  I'd note, though, that you can't run the release-medley.sh on macOS as tar there doesn't support --exclude-backups.  If we need it to run on macOS (in the case that github desktop is installed) then --exclude-backups is equivalent to excluding ".#*", "*~", and "#*#"

nbriggs:
There's a spelling error for release in the release notes -- and the description of "the medley release is here..." seems to be attempting to use backquotes in odd ways and failing to get the desired result.

PR #448, opened 2021-09-02
Release tweaks

masinter:
wasn't subsumed by #447 after all


masinter:
it's hard to track what changes are in which branch. But this one edited different files than #447  even though they both have 'release' in their title.

nbriggs:
I'll review this before the end of the day.

nbriggs (approved):
LGTM.

nbriggs:
It would be good though if the commit comments explained more of the reasoning behind the changes (e.g., MSRECORDTRANFLG should not be set to T because... and DFNFLG should be PROP because... when loading system code it is dangerous to replace compiled definitions with interpreted ones).

PR #451, opened 2021-09-02
Move SIGMA font from LispUsers (as part of EQUATIONEDITOR) to fonts/other/c0 where FONTCREATE will find it

masinter:


nbriggs (approved):
Looks good to me.

PR #458, opened 2021-09-05
Recompiled all FDEV creators to install *DEFAULT-EXTERNALFORMAT*

rmkaplan:
FILEIO:  FDEV record now defaults DEFAULTEXTERNALFORMAT to the value of *DEFAULT-EXTERNALFORMAT*, currently :XCCS.

Otherwise, only the compiled files are changed.  The new source versions are an artifact of the fact that I marked all the functions that need to be recompiled as "CHANGED" so that CLEANUP would use the right compiler.  That caused the files to be remade, unfortunately.  (Do we have a top-level compiler function other than CLEANUP that always choses the right compiler
   Block vs. no-block
   Compile vs recompile
   Byte vs. fake vs. dfasl


masinter (approved):
NEVERMIND
No problems; I was seeing changes made by DWIM getting in when Masterscope analyzing definitions.

rmkaplan:
The only thing that actually changed (i.e. that I intended to change) was the FDEV declaration in FILEIO.  I intended just to recompile the functions that create FDEV in all the other files.  But I did that by MARKASCHANGE on those functions so that CLEANUP would choose the right compiler (I don’t think we have a separate interface that does that).  Which created new sources too.  But there were no intended changes in those sources.

nbriggs:
What I'm seeing is a lot of ^F^C changed to ^F^F introducing the comments [green comments became yellowy-green comments :-)], and a few places where the indentation/line breaks changed a little from re-prettyprinting.

Otherwise, I haven't seen any unexpected code changes yet.

nbriggs:
You merged that too soon -- it moves a copy of TEXTOFD and UNIXCOMM from library/ to sources/ -- which it probably shouldn't.

rmkaplan:
Hmm, SAMEDIR is supposed to catch things like that.

rmkaplan:
Actually, the best way to proceed would be to undo this merge and then delete this branch.  The cleanup didn’t produce any DFASLs, so I think I can more easily do the recompiles properly without remaking the sources.

Can one of you do the magic git incantations?

masinter:
once a merge has been committed to master, the only reliable incarnation is to 'revert' the change with another commit and PR.

PR #459, opened 2021-09-05
Add a tar with only loadups for use with repo

masinter:


masinter:
This is another PR where I rebased incorrectly

Issue #457, opened 2021-09-04
OPENSTREAM broken in lisp.sysout for {unix} FDEV - \GENERIC.READCCODE is not an external format


nbriggs:
Trying to run a lisp.sysout, built 3-Sep-2021:
```
(OPENSTREAM "{unix}/tmp/output.txt" 'OUTPUT 'NEW)
\GENERIC.READCCODE is not an external format
```
but it works fine for `"{dsk}/tmp/output.txt"`

nbriggs:
`\DO.PARAMS.AT.OPEN` was called with an output stream on the `{UNIX}` device, OUTPUT, and NIL for the parameters... which resulted in a call to
```
(\EXTERNALFORMAT the-output-stream '\GENERIC.READCCODE)
```

nbriggs:
It's failing because the DEFAULTEXTERNALFORMAT of the UNIX FDev is indeed the atom \GENERIC.READCCODE

nbriggs:
There's another problem, though it's not the cause of this problem -- the FDEV struct in the Maiko C code no longer matches the Lisp FDEV object.

nbriggs:
`\UFS.CREATE.DEVICE` creates an FDEV with wrong info.  Perhaps it was not recompiled with the correct FDEV datatype definition?

rmkaplan:
Yes, I didn't have the full.database at the time that I made this switch, so didn't know what needed to be recompiled.  And then I forgot that I needed to do that. There are a bunch of other FDEV creators that I will recompile.

I haven't seen this because I typically run in the full.sysout which has Unicode loaded, and Unicode says that the default format for UNIX is :UTF-8, so \EXTERNALFORMAT doesn't fall through to get the default format from the device.  But DSK has :XCCS even in Lisp.

\GENERIC.READDCODE no longer exists, deprecated in favor of the even more generic \INCCODE.

My understanding of the maiko/lisp compatibility for streams, fdevs and other datatypes that are known at both levels is that old fields and field descriptors cannot be changed unilaterally, but that fields can be added.  Is that not the case?

nbriggs:
Perfect, thanks.  I recompiled UFS and found it worked but hadn't got to saying so.
[...]

PR #463, opened 2021-09-06
Second recompile fdev #457

rmkaplan:
I hope this is better, just the compiled files.  I also included the fix to FONTPROFILE to regularize the COMMENTFONT in the NS profile, and eliminated a very old FONTPROFILEPATCH that duplicated FONTPROFILE.

Note: the files CLSTREAM and LLDISPLAY have FUNCTIONS, and they had a FILETYPE property of CL:COMPILE-FILE.  But the compiled files were LCOM's instead of DFASL's, and the functions were actually not compiled.  I compiled them with FAKE-COMPILE-FILE that compiles everything but produces LCOMs to match what was there.  Maybe we need to sweep to find files with LCOMs that don't have the right FILETYPE.


nbriggs (changes requested):
This looks much better.  I notice, though, that the font set BIGGERNS has a similar problem to the font set NS that you fixed here.  No MENUFONT 5 or COMMENTFONT 6.

Perhaps just add another commit of the fixed source & LCOM for FONTPROFILE to the existing branch...  OR we could leave it until after this is merged.

rmkaplan:
I committed a new FONTPROFILE.

But also note that something else is a little off, even in STANDARD.  The profiles assign the same number but different actual fonts to DEFAULTFONT and ITALICFONT.  It may be that ITALICFONT is not used by PRETTYDEF but is referenced directly in other contexts, so that the number doesn’t matter.  But it still odd.

masinter:
on my wish list is a way to set "higher resolution" fonts, especially for who-line and the lispx event numbers.

FONTSET isn't right

masinter (approved):
was FONTPROFILEPATCH deleted but not the .LCOM?

rmkaplan:
I don’t have it anymore, it might have been missed on the first commit and then done on a follow-on.

nbriggs:
FONTPROFILEPATCH.LCOM is gone, as it should be:

```
commit 74ea166c38bd5156ac886fab8ab8c1da06f87509
Author: rmkaplan <ron.kaplan@post.harvard.edu>
Date:   Mon Sep 6 15:58:22 2021 -0700

    FONTPROFILE:  Fix NS COMMENTFONT, delete FONTPROFILEPATCH
    
    This reverts commit 2615140ede9117b722f555e6603029f20274dd0e.

:100644 100644 106d0d0... 2cd849a... M  sources/FONTPROFILE
:100644 100644 a58f35d... f1435c5... M  sources/FONTPROFILE.LCOM
:100644 000000 04db1fc... 0000000... D  sources/FONTPROFILEPATCH
:100644 000000 df5b655... 0000000... D  sources/FONTPROFILEPATCH.LCOM
```

nbriggs (approved):
Looks good now.

PR #453, opened 2021-09-03
reset HOME during loadups to avoid (GREET)

masinter:
MAKESYS differs from SYSOUT in insisting that a GREET comes after, with an empty history

Scripts that start with using the LDEINIT to set a different site greeting are subject to finding the user's GREET file anyway.  While this logic is bonkers, the simplest fix is just to set HOME to /dev/null when running a loadup that starts with a MAKESYS'd image ....

PR #464, opened 2021-09-08
changes from #453 redone: NOGREET for run-medley, dummy HOME LOGINDIR for loadups

masinter:
when doing loadups, remove all paths to find the INIT file for the person doing the loadup.
When doing run-medley -nogreet also supply NOGREET as the site init file.


rmkaplan (approved):
Finagling also the logindir seems to have suppressed the personal greeting in the loadup-full.  The personal greeting then does get loaded when the full sysout starts up.  So, seems good.

PR #465, opened 2021-09-11
Add to medley release a tar of loadups-only, for those who want that

masinter:
redo of PR #459


masinter:
@nbriggs MEDLEY-INIT-VARS (in MEDLEYDIR) sets WHERE-IS files to point to MEDLEYDIR/loadups

rmkaplan (approved):

Issue #462, opened 2021-09-06
run-medley -nogreet still greets


masinter:
@rmkaplan reported https://github.com/Interlisp/medley/pull/453#pullrequestreview-747279243

which is related but different problem. The code in GREETFILENAME (on APUTDQ) is the basis for what GREET does.
It was written for a world where there were sites with sysadmins who would manage site-specific initialization (shared by all users) and then individual init files.

(GREETFILENAME T) returns the site init file and it looks at "LDEINIT", {DSK}/usr/local/lde/site-init.lisp, or {DSK}INIT.ext for ext=LCOM, DFASL, LISP

nbriggs:
It's still a useful concept -- when you install a release, the "site init file" should incorporate the information about where the installation was unpacked, where the fonts are, and anything else that needs to be set for every installation. Then your personal preferences should go in the user greet files

masinter:
I agree it's a useful concept. It's the implementation that doesn't fit. It also got tangled with loadup automation.

masinter:
it would be nice if the result would run with starter.sysout and the venuesysouts. So requiring a new version of GREET would make trouble.

What if -nogreet set LDEINIT to medley/greetfiles/nogreet
and set up nogreet to (P (AND (STKPOS 'GREET) (RETFROM 'GREET)))
(need to close the 'nogreet' input stream)

I'm guessing the use case for -nogreet is to do manual loadups ? It was added not so long ago

nbriggs:
Can the "nogreet" system greet file set USERGREETFILES to NIL to avoid user greet files being loaded.

nbriggs:
It seems wrong that if I do:
```
LDEINIT="" ./run-medley tmp/lisp.sysout
```
then `(UNIX-GETENV "LDEINIT")` still returns `"/Users/briggs/Projects/medley/greetfiles/SIMPLE-INIT.;1"` and that if I do
```
./run-medley tmp/lisp.sysout --nogreet
```
then `(UNIX-GETENV "LDEINIT")` returns the string `"NIL"` instead of the atom NIL.

nbriggs:
I believe that in the "nogreet" case in run-medley it should do "unset LDEINIT" rather than setting it to the empty string.

nbriggs:
[...]

PR #466, opened 2021-09-11
Stop-gap NOGREET for use with run-medley -nogreet

masinter:
somehow this file got left out and code in run-medley expects it
#462 has some discussion


masinter:
subsumed by #476

PR #471, opened 2021-09-11
Add NOERROR extra parameter to STKARGNAME, and set it in DEBUGGER

### LINKED ISSUE #469: Illegal stack arg: 2 while browsing stack
While looking into #55 I got an error under  INSPECTW.CREATE

image

In STKARGNAME
Illegal stack arg 2
trying to inspect the LET frame

Easy way to reproduce:
```
(LET(X)(HELP))
```
middle mouse in the titlebar of the HELP window and select BT!
Left-click on the 'LET' frame

you will get the error under

masinter:
fixes #469 Illegal stack arg


masinter:
There's another problem which I thought was related but I think now it's not.
```
(PROG ((X 'TEST))(HELP))
PB X
```
image
I'll raise a separate issue

rmkaplan (changes requested):
I still get this error, with ASTACK created 11-Sep at 12:54:19.

nbriggs:
I just did a loadup with the same one Ron mentioned, and it's working OK for me.  One thing I didn't expect though is the way it's presenting  `"("                    ((X) (HELP))` 


Screen Shot 2021-09-11 at 6 33 53 PM

rmkaplan:
Here’s what I see.  Isn’t it the PB that’s supposed to be fixed?

nbriggs:
This fixes the `BT!` middle-mouse-click then select "LET" in the menu... gets you an illegal arg error.
It doesn't fix the PB error.

rmkaplan:
I see that the BT! was the problem cited in #469, but PB X is mentioned in the PR #471.

I wasn’t sure what Larry was comparing when he set that the PB was a different issue.

I’ll approve it for the BT!, the PB is still open.  (As is the fact that the frame for a broken function is goofy—that’s also a live wire).

rmkaplan (approved):
Fixes looking at a LET frame in a BT! list.  Does not fix PB X.

masinter:
i'm trying to follow the Git workflow of only fixing one thing at a time. I noticed another problem with PB and made it into one PR. It's an awkward workflow for Lisp more so than other languages? That's what I think we should talk about.

Nick noticed a problem that I had seen but didn't make an issue on is the odd display of 
"("     at the beginning. SO I'll mark this PR as draft so it's clear. #472 
Please refer to issues by issue number. It will help avoid confusion.
If you don't know the number, you can search for issue title while typing; start with #.

rmkaplan:
Git workflow is a royal pain in the ass in our universe, an enormous source of inertia.

Suppose you are working on problem P on branch B, and you notice a typo in some function on some file that is unrelated to P.  Correct me if I’m wrong, but wouldn't the “proper” git workflow (which discourages fixing unrelated things in a single branch) be:

1. Switch to Master, clicking (if you use the git Desktop) to stash your current changes.

2. Create a new branch under Master, and think for a moment to create and type in a name for the branch that will be somewhat meaningful in the history.

3. Fix the typo

4. Push it back to the repository.

5. Issue a pull request.

6. Wait for someone to approve the typo-fix.

7. Merge the typo branch back into the Master.

8.  Go back to branch B.

9. Unstash your changes.

(10.  If you want to continue working with the typo fix, merge the new Master into your original branch B…if you still remember what problem P was all about)

Given all of this, maybe you’ll wait until sometime next year to fix that typo.

(And of course, the diff that the approver is supposed to look at is totally inappropriate for our Medley source files, highlighting so much irrelevant stuff that you can’t see the trees for the forest. Even such a simple typo fix would show up as changes to the file preamble, possible date changes and line shifting around the fix, and a new filemap. The diff-approval process can and does find mistakes and glitches, but traditionally those have also shown up with lighter-weight and less rigid protocols.)

The git conventions are very helpful for managing the effects of major surgery (like ripping through all of the IO functions to create essentially a new Medley release), but I think they are overkill for many other Medley-typical scenarios. 

We should be able to construct some easier-to-use tools inside Medley that can highlight real changes in a systematic way and with an intelligent presentation.  E.g. a simple interface to COMPARESOURCES that runs on all files with changes after a specified base-date.  But this would require that previous file versions are accessible to Medley (as they used to be, with normal file servers), which may mean that Medley would have to internalize some ways of interacting with the git substrate (a git file device with a special directory enumerator and a special read-only open-stream?).

masinter:
I'm still tracking down the source of the spurious "(" in the LET frame display, but:

the propertylist of LET has a ARGNAMES property that starts with 
`(NIL #\ (#\( #\{`
in lisp.venuesysout too.

not obvious who sets that or in what file

masinter:
@rmkaplan wrote:
> Git workflow is a royal pain in the ass in our universe, an enormous source of inertia. Suppose you are working on problem P on branch B, and you notice a typo in some function on some file that is unrelated to P. Correct me if I’m wrong, but wouldn't the “proper” git workflow (which discourages fixing unrelated things in a single branch) be:

I think a better workflow than the one you outlined is: 
1. open a new issue describing the typo
2. go back to working on P.  When you're done with P and have submitted a PR that fixes P, THEN you can
3. self-assign the issue you made in step 1
4. checkout master, do new loadups as necessary first, fix the typo etc.

Don't fix problems  for which there are no issues
Work on one problem at a time

nbriggs:
It's not so much avoiding unrelated fixes in a single branch as unrelated fixes in a single commit (though with a squash and merge, all the commits on a branch are going to become a single commit -- but if the author of PR suggests the commits should be kept separate, that's easily doable too (create a merge commit vs. squash and merge).

Branch names are ephemeral -- after a branch is merged it gets deleted, so don't spend any effort coming up with a descriptive branch name except in as much as it will let you remember what you were doing there if you happen to leave it and come back later.

The commit comments, however, will last forever.  When someone is looking through the log that's all they're going to see as a reason the code is being changed -- the issue the change is fixing *might* be traceable back to github.com, but there's no guarantee.  If there's an issue the commit comments should mention the number (e.g., `Fixes/resolves/closes #xxxx`).

masinter:
i'm going to merge this change -- other problems have other issues

rmkaplan:
Are you adding yet another step:  create an issue?

nbriggs:
I am selective about when I create issues-- not for trivial changes that I'm going to fix right away -- but if

- it needs further investigation (by me or someone else)
- I'm not going to work on it right now and I don't want to forget it

PR #475, opened 2021-09-12
Ignore #\( #\{ patterns in 'smart' argnames when showing stack frames

### LINKED ISSUE #473: debugger displays bogus "(" as argument name to LET
I just did a loadup with the same one Ron mentioned, and it's working OK for me.  One thing I didn't expect though is the way it's presenting  `"("                    ((X) (HELP))` 

Screen Shot 2021-09-11 at 6 33 53 PM

_Originally posted by @nbriggs in https://github.com/Interlisp/medley/issues/471#issuecomment-917528737_

masinter:
The ARGNAMES property of special forms like PROG LET LET* have special syntax characters to show more complex patterns.
These aren't useful when displaying a stack frame and and showed up as a spurious "#\(" in some circumstances.

Fixes #473


rmkaplan (approved):
Note that copyright was updated

Issue #437, opened 2021-08-26
run-medley script should not "know" that ../maiko is the only place where lde and friends are found


nbriggs:
I've just wasted hours debugging a loadup issue because "run-medley" was ignoring the fact that PATH pointed to a different maiko tree with updated lde and ldex that was NOT named "maiko".

The scripts should have a default if lde/ldex/ldeinit are not on the PATH but they absolutely must honor the PATH setting if they are found in a different location.

PR #476, opened 2021-09-13
run-medley improvements

masinter:
* if lde is on path, don't override
* if it doesn't find os.machine combo be more explicit (pi4 64 case)
* add a -loadup flag which sets an environment variable
* use -loadup instead of -greet in loadup scripts

### DIFF
--- a/greetfiles/NOGREET
+++ b/greetfiles/NOGREET
@@ -1,0 +1,19 @@
+(DEFINE-FILE-INFO PACKAGE "INTERLISP" READTABLE "INTERLISP" BASE 10)
+(FILECREATED "10-Sep-2021 21:25:42" {DSK}<home>larry>medley>greetfiles>NOGREET.;1 537    )
+
+
+(PRETTYCOMPRINT NOGREETCOMS)
+
+(RPAQQ NOGREETCOMS [(P (COND ((STKPOS 'GREET)
+                                  (SETQ USERGREETFILES NIL)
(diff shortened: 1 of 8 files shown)


masinter:
fixes #462 #437

nbriggs (approved):
It's a step in the right direction.   When we get to installing lde/ldex/ldeinit executables in a standard place, so that the path to the executable doesn't depend on the OS and cpu architecture, we'll be in a better place.  As it stands, if lde/ldex/ldeinit is on the path it will still fail because it's trying to derive the location for osversion and machinetype -- which might be on the path, but it doesn't even try that.

masinter:
As it is now, if command -v $prog succeeds, it doesn't call osversion or machinetype

nbriggs:
Uh, no?
```
silversword% which lde
/Users/briggs/Projects/maiko-testing/darwin.386/lde
silversword% ./run-medley tmp/lisp.sysout
MAIKODIR has no bin: /Users/briggs/Projects/medley/maiko
silversword% git log
commit d6f7ad7de9b8762f2d29a6e8385ec6949bf07a1c
Author: Larry Masinter <lmm@acm.org>
Date:   Mon Sep 13 05:54:31 2021 -0700

    run-medley improvements (#476)

```

masinter:
oops. 

Updated run-medley doesn't bother with MAIKODIR or osversion or machinetype if lde is available on PATH

PR #478, opened 2021-09-13
If lde is on path, don't bother with MAIKODIR

masinter:


nbriggs (approved):
LGTM.  Tested by moving my maiko directory to a different name and putting its darwin.386 directory on the PATH before starting run-medley.

PR #479, opened 2021-09-14
Minor cleanups in LLSUBRS to support new subrs.h for YIELD subr.

nbriggs:
Keep the \INITSUBRS in sorted order: move (YIELD 210) to the end of
the list.

Change the output of WRITECALLSUBRS so that the guard wraps the whole
subrs.h include file.

Avoid using literal tabs in the output #defines, pad with spaces and
arrange the minimum field widths so that everything lines up better.


rmkaplan (approved):
LGTM

Issue #460, opened 2021-09-06
ALTOFONTS are not accessible


rmkaplan:
Event if the altofonts/ directory is on DISPLAYFONTDIRECTORIES, FONTCREATE cannot find strike fonts there.  An example is CYRILLIC.  There is a file altofonts/CYRILLIC10.STRIKE, but (FONTCREATE 'CYRILLIC 10) fails.

Perhaps the files should be renamed to follow the conventions on displayfont directories, in which case it would be something like

c0/CYRILLIC10-MRR-C0.DISPLAYFONT

It may be that some of these fonts are obsolete, and it appears that some of them (perhaps HIPPO and MATH) have been copied to the other directory and renamed.

Should altofonts/ be flushed?

masinter:
I think we'd be better off if we could read the fonts we have. The code in FONTCREATE was a nightmare when I looked earlier, however.

rmkaplan:
We can read them if we rename them or copy them. The names are screwed up

masinter:
more than just the names. I ran a duplicate file detector
a) the fonts>adobe> have odd palatino sizes:
![image](https://user-images.githubusercontent.com/1116587/133336773-1180baef-077e-433f-a5cd-64f397f331ab.png)
odd fonts not copied
![image](https://user-images.githubusercontent.com/1116587/133338090-73c8d21d-7c15-4559-93d7-29b3132c88c3.png)

some the so-called "big" fonts I found in the envos files were fake
![image](https://user-images.githubusercontent.com/1116587/133338581-90a27626-eb02-4861-a66b-238f9edd34db.png)

as are some tiny-numbered fonts (for fine print?)
![image](https://user-images.githubusercontent.com/1116587/133338722-b452e8ca-49ab-472f-8c7c-243c15ea00a1.png)
as were the odd size titan fonts that i'd put into 'other'
![image](https://user-images.githubusercontent.com/1116587/133339552-e3296884-a82a-4df3-858b-48dadb8ff31e.png)
Finally there are compressed .wd.Z files -- all identical
![image](https://user-images.githubusercontent.com/1116587/133340780-5dc4341e-0a4a-4bbb-8ddf-9b0f84addfa0.png)

masinter:
i was looking for a way to compare the fonts from Lisp, #483 came up (see description there)

rmkaplan:
I don’t know what Adobe fonts are, my comment about names was just about altofonts.

rmkaplan:
[...]

PR #482, opened 2021-09-14
Delete fonts files that have identical content in displayfonts/

masinter:
as per #460 -- removed files from 'altofonts', 'adobe', 'other'

only if they had an equivalent in 'displayfonts'

This commit leaves files where comparison is needed


nbriggs:
So in a case such as this:

```
SHA1(./displayfonts/c41/TITAN10-BRR-C41.DISPLAYFONT)= ffb5f6d09ff47aea445e744020cacee3682a2b5d
SHA1(./other/c41/TITAN06-BRR-C41.DISPLAYFONT)= ffb5f6d09ff47aea445e744020cacee3682a2b5d
SHA1(./other/c41/TITAN09-BRR-C41.DISPLAYFONT)= ffb5f6d09ff47aea445e744020cacee3682a2b5d
```
Do you know which size the font actually is (6, 9, 10), and therefore which one should be kept?
Have you checked if the ones you're keeping are the ones that also have Interpress font widths files?

masinter:
The "other" fonts were added by me recently from files found in "Envos".
I think those are dubious (I'm glad I kept them separate) since it seems it was common to substitute font sizes.
The 'right' fix to the problem that motivated this would be to remove the requirement to have displayfonts if you're only opening a text stream to hardcopy it (What You Don't See You Don't Need To Get).

masinter:
will redo this when there are better font tools

PR #490, opened 2021-09-20
TEDIT + FILEBROWSER:  Add function TEDIT-SEE, call from FILEBROWSER

rmkaplan:
I moved the logic for the file browser SEE command to a separate entry function TEDIT-SEE in TEDIT.  This gives a simple command-line way of  bringing up a READONLY TEDIT on a file, and if it's a Lisp source file then you have a scrollable font-change-interpreted view of that file.  

I also corrected the FB Edit and See commands so that they now unshade the top-level Edit/See menu item when the actual selection is a subitem.


masinter (approved):

rmkaplan:
Hmm, it will take awhile to root out things like this.

Maybe putting this inside a CHARCODE, with a charset/char representation would at least let us track them.

PR #501, opened 2021-09-24
Change BOUNDP \EVALVAR so variables on stack bound to NOBIND don't error

### LINKED ISSUE #55: spurious "unbound variable" error when variable is bound to NOBIND (e.g., in MAPATOMS with interpreted fn)
According to the IRM, MAPATOMS takes a function of one argument which is called
with each litatom in the system -- so
```
(LET ((FNCOUNT 0)) (MAPATOMS (FUNCTION
     (LAMBDA (FOO) (IF (GETD FOO) THEN (SETQ FNCOUNT (ADD1 FNCOUNT)))))) FNCOUNT)
```
should return the number of atoms with a definition.   Instead it fails with
```
FOO is an unbound variable
```
If you don't reference the lambda parameter it does indeed call the function 28000+ times successfully.

masinter:
should fix #55 problem with NOBIND in MAPATOMS.

Most of the work is done by STKSCAN opcode.
There is a similar problem in CL:EVAL code which this PR doesn't change.
Need to check ANSI standard if the error condition is specified.
image

(will open a new Issue)


masinter:
I had misunderstood how the STKSCAN opcode worked (like the \STKSCAN function) it returns a pointer to the VAR binding slot in the frame OR the value cell of the symbol if it isn't bound on the stack. It never returns NIL (unlike STKSCAN which returns a (GC'd) datum which contains the stack pointer and should be RELSTKed when done.

I think I need to change BOUNDP and \EVALVAR (and the equivalent for CL specials)  back to asking if the value is NOBIND and then only if it is, seeing if the location is on the stack or a symbol storage slot.

PR #504, opened 2021-09-27
two variables COPYRIGHTFLG and *REPLACE-OLD-EDIT-DATES* set for Lispcore developers

masinter:


rmkaplan (approved):
I thought you were using PRESERVE for the copyrightflg.  Are you changing that to RETAIN, or is that something else?

PR #505, opened 2021-09-28
Externalformat collected in a separate file

rmkaplan:
Mostly, this moves the pieces of the externalformat implementation from LLREAD and FILEIO into a separate small file.  A few other miscellaneous improvements


masinter (commented):
I don't understand the benefit of moving some but not all functionality that :EXTERNALFORMAT entails.
Also :DISPLAY as a value seems odd. In the HPRINT case it was a case of wanting to see what lsee would show?

rmkaplan:
What did I leave out?

Do you want to take out the CRLF EOL convention altogether?  This code is just an updated version of what was always there.

The test case is simple:

(SETQ FOO (OPENSTREAM '{NODIRCORE} 'BOTH 'NEW '((EOL CRLF]
(TERPRI FOO)
(SETFILEPTR FOO 0)
(BIN FOO) -> 13
(BIN FOO) -> 10
(SETFILEPTR FOO 0)
(\INCCODE.EOLC FOO) -> 13
(BIN FOO) ->  End-of-file-error

rmkaplan:
Am I supposed to do something?

masinter:
did you intend to delete FILESETS or is that a mistake?

rmkaplan:
No, git screwed me again.  I’ll figure it out.

rmkaplan:
Is there still a problem?

masinter (changes requested):
i'm not sure what you're doing, but something is awry.
Git has well defined semantics but it's really different than other version systems.

Anyway, this commit is missing sources/LLREAD and .DCOM

masinter:
maybe there's some other way we can do this than Pull Request / Request Changes looping

rmkaplan:
LLREAD and LLREAD.LCOM are correct in my local clone.

But I see that it now says that I have 13 commits that haven’t been pushed, so I pushed them.

I don’t understand that, because I thought I was pretty serious about pushing even at various intermediate states.  And also, there weren’t that many commits for this particular problem, so I think I must have inherited in some way from other branches (e.g. the update from Master option to catch other updates along the way?).

nbriggs:
From the command line, on this branch,  do a “git whatchanged master..HEAD” and look through what’s in the commits, especially for files added or deleted.  If you got commit everything when a tracked file doesn’t exist it will track the deletion. LLREAD was deleted in one of the commits.  (Sent from my iPhone)

rmkaplan:
This is a heavy duty work practice, that’s for sure.

Part of the problem is the convention that commits and branches should be relatively indepependent and relatively small, to make the review process (with it’s Medley-terrible diff) easier for the reviewer.  But the natural way of operating in Medley is to hop back and forth to different and often unrelated issues as they emerge in the course of working on something else.

Breaking these unrelated changes out into branches on the fly gets things muddled.  The possibility (that I have now become aware of) of using the git cherry-picking machinery to move somewhat related changed files into a new branch (descending directly frm Master) may reduce the likelihood of error. Basically, after the fact I can move files with related changes into separate commits, and then move those commits into newly created branches.  And never leave my working branch, never stash and unstash etc., never get confused that a branch switch has changed the local file system out from what I thought I was working in.

The problem, abstractly, is that Medley is a residential system that doesn’t slice up the environment the way git likes to do.  Instead of Medley running inside a git repository, the git repository should be “running" inside Medley.

A lightweight first approximation to that, that I talked about today, would be to fake up a git file device that reveals some of the git branch organization and other-branch files as ordinary Medley files and perhaps even directories. The implementation would be some (hopefully simple) ShellCommand calls that simulate the primitive file operations (infilep, openfile (= git show to copy to someplace out of the clone), directory).  Initially enough to see the current state of a branch, not necessarily to see the version history of files through a sequence of commits (although an implementation of the version hack inside Medley might enable that).

masinter:
What everyone says is that there is no substitute, no shortcut to learning the git model.

Especially if you aim to extend it. 

There are relations between commits and files that lead you to bad places when you use GitHub Desktop to switch branches,

 

 

 

--

 <https://LarryMasinter.net> https://LarryMasinter.net  <https://interlisp.org> https://interlisp.org

masinter (approved):
I don't understand what you are trying to do with these moves, does it make something simpler?
 Will it be clearer after a merge?

rmkaplan:
I’m trying to get all of the external format implementation into one place, for better understanding and future updates.  In the legacy set up, this was spread out between LLREAD and FILEIO and others, not to mention the open coded NS macros.

Issue #483, opened 2021-09-15
HPRINT doesn't EOL correctly


masinter:
I was looking for a way to compare two fonts (one read from hevetica12.strike in AltoFonts and one from helvetica12-mrr-0 in displayfonts>c0>.  After patching \fontfilename.old (and *display-font-name-map* which looks very dossy), adding 'strike to displayfontextensions, setting displayfontdiectories in various orders

I was able to read both using (\readdisplayfontfile 'helvetica 12 'mrr 0 'display 0)

I thought I would try HPRINT to see any diffs, but HPRINT produces output like this

![image](https://user-images.githubusercontent.com/1116587/133355030-fff01893-452c-4895-90db-ef9edd463372.png)

Notice the odd linebreaks when printing the bitmap.

masinter:
clue: if HPRINT is printing to a non-random-accessible stream, it opens a {NoDirCore} stream, prints to that, and then copybytes .

masinter:
pr #506 fixes the EOL problem but the other problems (how to compare font files? HPRINT on DEFAULTFONT fails) remain

PR #506, opened 2021-09-28
Use COPYCHARS instead of COPYBYTES when HPRINT is copying from NODIRCORE buffer

masinter:
I couldn't figure out how to set the NODIRCORE file's :FORMAT or end-of-line convention (or rather, get the current output stream values).

Note SPECIAL declaration needed but missing? Might be a problem when recompiling other files.

Note filetype is CL:COMPILE-FILE which it seems to use but produces an LCOM.

Finally it is still the case that (HPRINT DEFAULTFONT) fails to print an arrayblock, but that is (or will be) a separate issue.
#483


rmkaplan:
With respect to format, you should open the nodircore stream with the format of the destination file.  That should be a parameter `((FORMAT , (STREAMPROP destinationstream ‘FORMAT)).  If that doesn’t work, well, something else is not as expected.

But then, you don’t want to close the nodircore and open it again, because the format is not stored in such a file (it doesn’t have a define-file-info).  You would happen to open it for input with the same format specification.

Or,  better would be to open the file ‘BOTH ‘NEW and set the fileptr back to 0.

Is the SPECIAL with respect to *STANDARD-OUTPUT*, whether that is globally declared as special someplace?  Otherwise, presumably has to be declared here?

Note that we have a general legacy confusion now about whether parameter-names for Interlisp functions are in the keyword package, in the Interlisp package as is the legacy case, or in both as a transition to a keyword end-state.  I used the legacy IL for STREAMPROP.

masinter:
i changed HPRINT to use :FORMAT of standard output in NODIRCORE file; should it delete it?
The problem with raw arrayblocks remains -- will open separate issue.

rmkaplan:
You shouldn’t have to delete NODIRCORE, if you don’t hold onto it, it will get collected.

PR #509, opened 2021-10-01
Add -noscroll option to run-medley; turns off scollbars

masinter:
use BEFORE -dimensions to not add a margin.


nbriggs (changes requested):
If -noscroll must come before --dimensions to work correctly then it should be before --dimensions in the Syntax: comments too.

nbriggs (approved):
LGTM now

Issue #508, opened 2021-09-30
VTCHAT attempts to BOUT a string rather than a number


nbriggs:
VTCHAT.STATUS calls BOUT with a string rather than a number representing the byte to be output, which results in an arg-not-noncomplex-number error... I believe it should wrap the (MKSTRING ...) in a (CHARCODE ...) in order to generate the correct escape sequence.

```
                   (6                                        (* ; "Host wants cursor coords")
                      (BOUT OUTSTREAM (CHARCODE ESC))
                      (BOUT OUTSTREAM (CHARCODE %[))
                      [BOUT OUTSTREAM (MKSTRING (ADD1 (IQUOTIENT (ffetch (CHAT.STATE YPOS)
                                                                        of CHAT.STATE)
                                                             (ffetch (CHAT.STATE FONTHEIGHT)
                                                                    of CHAT.STATE]
                      (BOUT OUTSTREAM (CHARCODE ;))
                      [BOUT OUTSTREAM (MKSTRING (ADD1 (IQUOTIENT (ffetch (CHAT.STATE XPOS)
                                                                        of CHAT.STATE)
                                                             (ffetch (CHAT.STATE FONTWIDTH)
                                                                    of CHAT.STATE]
                      (BOUT OUTSTREAM (CHARCODE R)))
```

nbriggs:
Nope... wrong, the string may well be more than 1 character.

PR #510, opened 2021-10-01
Fix long-standing error wherein VTCHAT.STATUS attempts to BOUT a string

nbriggs:
In replying to a request for the cursor position, the VTCHAT.STATUS
code attempted to construct the escape-sequence response passing a
string representing the X (and Y) coordinate as text to BOUT rather
than using PRIN1.


nbriggs:
Closes #508

rmkaplan (approved):
It isn't clear from the diff where the change was made.  But it is obviously the right thing to do.

nbriggs:
> It isn't clear from the diff where the change was made.

Sorry, it's line 114 et seq. and 118 -- it used to (BOUT OUTSTREAM (MKSTRING ...)) and is now (PRIN1 (MKSTRING ...) OUTSTREAM).  All other changes are from prettyprint reformatting.

Issue #511, opened 2021-10-01
UNIXTELNET should use "ssh" rather than "rlogin" to chat to other hosts


nbriggs:
Very few modern systems are configured with "rlogin" capability, preferring the more secure "ssh".

UNIXTELNET provides a "chat" (terminal window) to another host, when on Maiko systems, by executing "rlogin" to connect to the remote host.  This can trivially be replaced by "ssh".

masinter:
added UNIXTELNET to be loaded by UNIXCHAT.  pr #586 
Considered merging the two files, but left it.

PR #512, opened 2021-10-01
Replace chat via "rlogin" with chat via "ssh"

nbriggs:
Modern systems are unlikely to be configured with "rlogin" access,
remote login, if available, is likely to be via "ssh", so use that.


nbriggs:
Closes #511

rmkaplan:
UNIXMAIL is in library.  Why isn’t it in library/lafite with the other lafite stuff.  Is it independent of Lafite?

rmkaplan (approved):
There is now an extra space in the RPAQ? diffs, I wonder why.

Otherwise, LGTM

nbriggs:
> There is now an extra space in the RPAQ? diffs, I wonder why.

The prettyprinting seems to have added font-change characters around the variable names which were not previously present, and added a space after them before the closing ")" -- don't know why.

> UNIXMAIL is in library.  Why isn’t it in library/lafite with the other lafite stuff.  Is it independent of Lafite?

Unrelated to this change, I think?

PR #513, opened 2021-10-01
Lafite cr to lf

rmkaplan:
Lafite files were in a subdirectory and didn't get systematically converted to LF from CR.  Unicode also seemed to regress.  These were fixed by MAKEFILE NEW.  LAFITETEDIT had a bad file reference, UNIXMAIL and MAILSCAVENGE.TEDIT were in the wrong places


masinter (approved):
I'd like some help with reviewing lafite files in the Envos repository.

The goal in this phase of restoration would be to establish a working baseline.
Not so much for reading mail as it would as a tool to convert old lafite mail archives.

nbriggs:
Please don't merge this -- it needs a bunch of things removed.

nbriggs:
> Not so much for reading mail as it would as a tool to convert old lafite mail archives.
Lafite uses "*start*" to delimit messages, followed by a line with two number and a set of flags.  The 2nd number is usually 24 and is the offset from the *start* to the beginning of the message headers, and the 1st is the length of the message (I don't remember what's included in the count).  Other than that, they're pretty much RFC822 format messages with header/body.  I wrote various awk/sed scripts at various times to process them into exactly the Unix mbox format so  I could import then import them into whatever other mailer I was using.

rmkaplan:
What needs to be taken out?

nbriggs:
> What needs to be taken out?

Did you really intend the `TEDITSCREEN \DISPLAYLINE` and `TEDIT-SEE` to be part of this?

nbriggs:
On second thought -- if you think those commits are good, then it doesn't matter to me whether they come together as part of this PR or not.  I think we should do a merge, though, rather than a squash-and-merge, so that the commits are distinct in the final result.

rmkaplan:
Those may already be in master, I can’t check right now

PR #521, opened 2021-10-13
Improve tedit interaction with modernize

rmkaplan:
Changes mostly to improve interaction of split windows with MODERNIZE, minor adjustments for formatting


rmkaplan:
Updated to add another explicit fontchange to TEDIT-SEE

PR #523, opened 2021-10-13
Lispusers packages modernize, thinfiles, tedit=pf=see

rmkaplan:
Fixed MODERNIZE to work better with TEDIT, THINFILES works on file names, not just extensions.  TEDIT-PF-SEE provides commands tpf <fn> and ts <file> that show functions/files in read-only Tedit windows.  Includes GET-TYPED-WINDOW to reuse regions for windows of a previous type (like PF or SEE).  Also throwing in an update to INSPECTCODE to use the default (fixed pitch) font.


masinter:
> INSPECT: INSPECTCODE starts with DEFAULTFONT (presumably fixed pitch)

When does INSPECTCODE use any other font?

rmkaplan:
Could be converted to Tedit, it just started this way.

PR #526, opened 2021-10-16
Move material from 'release-notes' to README

masinter:
This material needs more attention.  I think the barriers for new users are just too high.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -1,0 +1,16 @@
+# How to build a medley release
+
+Originally done only with shell scripts:
+
+./scripts/loadup-all.sh 
+./scripts/loadup-and-release.sh
+
+
+# Using github actions
+
+In the github medley repository (Interlisp/medley) go to the Actions tab. 
+
+It should list the available github actions, select the bottom one, Build Medley Release. 
+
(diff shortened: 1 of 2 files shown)


masinter:
would the instructions on getting started work for you?

skaisler1:
Larry:

1, Uploaded MedleyInterlisp-InteractiveProgrammingEnvironment.pdf to docs on Interlisp.org

2. Tried the gh command on nedley 211015.
a. Did not work for me.
b. had to sudo apt-get update
c. then had to sudo apt install gitsome
d. both succeeded
e. then tried the gh release download -R Interlisp/medley -p "*"
f. error message: No such command "release"
g. tried again with just "download"
h. error message: No such command "download"

I also tried:
sudo git clone https://github.com/Interlisp/maiko.git
cd maiko/bin
sudo ./makeright x

This worked.

I also tried:
sudo git clone https://github.com/Interlisp/medley.git
cd medley
sudo ./ruun-medley

which worked fine.

One thing I would suggest is that on the Medley Interlisp line to put the current version there so we know what we are running.

I'll do more tomorrow.

Steve K.

masinter:
In reaction I tuned the README in the last push to releases-readme branch.
please re-review

PR #529, opened 2021-10-17
Eliminate implicit calls to \FILEOUTCHARFN

rmkaplan:
Also, update DATE to modern readtable (don't know what it does), add LLETHER to EXPORTFILES in FILESETS (may also need the file that exports pup records).


rmkaplan:
I saw that, could have done it.  But then, I don’t know what this is trying to do (even though it appears that I wrote it).  But you’re right, I might as well compile it.

rmkaplan:
Interesting, could be that DATE was part of the TMAX suite of image objects, maybe it should be deprecated to obsolete.

nbriggs:
Yes, DATE says it is part of the TMAX files -- what's wrong with bringing that forward?  The DATE image object is actually useful -- unless we have a different implementation of it.

masinter:
But there's TMAX-DATE

rmkaplan:
There is a separate TMAX-DATE file which overlaps in various ways (but not completely).  My bet is that DATE was a first implementation (it has an earlier original date) and that the functionality was moved into the TMAX framework.  Confusing to have both.

rmkaplan:
BTW, TMAX-DATE also has a Y2K bug.  I might try to fix that, FWIW.  It thinks the current year is 1921.

nbriggs:
Re: DATE -- yeah, TMAX itself loads TMAX-DATE, so I suspect moving DATE to obsolete is reasonable.  I wouldn't junk TMAX though.

rmkaplan:
No, TMAX is very good.

rmkaplan:
Yes, I spend about half my time trying to figure out how to push trivial stuff back to git, in a way that is semi-reviewable.  It never works out—the models are incompatible.

Issue #514, opened 2021-10-05
Medley docker images no longer functional 


lellaninc:
Medley docker images are no longer functional - ever since loadups were gitignored (in late-August).

"docker run -p 5900:5900 interlisp/medley:latest" returns immediately with error of "Couldn't find a sysout to run".

If you "docker run -it --entrypoint bash interlisp/medley:latest", you can see that full.sysout is not in /app/medley/loadups.

Cause is that the Dockerfile just copies the github repo Medley directory into the image and the loadups (incl. full.sysout) are no longer included in the github repo.

Only tested this on ubuntu.  But 99% certain will be true on all platforms.

Solutions:
1.  Change the Dockerfile to run loadup scripts as part of creating image.
2.  Make docker image builds part of the release process and add copying loadups into the image to the Dockerfile.

stumbo:
There are several layers to this problem and the long term solution is probably an overhaul of how we create Docker images.

The first problem that needs to be resolved is the Maiko docker image currently doesn't contain ldeinit.  Which is needed to build the loadups within the Medley Docker image.  My first brute force idea was to just build it with each architecture - that failed for Pi - I need to investigate whether that's a limitation or an error on my side.  Assuming it's an error on my side, I'll fix it and have the tools in the Maiko image to build loadouts when we construct the Medley image.  

This however is inefficient, building a new loadup for each architecture within the Medley docker image -- I'm assuming I can create one loadup and use it for each architecture.  Getting to this means reworking the docker build process.  

Eventually this should all tie in with how we do  Git releases - the whole process should be a chain of GitHub actions.

nbriggs:
The maiko build for docker needs to do (from bin) `./makeright init` as well as the `./makeright x` that it currently does.

It needs to build a set of lde/ldex/ldeinit for each architecture, but the sysouts it builds/packages can be done once, in a native environment that is fast.  It doesn't matter whether the sysouts are built on a 32-bit or 64-bit system.

masinter:
Looks good running in docker image on my computer.
I still need a kind of "run-medley" for back-end "docker" which will open the docker image, start vnc viewer, mount my LOGINDIR
for people who want to give access to only a medley-specific area and not all their files

PR #534, opened 2021-10-21
Build loadup

stumbo:
Adds new manually triggered github action to build Medley.  This starts to address #514.

The workflow is manually triggered and allows input of a release tag (release name).  If no name is given it uses the current form: medley-YYMMDD.

Currently, this action builds Maiko and uses it to create the loadup files.  I plan to replace this with pulling the latest Maiko release and using the prebuilt files.  

To complete the work I need to pull the built asset into the Docker container.  That will be a separate PR.


masinter:
how do I manually trigger a release?

PR #541, opened 2021-10-24
change editor interactions with MARKASCHANGED to reset EDITMODE when invoking editor to revalidate

masinter:
- change editor interaction with file package in MARKASCHANGED.
- Update SEDIT::MARKASCHANGED to reset(form) EDITMODE to SEDIT when updating


masinter:
I downrated comments that weren't about this PR

masinter:
> I don't think it's a good idea to change the style from lowercase to uppercase. SEdit style (in all the SEdit related files) is lowercase.

@nbriggs would dealing with https://github.com/Interlisp/medley/issues/579 and remaking the file solve the "case" problem?

nbriggs:
> would dealing with #579 and remaking the file solve the "case" problem?

I think that would produce the results I expect.

masinter:
I updated this PR to remove the change to BACKGROUND-YIELD (and LCOM). I did this via
```
git checkout lmm1
git checkout master --  lispusers/BACKGROUND-YIELD*
git commit -m "Undo BACKGROUND-YIELD changes -- already merged"
git push
```

nbriggs:
How about doing
```
git checkout lmm1
git checkout 381c605a0f27df0c77af9938d54671795383e20f
git push origin HEAD:lmm1 --force
```
so that we don't have the history of having background-yield changed and then unchanged.

nbriggs:
That would leave you with:
```
% git whatchanged d1fe834e6f9e01cf40f107807199099730ecee10..
commit 381c605a0f27df0c77af9938d54671795383e20f
Author: Larry Masinter <LMM@acm.org>
Date:   Wed Oct 27 10:14:41 2021 -0700

    Set LCASEFLG in SYSEDIT to not lower-case if then and ISOprs

:100644 100644 8cddbfb... 1052281... M  library/SYSEDIT

commit 4056ccb930c5baacc61c8a4d6fe94efc8df44c63
Author: Larry Masinter <LMM@acm.org>
Date:   Wed Oct 27 10:12:28 2021 -0700

    Put back upper case i.s.oprs

:100644 100644 a335723... e08fbd7... M  sources/SEDIT-TOPLEVEL
:100644 100644 ee444bd... c61f3bc... M  sources/SEDIT-TOPLEVEL.DFASL

commit 22c2618255b81b3477f58d4c921c1d365aad8b57
Author: Larry Masinter <LMM@acm.org>
Date:   Sun Oct 24 10:51:53 2021 -0700

    Update SEDIT::MARKASCHANGED to reset(form) EDITMODE to SEDIT when updating

:100644 100644 7149cf2... a335723... M  sources/SEDIT-TOPLEVEL
:100644 100644 ef9de85... ee444bd... M  sources/SEDIT-TOPLEVEL.DFASL

commit 8dd8119051f14f73169d0565567e47acb88214dd
Author: Larry Masinter <LMM@acm.org>
Date:   Fri Oct 22 19:10:49 2021 -0700

    change editor interaction with file package in MARKASCHANGED.

:100644 100644 04fd688... 7149cf2... M  sources/SEDIT-TOPLEVEL
:100644 100644 5e29545... ef9de85... M  sources/SEDIT-TOPLEVEL.DFASL
%
```

nbriggs:
It's good on the commits to deal with BACKGROUND-YIELD -- but still has the SEDIT-TOPLEVEL text in uppercase instead of the original lowercase.

masinter:
There is no good way to make a new version of SEDIT-TOP-LEVEL with minimal diffs.
The file was already MAKEFIILE'd with different settings of PRETTYFLG (setting which controls whether to prettyprint or not) and different settings of CL:*PRINT-CASE* (the same symbol as IL:*PRINT-CASE*), which controls printing of symbols as :UPCASE :DOWNCASE or :CAPITALIZE (only when the READTABLE in effect is case-insensitive).   

I looked into extending MAKEFILE-ENVIRONMENT to also have a :CASE or :PRINT-CASE keyword and to use that could control this, but the result wouldn't reduce the problems for SEDIT-TOPLEVEL.

And READER-ENVIORNMENT isn't easily extensible (it uses a DATATYPE and I was reluctant to start messing with BOOTSTRAP)
I'll make another branch to show. see #589

rmkaplan:
The Medley git-compare that I have been using may not see those cosmetic differences as differences.  It uses COMPARESOURCES, and that operates on the s-expressions after they have been read according to the reader environment.  So, immune to font-change differences, pretty-printing, capitalization due to the readtable, etc.

I’m trying to fix it so that the summary differencences that it shows in the typescript can be set up as image objects, so that you can click through to see the different expressions in side-by-side sedit windows, etc.  COMPARESOURCES isn’t quite modularized so that it is easy to switch from printing to object-insertion, I’ve been trying to find the right interface level for that switch.  

Where are the SEDIT-TOP-LEVEL files that you want to compare?  I’ll show you how it reports what it thinks has really changed.

masinter:
The comparison is in the comments for PR #541

I ran compare sources and showed the output in https://github.com/Interlisp/medley/pull/541#issuecomment-953198982

rmkaplan:
The comparesources here seems to be what you want, it shows a change just to the markaschanged function.  (The define-file info changes are showing up too, because the non-default base 10 is being put out.  Maybe that should be suppressed.)

So, if this change is correct, I'm not sure why this isn't approved and merged.

The one systematic variance that COMPARESOURCES currently doesn't ignore is the clisp lowercasing. (if → IF and IF → if).  I can fix it so that it doesn't flag case differences in CLISPWORDS.

masinter:
you can convert Issues to Discussions but not PRs.
Long discussion here, but I want to close this PR since I redid the SEDIT::MARKASCHANGEDFN edit as #597

PR #542, opened 2021-10-24
Change WHEELSCROLL constants from LEFT,RIGHT etc to \WSLEFT etc

masinter:
constants are global, and the constant declaration after loading WHEELSCROLL PROP interferes with compiling other code.

I think this might argue against compiling anything in a loadup with unrelated files LOAD(... PROP) because of unwanted CONSTANT or SPECIAL declarations?
Until we're sure this anomaly doesn't affect behavior


masinter:
@rmkaplan how long should one wait for a "hold off" request?

rmkaplan:
I fixed this, and a bunch of other things.  But I haven’t had time yet to work out all the git commands for the new, safe protocol for sending things back from medley, doing commits, PR’s etc. (much more straightforward to pulling from git to medley).

rmkaplan (approved):
This may already have gone in, if let it merge.  I will work my other changes off of this

Issue #496, opened 2021-09-22
The README.md for each directory should explain what files are what


masinter:
The README.md for each directory should explain what files are what

@stumbo wrote:
What about *.venuesysout?  
```bash
-rw-r--r--  1 wstumbo wstumbo 11501056 Jul 18 22:22 full.venuesysout
-rw-r--r--  1 wstumbo wstumbo  8862208 Jul 18 22:22 lisp.venuesysout
```

Both of these files are in the `loadups` directory.  A poorly named sysout?  An artifact of value, an experiment, or something else?  

We may also want to consider setting a minimum size for objects we store in lfs.  The import command supports the `--above` option to set a minimum size.

_Originally posted by @stumbo in https://github.com/Interlisp/medley/discussions/102#discussioncomment-1024617_

PR #545, opened 2021-10-24
update READMEs and BUILDING; move NOXNSPATCH and extra files in greetfiles

masinter:
- update READMEs and BUILDING, move out some unused files
- update loadups/README.md

Toward #496 .... we need to clean out more


nbriggs (approved):
Seems alright.

PR #556, opened 2021-10-28
ACCESSFNS VCELL had bogus computation

### LINKED ISSUE #502: anomaly between \STKSCAN and STKSCAN opcode
In trying to track down #55 I ran into an odd disparity between \STKSCAN the Lisp function and the STKSCAN opcode 
even though calls to \STKSCAN are compiled into the opcode
image

It would be good to find documentation on the "bigatom" changes (more than just comments in the C code.

\STACKSCAN(X) when X is unbound should return a pointer to the valuecell of X.
Same with \STACKSCAN(XXX) although XXX is a "new" atom and "X" is an old one.

See also PR #501 comments

masinter:
fixes #502


rmkaplan (commented):
LGTM

rmkaplan (approved):
LGTM

PR #557, opened 2021-10-28
remove useless (BLOCK) from MSANALYZE

masinter:


masinter:
what shouldn't happen is this pull

PR #563, opened 2021-11-10
Update to use Maiko release artifacts

stumbo:
- Use latest Maiko release binaries to build Medley loadups. 
- Set release to DRAFT


masinter (approved):
not sure how to test this except to approve and merge

PR #568, opened 2021-11-15
Change init to MEDLEYDIR-INIT (suitable for run-medley configs) and added BACKGROUND-YIELD

masinter:
BACKGROUND-YIELD was too high; fixed that. (caused massive delay of gather-info).

The idea is that we'll have different INIT files for different situations (DOCKER-INIT etc). And SIMPLE-INIT wasn't simple.


rmkaplan (approved):
LGTM

PR #581, opened 2021-11-25
ron-1:  a number of little cleanups in various places

rmkaplan:
No biggies


nbriggs (approved):
I see no problems here.

PR #582, opened 2021-11-25
Rmk2

rmkaplan:
A few more minor updates


nbriggs (commented):
I'm wondering if it was the way it was so that one would notice that the compiled file date was the same as (different from) the source file date as a hint that they were (were not) a matched pair?

rmkaplan:
> I'm wondering if it was the way it was so that one would notice that the compiled file date was the same as (different from) the source file date as a hint that they were (were not) a matched pair?

That could have been the rationale.  But then, that would have been only one of the reasonable use cases. Maybe there should be (have been) a separate entry that more clearly lets you ask for the kind of date you want--date of this file, or date of the source file if this is a compiled file ((→ NIL if this is not a compiled file).  But I have patched around it for now.

masinter (approved):
LGTM

PR #586, opened 2021-11-27
UNIXTELNET was needed for SSH change to work

masinter:


nbriggs (commented):
UNIXCHAT gives you `(CHAT "SHELL")` without rlogin/telnet/ssh, but if you load UNIXTELNET then `(CHAT "anything-but-SHELL")` will give you ssh to that.  Not clear to me that UNIXCHAT *depends on* UNIXTELNET...

PR #587, opened 2021-11-28
Rmk3

rmkaplan:
Improved Tedit menu scrolling, updated filemaps


sources/CLSTREAMS
```diff
@@ -1,104 +1,94 @@
 (DEFINE-FILE-INFO PACKAGE "INTERLISP" READTABLE "XCL" BASE 10)
-(FILECREATED " 3-Apr-91 15:11:53" |{PELE:MV:ENVOS}<LISPCORE>SOURCES>CLSTREAMS.;4| 54013  
 
-      |changes| |to:|  (FUNCTIONS CL:WITH-INPUT-FROM-STRING)
+(FILECREATED "27-Nov-2021 13:30:46" 
```
masinter, line 3: can't tell what changed in CLSTREAMS (besides re-pretty-printing with a different fontset.

rmkaplan:
The filemap.  Now includes FUNCTIONS.

(LOAD x ‘PROP)
(MAKEFILE x ’NEW)

Should be done eventually to all the files containing FUNCTIONS

masinter:
Isn't there a function that rewrites the  FILEMAP without re-pretty-printing all the files?

I'm worried about e.g., some PRETTYPRINT macro being buggy and winding up with different code than intended.
I'm not sure all of the Lisp sources will survive the translation.
If a function has an EXPR definition (not compiled) and you LOAD(... PROP) with a new definition, won't MAKEFILE(... NEW) get the old definition instead of the new?

rmkaplan:
There is function UPDATEFILEMAP.  Is that what you are thinking of?

Dated bvm 1986, with a comment 

(* ;;; "Writes new FILEMAP on file currently open as STREAM.  If we return T, the stream has been closed. 
This has little hope of working any more.”)

I don’t know of a function that creates a new file map, other than when (re)printing.

masinter (approved):

PR #589, opened 2021-11-28
Makefile SEDIT-TOPLEVEL NEW, with *PRINT-CASE* set to :DOWNCASE

masinter:
alternative to #541 for review


masinter:
I think we need another issue to cover reprettyprinting all the sources, but consistently so they're not so ugly.

PR #591, opened 2021-11-29
Migration from Interlisp to CL format

Anzus:
Tool for translating File Manager format files to Common Lisp format


masinter (approved):
new code!

PR #592, opened 2021-12-01
change to keywords in FILECREATED expressions

### LINKED ISSUE #372: change FILECREATED expressions to use strings instead of IL: symbols
e.g., in sources/FASLOAD 
This is really illegible IL:|changes| IL:|to:|

 
@nbriggs nbriggs 5 days ago Member
Happens any time the FILECREATED expressions are written/read with a readtable other than an IL based one. IL:|previous| IL:|date|, etc.

 
@masinter masinter 4 days ago Member
a separate issue, but I think we should change the FILECREATED format to use strings or keywords

"changes to" IL:READ "previous" "12-DEC-2021"

or

:changes "IL:READ" :previous-date "12-Dec-2021"

so it isn't so ugly. This is in support of "Modernization" and "better support of Common Lisp".

masinter:
```
IL:|changes| IL:|to:|  => :changes-to
and similarly for previous-date:

fixes #372


nbriggs (approved):
LGTM.  I guess if one is writing a file with an XCL readtable and we're doing lowercase output it might lowercase the keywords, but that should be OK.

Issue #404, opened 2021-08-09
Lost INITVAR for FB.ICONSPEC in FILEBROWSER


masinter:
I went to shrink a filebrowser window and I got an error which turned out to be FB.ICONSPEC wasn't bound.
Using 
```
cd library
rm FILEBROWSER.~*~
../scripts/restore-versions.sh FILEBROWSER
```
and poking around, I see that the value of ICONSPEC disappeared between version 8 and 25
```
(DEFINE-FILE-INFO READTABLE "XCL" PACKAGE "INTERLISP")
(FILECREATED "22-Feb-2021 12:41:59" 
|{DSK}<Users>kaplan>Local>medley3.5>git-medley>library>FILEBROWSER.;25| 259384 
```

Version 8 was before EOL conversion, version 25 was after.
```
(RPAQ? FB.ICONSPEC '(#*(83 70)OOOOOOOO...
```
is what went missing

masinter:
I can do `LOADDEF(FB.ICONSPEC VARS FILEBROWSER;8)`
which loads it from the earlier version but I get
`(RPAQ? FB.ICONSPEC NIL)` when I `SHOWDEF(FB.ICONSPEC INITVARS)`

masinter:
PRETTYCOM has a big selectq selecting the built in COMS types. The path for (VARS ARRAYS) is slightly different between VARS and INITVARS  -- the VARS case calls PRETTYVAR1 which, if you say (VARS FB.ICONSPEC) it will get the top level value. Apparently, at the time the file was last made version 8, (INITVARS FB.ICONSPEC) would pick up the GETTOPVAL and print that.
But the path for INITVARS has no GETTOPVAL call. Maybe someone changed it. It seems suspicious to pick up the current value and use it as the initial value.

rmkaplan:
INITVARS should never pick up a current value, top level or otherwise.  It should only get the value specified literally in the (INITVARS command)

masinter:
It doesn't

PR #595, opened 2021-12-03
Shrink menu filebrowser icon restored

masinter:
Fix issue #404 -- restore filebrowser icon


library/FILEBROWSER
```diff
@@ -179,7 +185,8 @@ You specify how many versions to keep.")))
                                                              (* \; "Setup")
               (FNS FB.STARTUP FB.MAKERIGIDWINDOW)
               (FNS FB.PRINTFN FB.COPYFN))
-        (COMS                                                (* \; "commands and major subfunctions")
+        (COMS                                                (* \; 
+                                                           "commands and major subfunctions")
```
nbriggs, line 189: I'd still like to know why this got split onto two lines when, for example, at line 179 above it did not add a line-break.

masinter: i made a separate issue. Once we've fixed the various problems that make the files ugly when printed, we can re-prettyprint everything.

rmkaplan (approved):
I don't think this should be hung up on the pretty print question.  This does restore the ICONSPEC.  My medley-compare also showed a renaming of a variable (SUB to SI) in one of the subfunctions, but that's immaterial.

nbriggs (approved):
Yes, this does restore the FB.ICONSPEC.  Other questions can be handled later.

PR #596, opened 2021-12-03
reduce errors during GREET from out-of-order problems

masinter:
Need to MEDLEY-INIT-VARS() before loading other files.
In MEDLEYDIR, directoryname(<home>user>medley>sources) changed; make sure it ended in a ">" when it was intended as a directory.

With these changes (and a few tweaks) I'm able to GREET() in venuesysout


rmkaplan (approved):
LGTM

PR #597, opened 2021-12-03
Switch (back to) SEDIT mode when reacting to a MARKASCHANGED

masinter:
the logic around "markaschanged" is complex; this doesn't assume you haven't switched to EDITMODE(TELETYPE).
Replaces PR #541


masinter:
you can convert Issues to Discussions but not PRs.
Long discussion here, but I want to close this PR since I redid the SEDIT::MARKASCHANGEDFN edit as #597

nbriggs (approved):
Looks good to me.   I had to look carefully to see that the EDITMODE argument is in fact SEDIT:SEDIT, but fortunately the DEFINE-FILE-INFO showed up in the diffs.

Issue #359, opened 2021-07-29
save old (pre 2009) edit dates


masinter:
don't replace them

masinter:
@rmkaplan do  you have a patch to editdate? and fixeditdate which are on both FILEPKG and EDITINTERFACE

rmkaplan:
I have a patch, but I think the editdate stuff in FILEPKG should also be moved to EDITINTERFACE.  I was waiting, because I thought there were some other branches with draft FILEPKG changes.

masinter:
go for it. in general.  If we have any cases of branches where we've made different edits to the same file, in most cases we can do the merge very simply

rmkaplan:
The current set up is confusing.  It appears that the FILEPKG versions are overriding the EDITINTERFACE versions, for FIXEDITDATE and EDITDATE?.  I think the FIXEDITDATE on FILEPKG is better, I have movedit from FILEPKG to EDITINTERFACE (to be replaced by my better version).

EDITDATE? is harder to figure out.  Would be helpful to have a list of over-time formats to be looking out for, prefixed like "Edited by", dates (presumably satisfies a non-NIL (IDATE)), initials with and without semicolon.  Is there a list somewhere?

masinter:
Can't you just leave anything dated before 2021 or not in the current editdate format? or even by someone else?
We can manually edit down the list if the comments aren't informative.

rmkaplan:
So, recognize only dates of the current type and update them only for the current user’s initials.  Anything else is not a “date”, new dates always get packed on.

With the protocol that retains multiple dates, which I am going to install, I guess the only reason to recognize that a current comment is a date is to suppress gratuitous date accumulations for the same user in the same session.

Maybe that’s good enough, and alot of the complexity of both of these versions can be discarded.

Thanks

rmkaplan:
I have updated this on rmk5.  If *REPLACE-OLD-EDIT-DATES* is T, then any existing modern-form editdate comments will be replaced with a single new editdate comment (maybe cannibalizing the list structure).  Modern-form is "Edited <date> by <initials>".  Prehistoric formats are ignored, if you don't like them, delete them by hand.

If *REPLACE-OLD-EDIT-DATES* is NIL (initial value), then generally new a new editdate comment in front of existing editdate comments.  With the exception that if the most recent edit date is by this editor and it is less than a day old, the date in that existing comment will simply be updated.  This is to avoid a long sequence of meaningless editdate comments from a continuing edit session.

[...]

PR #599, opened 2021-12-04
Maintaining old edit dates #359

rmkaplan:


rmkaplan:
Also includes a trivial unrelated change to PFCOPYBYTES.

masinter (approved):
The code for sniffing an old edit-date might be useful if we ever want to go back and retrieve old dates and maintaners. On the other hand, we can just wait until then and sniff out the code.

LGTM

Issue #604, opened 2021-12-05
Add arbitrary user properties to SEDIT edit-contexts


rmkaplan:
There is curently no way for the user to hang arbitrary properties on an SEDIT edit-context.  There is a PROPS argument to the call to SEDIT, but that only allows properties that are already built in. Before I added an entry GET-WINDOW, there wasn't even a way to get the window that the SEDIT had created, to hang properties that might be useful after the sedit was constructed.

But it would be useful also to add properties to the edit-context itself, because that's actually the value that the call to SEDIT produces.  And it would be useful to specify properties in the call to SEDIT that might affect how the SEDIT is constructed, in particular for the user to be able to pre-specify the window or region to use.

So, add a PROPS field to the edit-context, initially populated from the already-present but extended PROPS argument to SEDIT, and provide functions GET-PROP and SET-PROP to manipulate user-specified properties (like windows, tedit streams, streams...).

Also, for better region management, extend SEDIT:GET-WINDOW-REGION to use the value of the property :REGION if present.

masinter:
I don't understand any circumstance where this would be necessary. The general way to add an arbitrary property to anything is to use an alist or hashtable or property list. Or a process property.
And I'm not sure we will ultimately want memorized window regions exactly aligned with sedit windows. Does this create circular structures that are not GC'd?

rmkaplan:
The main motivation, right now, is to provide information to the Sedit construction, like the desired region, BEFORE the process, window, and edit-context have been created.  This could be done by hash links and the like…but that’s not the way any of the other system components (Tedit, streams, windows) implement their property interfaces.

WIth respect to circularity, I think that there are so many back and forth pointers in Tedit and Sedit and their windows and processes that tghere are potential leaks all over the place., if objects are not shut down properly (and if the shut-down knows how to break the links). They just tend to happen at user-interface time-scales, so they never build up in a noticeable way.

The user could add a circular link in the sedit props field, but they could also do that in windowprops, streamprops, teditprops etc.  This doesn’t create any new danger.

masinter:
I don't see what this change has to do with SEDIT. You want a (LispUsers?) package to manage windows a particular way (some people say they want a Tiled window manager, for example; there are lots of styles depending on what you're used to).
Anyway, I would imagine some kind of generic SEdit-independent way of "asking for a region" with some supplied constraints
where there's a FLG that turns on the feature and you can try it out.

[...]

PR #613, opened 2021-12-07
SEDIT-TOPLEVEL:  Add functions for setting/getting properties #604

rmkaplan:
also a field in the edit-context to store them, and a change to the region subroutine to respect the :REGION property.

Note that SEDIT has a peculiarity in that the record declarations are on SEDIT-DECLS but the runtime INITRECORDS is separately on SEDIT-ACCESS.  So you have to make sure you dump SEDIT-ACCESS, not just SEDIT-DECLS if a record changes.  A bit of a trap


sources/SEDIT-TOPLEVEL
```diff
@@ -231,8 +265,7 @@
 
               (IL:* IL:|;;| "found a matching context elsewhere")
 
-              (IL:RESETFORM (IL:EDITMODE SEDIT)
-                     (IL:EDITDEF NAME TYPE NIL NIL '(:DONTWAIT))))))))
+              (IL:EDITDEF NAME TYPE NIL NIL '(:DONTWAIT)))))))
```
masinter, line 269: it looks like this PR loses the edit I made in a previous PR

masinter (changes requested):

rmkaplan:
Added your MARKASCHANGEDFN fix.  Might have missed it before because of a delayed merge

sources/SEDIT-TOPLEVEL
masinter: this is screwy -- I updated rmk6-redux to have the right EDITMODE but it should have been this one rmk7 instead.
once more? 

i got an error that I haven't tracked down -- too many arguments (2) to sedit::start-process (which only takes one).

rmkaplan:
I think the missing quote explains an error that I got, of a NOBIND when I was just trying to do a COPYDEF.  It got the error inside FNS.PUTDEF.

rmkaplan:
Should I put the quote in my version (which has the other stuff), and then commit it back to rmk7?

sources/SEDIT-TOPLEVEL
```diff
@@ -119,20 +121,38 @@
...
+
+    (IL:* IL:|;;| 
+    "Edited  4-Dec-2021 21:54 by rmk: The :REGION property gives the user directe control")
+
+    (IL:* IL:|;;| "RMK: The :REGION property gives the user directe control")
+                                                     (IL:* IL:\; "Edited  1-Dec-2021 22:51 by rmk:")
+                                                             (IL:* IL:\; 
+                                                             "Edited 19-Nov-87 10:18 by DCB")
+
+(IL:* IL:|;;;| "called to get a region for this sedit window.  should return the region for the sedit including the prompt window.  context is being built and needs a window.  the context will have at least the name (IconTitle) and type (EditType) of the object being edited, and can be used as desired to map between contexts and windows.  If reason is :CREATE, then this function must return a region.  If :EXPAND, then this algorithm returns a region from the stack only if SEDIT.KEEP.WINDOW.REGION is nil, otherwise it returns NIL, telling the window system not to reshape on expansion.")
```
masinter, line 135: comment about the REGION property appears twice

masinter (approved):

PR #614, opened 2021-12-07
Revert "MKPROGN from record was overridden by the better one on WTFIX; ancient bug tickled when compiling LIFE"

masinter:
Reverts Interlisp/medley#612


rmkaplan (approved):

masinter:
broke loadup for unknown reason

PR #615, opened 2021-12-08
rmk6 2nd change to EDITINTERFACE

masinter:


masinter:
The simplest way I understood for resolving this was to just do a new branch with the rmk6 files included

rmkaplan (approved):

PR #621, opened 2021-12-10
compile and try SINGLEFILEINDEX

masinter:


rmkaplan:
The .PS file seems to be a Lisp source file, not a PS file

PR #624, opened 2021-12-13
restore-versions now sets date of version to date of commit

masinter:


nbriggs:
Looks OK as far as setting the dates.  I wasn't sure what to expect it to do w.r.t. version symlinks for a file which hasn't had any updates committed to git -- it appears to do nothing.  For example lispusers/ACE/ACE-PRIM is a file which existed as version 2 (and only version 2) and it didn't make ACE-PRIM.~2~ as a link to ACE-PRIM.

rmkaplan (approved):

masinter:
The "restore-versions" doesn't restore any versions before a file was moved. I think `ACE.~2~` was at toplevel lispusers rather than the ACE subdir.

Issue #499, opened 2021-09-22
moving Medley files into standard places when fitting into Unix distros


masinter:
Converted back to an Issue for more focused discussion

### Discussed in https://github.com/Interlisp/medley/discussions/77

<div type='discussions-op-text'>

<sup>Originally posted by **blakemcbride** December  9, 2020</sup>

# Path Proposal for Medley

Revision History
----------------------
Blake McBride	2020-12-09	Original proposal

------------------------------

Problem
-----------

Special configuration options notwithstanding, Medley defaults to storing its lisp.virtualmem, SYSOUT files, and makefiles in the user's home directory.  This surely made a lot of sense in the days when machines were smaller and more single-purpose.  However, this is a problem with modern machines because:

1.  It clutters up a user's home directory.
2.  When working on multiple projects, the various lisp.virtualmem, SYSOUT, and makefiles get intermixed.
	
Proposal
-------------

1.  Medley's executable programs and scripts should be located in the /usr/local/bin directory.

2.  Store other common Medley files (such as standard images, system source code, documentation files, etc. in the /usr/local/share/medley directory.  /usr/local is called the Medley Home directory.
    
3.  The directory the user was in when Medley is started is called the "Project Directory".

4.  By default, Medley accesses the lisp.virtualmem, SYSOUT files, and makefiles in the Project Directory.  It may also have pointers to and read files in the Medley Home directory.
    
5.  If there is no lisp.virtualmem file in the Project Directory, Medley will load a standard SYSOUT located under the Medley Home.
    
6.  Under normal circumstances, Medley does not access any files outside of the Medley Home directory or the Project Directory.
    
7.  Unless explicitly changed, Medley will treat the Project Directory as its home or default path.
    
8.  Users are not allowed to write to the Medley Home. 
    
9.  The medley build process should have a make target "install" that sets up and installs the Medley Home.

Advantages:
--------------------------------------

1.  The system will be laid out in a form well understood by the community.

2.  The system will be laid out in a practical and clean way that is cooperative with other applications on the system.
    
3.  The system will best support multiple projects as well as multiple users on the same system.
    
Potential Issues
----------------------

1.  Users are not allowed to write to the Medley Home.  Therefore, some procedures will need to be used to support efforts to modify Medley system files when modifying Medley-specific files.

2.  This proposal makes sense for Linux, BSD, and other standard Unix-like distributions.  It may need adjustments for the Apple Mac environment.
</div>

masinter:
[...]

PR #627, opened 2021-12-15
Add environment variables move medley files to /home/medley.

stumbo:
Add Build_Date, Maiko_Release and Medley_Release environment variables.

Touches on issue #499


masinter (approved):
LGTM, thank you

nbriggs:
See the red circle at the end?  That says "missing newline at end-of-file", so it shouldn't look good.

Screen Shot 2021-12-14 at 11 18 05 PM

masinter:
? is that a problem? it was missing before this update

nbriggs:
It's definitely something to avoid.  It's absolutely something to fix if you're already editing the file.  Not *all* utilities that read files treat EOF the same as EOL with respect to processing the last line in the file.  That's why it's flagged in red, and shows up with `\ missing newline at end-of-file` any time you run a (git) diff.

PR #633, opened 2021-12-22
Rmk11: Mostly minor tweaks

rmkaplan:


masinter:
commits already merged with rmk12

PR #634, opened 2021-12-22
Rmk12: Lispusers packages of general utility, but especially for git compare

rmkaplan:


rmkaplan:
Note that rmk12 was created under rmk11 instead of master, so I presume it would inherit the previous rmk11 commits when they are merged.  I hope that's the right way of tracking sequential pull's.

masinter (approved):
LGTM (without testing alas)

PR #641, opened 2022-01-03
Rmk13: More infrastructure for flexible region management and support for comparisons

rmkaplan:


masinter (approved):
browsed changes. looks OK. That TEDIT(ATOM) works and TEDIT(PATHNAME) (using Common Lisp pathnames) is an ancient non-feature.

PR #642, opened 2022-01-04
Rmk14: Browsers for COMPAREDIRECTORIES, COMPARESOURCES, COMPARETEXT for TEDIT files

rmkaplan:


masinter (approved):
i'm not sure it works to merge the two branches into master when one is just the subset of the second; does this work with Git?

rmkaplan:
I was going to write a message about this.  I am making new branches sequentially in Medley, based on initials and bumping the number by one.  So, if 13 is the largest n such that rmkn is a branch, my next branch will be rmk14.

The question is, should rmk14 be made while rmk13 is the state of my current clone, or should I first switch back to origin/master and then make rmk14.

In the first case, rmk14 will inherit all of the commits in rmk13, plus its own commits.  If I  compare rmk14 to rmk13, I see just the additional commits.  But if I compare it to origin/master, I see the union of all the changed files.

So I presume that approving and merging the newest one would get them all in master.  But that’s a lot to approve in one go.  So approval/merge of the parent one first would allow incremental focus on a smaller number of files.  

I think it would be more dangerous if the later one were made directly on master, and it was merged in first.  If it had dependencies on rmk13 files that had not yet been merged, it might be messy to unwind.

But what do I know.

masinter:
in the case in hand, you have rmk13 and rmk14, where 14 includes all of the changes of 13. But either they are independent or they are not. If they are both approved, it would be better to merge 14 and delete 13.

masinter:
hmmm. I merged 14 and then merged 13 with no  problem. No conflicts so.... so far so good.

rmkaplan:
My theory is that if you merge 13 and then 14 you would get the same result as if you just merged 14.  The difference would be in how many files you have to look at to do a single approval, and whether it is better that they are set up in stages.

PR #651, opened 2022-01-25
Rmk17:  Minor tweaks to sources

rmkaplan:
Mostly extending/smoothing some rough edges


nbriggs (changes requested):
UNPACKFILENAME.STRING fails to produce correct results on
 `(UNPACKFILENAME.STRING "{host}dev:/dir1/dir2/foo.xx:yy")`
that is, any time there is a second colon after the first in a legal device place.   That example produces `(HOST "host" DEVICE "dev:/dir1/dir2/foo.xx:" NAME "yy")`

rmkaplan:
Fixed the double colon problem.

nbriggs:
The fix doesn't quite implement the grammar and quoting rules from the IRM --
```
The standard syntax for a file name that contains all of those fields is
   {HOST}DEVICE:<DIRECTORY>NAME.EXTENSION;VERSION
```
and
```
Different hosts have different requirements regarding which characters are valid in file names. 
From Interlisp-D's point of view, any characters are valid. However, in order to be able to parse
a file name into its component fields, it is necessary that those characters that are conventionally
used as file name delimiters be quoted when they appear inside of fields where there could be ambiguity.
The file name quoting character is "'" (single quote). Thus, the following characters must be quoted
when not used as delimiters: :, >, ;, /, and ' itself. The character . (period) need only be quoted
if it is to be considered a part of the EXTENSION field. The characters }, ], and)
need only be quoted in a file name when the host field of the name is introduced by {, [, and (, respectively.
The characters {, [, (, and < need only be quoted if they appear as the first character
of a file name fragment, where they would otherwise be assumed to introduce
the HOST or DIRECTORY fields.
```
As someone (whose name I don't remember, from the CSL quote board) once said: Parsing is the Vietnam of Computer Science.

nbriggs:
Maybe it's close enough, though.  I don't know whether PACKFILENAME.STRING should be introducing quote characters when creating a name out of parts that would create an ambiguity when they're packed together.

rmkaplan:
It does parse the canonical example properly, and it does treat the ‘ as an escape.  But all of this parsing code is a mess, so it wouldn’t surprise me if something is screwed up..  But it’s always been that way, this change was just to make sure that the device colon (unquoted) comes before the directory brackets or slashes.  There are 2 different (but I think exactly equivalent) subfunctions \UPF.NEXTPOS and LASTCHPOS that seem to interpret the ‘.

So I tried to fix the thing that was clearly broken (and that interfered with my use of time-stamps in directory names).  There may be other departures from the nominal specification, both in unpacking and packing, it’s hard to say.  (What we really need is a light-weight finite-state-transducer package for stuff like this, perhaps one that compiles its patterns out into code.)

nbriggs:
... and a grammar for what it's actually supposed to recognize, which isn't exactly well specified, with descriptions such as quoting being required "where there could be ambiguity"...

masinter:
It’s OK to say  that some perfectly valid Unix pathnames can’t be used by Medley {DSK}. That includes any unicode filenames.

I'd suggest adding *%  to .gitignore.
Are there popular conventions for modern commonlisp implementations? 

--

masinter:
But this discussion is very confusing... which {HOST} do we have that has any "device:"?
Seems like a lot of discussion of a non-existent problem, in ancient code with other, serious problems.

rmkaplan:
Yes, the device issue is an anachronism.  But it was there, and I think not correctly/usefully implemented—the first colon should end the device field, not the last one.

nbriggs:
... and when you had a colon in the name NOT as part of a device it confused the parser.

When you're running on a Windows system, host {dsk} should have a device -- "{dsk}D:/foo/bar.txt" would be the way I would expect to refer to a directory on the D: drive.

sources/UFS
```diff
@@ -1,10 +1,11 @@
 (DEFINE-FILE-INFO PACKAGE "INTERLISP" READTABLE "INTERLISP" BASE 10)
-(FILECREATED "21-Apr-2021 11:36:54" {DSK}<Users>kaplan>Local>medley3.5>git-medley>sources>UFS.;5 69271
```
nbriggs, line 2: I'm not sure how this code interacts with the Maiko ufs.c code and dir.c code...

rmkaplan:
I think that the directory depth enumeration is done by faking recursive calls to the top-level Maiko enumerator.  All this was intended to do was to make the top-level parameter available in the recursive calls, as opposed to it just always going to * for the version and extension, and enumerating more than subdirectories were asked for.

So, my hypothesis is that doesn’t interact with Maiko in any new way, just passes better arguments to the recursive calls.

nbriggs:
I *think* the Maiko code only gets a string, so passing the DEFAULTEXT and DEFAULTVER will probably turn into a "better" string being sent on down.  But, there are comments in the "dir.c" code like:
```
/*                                                                                                                                                                                                  
 * The way to deal with the version field in file enumeration is a little                                                                                                                           
 * bit tricky because of the bad specification of original {UNIX} device.                                                                                                                           
 *                                                                                                                                                                                                  
 * According to the Medley 1.1 release note, in the representation                                                                                                                                  
 * "{UNIX}<dir>name.ext;3", ';' and '3' are regarded as a part of the                                                                                                                               
 * file name, not its version.  On the other hand, in 1.1 implementation,                                                                                                                           
 * in the pattern "{UNIX}<dir>*.*;*", the ';' and the last '*' are regarded                                                                                                                         
 * as a version field, not part of the file name.  Actually the pattern                                                                                                                             
 * "{UNIX}<tmp>*.*;*" enumerates all of the files on /tmp directory                                                                                                                                 
 * even if they never include ';' character in its name, as well as '.'.                                                                                                                            
 *                                                                                                                                                                                                  
 * Thus I believe, the specification should be clean upped as like,                                                                                                                                 
 * "UNIX device always ignores the version field in it file name representation                                                                                                                     
 * even if a user specifies it explicitly".                                                                                                                                                         
 * But to keep a compatibility to an already released version, we have                                                                                                                              
 * to do some trick here.                                                                                                                                                                           
 */
```

nbriggs (commented):
I don't see STRING.EQUAL in the IRM, though that surprises me a little, but I guess that means there's no corresponding documentation change necessary.

nbriggs (approved):
I think that this will work.

rmkaplan:
Yes, it isn’t in the IRM, for whatever reason.  I was surprised.  But I figured that a simple extension wouldn’t screw anything.

rmkaplan:
For {UNIX} it must be a little different than for {DSK}.  Where does it do the ~n~ transformation?

What this should do, and I think it does, is honor the case of “”/NIL (highest version), * (all versions) and a paricular n (just that version), in a subdirectories as well as the top directory.

masinter:
> I don't see STRING.EQUAL in the IRM, though that surprises me a little, but I guess that means there's no corresponding documentation change necessary.
?
image

masinter:
I'm not sure what I'm worried about -- loading things into old sysouts?

rmkaplan:
> I'm not sure what I'm worried about -- loading things into old sysouts?

I don't think that's an issue for this, the function existed and its old behavior is preserved, it now has a new optional argument

masinter:
when you add a feature in sources (core ) and lispusers that aren't in the loadup depends on that feature?


I'm also worried about introducing bugs that only happen with case sensitive file systems. Mac and windows are case insensitive but linux and WSL are case sensitive.

I've noticed that IL:DIRECTORY returns all upcase filenames but INFILEP returns case preserved names.
There were problems with matching lower case font filenames -- I suspect FONTSAVAILABLE had the problem.

Common Lisp uses another datatype (pathname) instead of Interlisp's strings, although there seems to be some coersion.

TEDIT won't take a string as a file name -- it treats it as the string to edit.

probably 3 separate issues

PR #655, opened 2022-01-27
Rmk18

rmkaplan:
More simple cleanups.  Some were in the now abandoned rmk15


nbriggs (approved):
Looks good.  It would probably be interesting to do a little bit more to investigate the RESETLST/RESETSAVE issue though in case it's a symptom of something more insidious.

PR #658, opened 2022-01-28
Rmk19: Updates and remaining components for managing comparisons and interactions between git and Medley

rmkaplan:


masinter (approved):
All lispusers packages not in loadup so I didn't look hard

rmkaplan:
It would be good for others to do a sanity check now.  Loading GITFNS.LCOM should load all of the component pieces. Make GITMEDLEYDIR point to your clone, and then try  "prc lmm10" to compare branch lmm10 to origin/master. (See GETFNS.TEDIT).

As a starting point for the "my medley" mode of separating working Medley from the clone, make a copy of the origin/master clone (without the . files) in a new directory, and let MEDLEYDIR point to that.  If you now check out lmm10 in the clone, the command gmc will compare library, lispusers, sources of lmm10 against your working Medley files (which at the beginning are the same as origin/master, so the result should be equivalent to the prc test).

PR #664, opened 2022-01-30
Rmk19

rmkaplan:
A few gti bug fixes, mostly clean up around window/region management.


masinter (approved):
lispusers under development will try soon

PR #665, opened 2022-02-01
New version of IRM

Anzus:
New version of the IRM, updated to Medley.


rmkaplan:
When I try to open a TEDIT file, I get a font not found error for the display font Palatino 9 MRR.  Should the site greeting file include fonts>adobe> on DISPLAYFONTDIRECTORIES?

Having pushed that on, the files open but with a lot of unknown image object boxes for IM-INDEX.   Maybe a readme file in the directory or a separate loader that gets all the right stuff?

Should the pdf's be in a separate subdirectory?

And should this directory be renamed?  "new" is a relative notion.  What about "medley-irm"?

nbriggs:
I agree there should be a package (internal/library ?) that makes sure all the image objects are defined and the fonts are available.

masinter:
> When I try to open a TEDIT file, I get a font not found error for the display font Palatino 9 MRR. Should the site greeting file include fonts>adobe> on DISPLAYFONTDIRECTORIES?

does that help? I thought there might be problems with font file names.

> 
> Having pushed that on, the files open but with a lot of unknown image object boxes for IM-INDEX. Maybe a readme file in the directory or a separate loader that gets all the right stuff?

Image objects code needs to be loaded. If you load MAKE-PS it has a lot of them, except Notecards has its own not in the medley repo. 
 
> Should the pdf's be in a separate subdirectory?

I've been putting pdfs in Interlisp.github.io (interlisp.org) or in Google docs because they're useful without running Medley.
I've been putting "original source" files in the "history" repo and the "latest version" (for updating) in the "medley" repo.

> And should this directory be renamed? "new" is a relative notion. What about "medley-irm"?

in the medley repo i'd call it docs/irm.  When you render a filename.tedit to a printable format the image object creates a filename.imptr which I imagine is used by helpsys. You'll see that '*.IMPTR' is in .gitignore.

nbriggs:
Ah, I see in `medley/docs/Documentation Tools` there are all the tools and TEdit documentation for the IRM process (which might shed more light on the DINFO situation too... I haven't read through them all at this point).

Anzus:
Yes, the IM-INDEX image objects are used by IM-TOOLS to create the index (if I can reconcile John’s last edits to the code with the original). The .IMPTR files are indeed used for the index files. This repository does not contain that index, because I haven’t been able to get it to work yet, but I know there was interest in the IRM in TEDIT form. There’s a fairly arcane process that is only partially documented.

I’m happy to rename the repo to medley-irm.

masinter:
It looks like this was mainly a substitution of Medley for Interlisp-D in the text.
image
21-PERFORMANCE (Medley) and 22-PERFORMANCE (Interlisp-D)
image
22-PROCESSES (Medley) and 23-PROCESSES (Interlisp-D)
23-SSTREAMS (?) and 23-STREAMS (?)

PR #672, opened 2022-02-10
cleanup lispusers and library H* I* K* L* and started on M*

masinter:
making a little progress on #7


masinter:
I  put a README.md in lispusers that points to t he  (inaccurate) LispUsers files until I've gone through the alphabet

PR #674, opened 2022-02-12
Complete revamp of the buildRelease and buildDocker workflows for Medley

fghalasz:
Complete revamp of the buildRelease and buildDocker workflows for Medley to use new release tags and the sentry model introduced by jstumbo.  Also adding the buildReleaseInclDocker composite workflow.


masinter (approved):
wow

PR #675, opened 2022-02-12
Fgh002.1: Workflow to test Docker Hub Login in this repo

fghalasz:
The buildDocker workflow is failing because it fails to login to Docker Hub.  Probably an issue with the DOCKERHUB secrets for this repo.  This workflow allows us to test/debug this.  Needs to be on master branch for it to work.


masinter (approved):
getting to a stable base

PR #679, opened 2022-02-15
get rid of unused font files from medley release

masinter:


rmkaplan (approved):
Seems reasonable.  If we need to retrieve any (like some sizes of Optima?), we can find them under obsolete.

masinter:
i had forgotten this PR when I did them again in the later promote-internal branch

masinter:
oops I messed up; merging this was wrong

rmkaplan:
BTW, there is still a directory fontsold/, is that still relevant?

masinter:
Fonts are still a mess. We need some focus on the "get modern fonts" vs. patching around.

PR #684, opened 2022-02-18
lispusers cleanup, move directories

masinter:
- restore-versions now sets date of version to date of commit
- cleanup lispusers and library H* I* K* L* and started on M*
- Add lispusers/README.md and move more overview files to docs
- More cleanup of lispusers M* (still not done)
- cleanup of cleanup
- More cleanup for I* thru P*
- move directories around
- fix loadups to match moved directories
- FINDFILE now also looks for foo>FOO when looking for FOO
- Check in medley-irm
- Add more font directories to release
- emacs inserted tabs instead of spaces, now fixed


masinter:
subsumed by other (merged) PRs now

PR #691, opened 2022-02-24
Move test files to its own repo

masinter:


nbriggs:
Not mentioned:  rename of `dot.read-me-first` to `readme-first.tedit`, and the rename of some directories from uppercase to lowercase, for example (new) test/io/Hardcopy/Hand/STRESSTEST which used to be medley/internal/test/IO/Hardcopy/Hand/STRESSTEST but otherwise I checked that everything that was in test is in the new test repo.

nbriggs (approved):
Looks OK.  Everything checks out with the new Interlisp/test repository.

PR #692, opened 2022-02-24
remove unused Xerox files.Z and shorten dir name

masinter:


rmkaplan (approved):
Looks like xeroxprivate got renamed to xerox, and that all the .Z files got deleted.  That's also  what I see in my upgraded GITFNS, so LGTM

PR #693, opened 2022-02-24
Revert "get rid of unused font files from medley release"

masinter:
Reverts Interlisp/medley#679

wrong move, shouldn't have merged


rmkaplan:
What was wrong?  I just updated my Medley working directory to be compatible with something.  It renamed xeroxprivate to xerox, and deleted a bunch of .Z files.

Should I just wait until this settles down, and then just do another synchronization?

masinter:
sorry I see 679 was ok

PR #694, opened 2022-02-24
change MEDLEYDIR-INIT to be more suitable for newcomers

masinter:
these are a matter of taste to some degree
My intent is to set things up that new users will be less confused, based on limited bug reports and questions asked we've seen

todo still: open Interlisp Exec#2
 adjust layout depending on screen size at any restart


masinter:
handled by other PRs

PR #695, opened 2022-02-25
Change OPENTEXTSTREAM to not error when finding display font (for hardcopy)

masinter:
Don't need display font to hardcopy


rmkaplan (approved):
LGTM

PR #698, opened 2022-02-25
Move LispUsers documents to subdir for later sorting

masinter:


masinter:
these files need to be sorted out based on issue #7 resolution. They were sprinkled through the directory and difficult to distinguish.

rmkaplan (approved):
It looks like this is just a rename from lispusers/x lispusers/000-docs/x

At least it collects these together

PR #700, opened 2022-02-25
Move logic files to subdir

masinter:


rmkaplan (commented):
I don't think that "h" as a subdirectory is very illuminating.  Good to group these together, but maybe make the subdirectory be horn/ ?  Even though that doesn't fit into your scheme find-file scheme.

logic/ works better

masinter:
part of my constraint was not wanting to change the file names (which would involve loading and changing the filecoms)..
It might be OK to just rename the files? Or alternatively, find some other way to organize the files using one of the 000-docs file index documents. I hadn't understood why they were there (the files in 000-docs).
I wanted to make progress on Issue #7 -- clean up "LispUsers".  We set off some activity around filling out the LispUsers spreadsheet, but we didn't have a process for making use of that information.
I started in the middle of the alphabet because that was a stretch without many reviews.
(Find the spreadsheet by look;ing in comments for issue #7 )

nbriggs:
I think the "clean up lispusers" could (should?) happen without moving files to sub-directories.  Then (or in parallel) we could do a design (and implementation) for changing FINDFILE and related functions so that they all work together, "properly", in the presence of sub-directories.  Or, we could think about what problem(s) with the lispusers packages we were trying to solve in the first place... not knowing what files were part of a package, not knowing the dependency relationships between files and not knowing whether they are utilities shared between packages, etc.   Some of that might just be updating the FILESLOAD coms, some might be updating the documentation.

masinter:
hold off on this PR until strategy is clear.

PR #701, opened 2022-02-25
Rename migration directory

masinter:


rmkaplan:
I think it is a very bad idea to have files with copies of FILEKGRECORDS, SEDIT-DECLS, and TABLEBROWSERDELCS in this directory.  If you load it in, you may be screwed.

If this needs those records, it should load those files from library or sources.

masinter:
@Anzus ?

rmkaplan:
Whatever it is doing, it shouldn’t have its own version of system declarations.  It should recompile with the standard ones.

masinter:
do this later

PR #702, opened 2022-02-25
obsolete many lispusers

masinter:


rmkaplan (changes requested):
NSDISPLAYSIZES is not obsolete.  You need it if you want to synchronize display and hardcopy with fonts like Classic that have a modicum of non-ascii characters

masinter:
Put NSDISPLAYSIZES back

rmkaplan (approved):
LGTM

PR #704, opened 2022-02-25
*** Update scripts and utilities to use renamed directories

masinter:
updates FINDFILE-WITH-EXTENSIONS (when called on a root name FOO or FOO-something) to look at FOO>FOO-something.extension) to handle subdirs,


rmkaplan:
Perhaps a separate issue, but if we organize multi-file library or lispusers packages with their own subdirectory (a good idea), what is the FILES argument to EDITCALLERS or FINDCALLERS that will sweep through and find everything?  (FILDIR '*.;) sweeps just the top level.

masinter:
I split this into separate PRs to facilitate review

masinter:
this was spllit into separate PRs but has a correction to the release-medley script ("medley/docs/doctools" => "medley/doctools"

masinter:
will include in a separate PR

PR #708, opened 2022-02-28
FINDFILE-WITH-EXTENSIONS update only

masinter:


masinter:
This extends FINDFILE-WITH-EXTENSIONS to look for files in a subdirectory with the same root name as the file name being loaded -- if you are looking for FOO-BAZ.COM in DIRECTORIES it will also look for FOO>FOO-BAZ.COM.
This may not be extensive enough but it's enough for FILESLOAD to work and doesn't require lots of renaming of files.

rmkaplan (approved):
LGTM

PR #711, opened 2022-03-01
Medley logow

masinter:
New Medley Interlisp logow (using old KOTOLOGOW); reenable LOGOW in Loadup


masinter:
this isn't working, but I think it isn't needed for release

masinter:
The problem was the FILEPKG bug... now ready for review again.
Changes LOGOW to use Medley as the name in a KOTOLOGW look.
Further refinements of LOGOW still pending (include release tag?)

rmkaplan (commented):
The new LOGOW seems to have a lot of stuff under (if NIL...)
Is that intended?

masinter:
I wanted to comment it out but just meta-; in sedit leaves it illegible. I suppose I should have left a comment.

rmkaplan:
In that situation, I revert to (* …)

rmkaplan (commented):
LOGOW has both a DFASL and an LCOM.  Doesn't have FUNCTIONS or DEFMACROS, LCOM should be OK

masinter:
got rid of DFASL and venue logo code

rmkaplan (approved):
Looks good.  I take it that the way to programmatically close the window is (AND LOGOW (CLOSEW LOGOW))

I think FILEPKG is the same file as in a separate PR, but that's OK

PR #712, opened 2022-03-02
Odds n ends

masinter:
in the course of cleaning up lispusers, some odds and ends


rmkaplan (approved):
2 of the files have the same content, except for the copyright date.  Change seems to be prettyprinting.

On the histmenu.tedit, you switched from Xerox to Interlisp.org.  On the Tedit files I have created (e.g. GITFNS, PSEUDOHOSTS...), I used a bigfont "Medley" as the header icon. We can come synchronize later.

PR #717, opened 2022-03-04
Fix bug introduced by 'changes to%' => ':CHANGES-TO'

masinter:


masinter:
I'm surprised this didn't hit more broadly, but most cleanups don't use FILECHANGES apparently; haven't found any more files (other than LOADUP-LISP) this had affected, but I'm still looking.

PR #720, opened 2022-03-05
DIRECTORY:  Added DEPTH parameter for FILEING.ENUMERATION.DEPTH

rmkaplan:
DIR * DEPTH 2


masinter:
I had thought we might get better compatibility between CL:DIRECTORY and IL:FILDIR (and IL:DIRECTORY?) if we extended the wildcard syntax rather than adding another parameter. I was thinking about syntax patterns like '**' or '*/' or '**/' to indicate depth.

rmkaplan:
Could do that.  I was thinking at one point of something like  2*/ or 3*/, and **/  for all the way (here I’m thinking of Kleene star, but maybe just */).  I wasn’t sure if that would have to be interpreted in lots of different places, but maybe it is only DIRECTORY that would have to unpack it into a value for FILING.ENUMERATION.DEPTH.

That could be an alternative way of specifying this.  The DEPTH parameter would still be useful, since you could specify it without a lot of string unpacking and packing.

sources/DIRECTORY
```diff
@@ -95,17 +101,26 @@ Copyright (c) 1986, 1987, 1988, 1990, 1992 by Venue & Xerox Corporation.  All ri
...
+                                    then                     (* ; "A number of days")
+                                         [IDIFFERENCE (IDATE)
+                                                (TIMES COM (DEFERREDCONSTANT (IDIFFERENCE
+                                                                              (IDATE "2-JAN-77 00:00"
+                                                                                     )
+                                                                              (IDATE "1-JAN-77 00:00"
+                                                                                     ]
                                   elseif (IDATE COM)
                                   else (\ILLEGAL.ARG COM))))
+                    (DEPTH [SETQ FILING.ENUMERATION.DEPTH (IF (AND (SMALLP (CADR COMTAIL))
```
masinter, line 113: I think this isn't enough to work well -- probably need to make FILING.ENUMERATION.DEPTH special and rebind it rather than resetting it globally.

masinter (changes requested):
provide context for depth (see note) and/or get the same effect by indicating depth by different patterns of * and / delimeters

rmkaplan:
I did rebind it, and declared it special in this function.   As far as I could tell, it is not declared global anywhere.

You are asking for a toplevel SPECVARS declaration?

masinter (approved):
sorry misunderstood. I think  check filebrowser and various INIT files

Issue #715, opened 2022-03-04
MEDLEY causes an error if the file isn't found, even if NOERROR=T


rmkaplan:
Suppose foo is not a directory

(MEDLEYDIR "foo" NIL NIL T) returns NIL. Good

But
(MEDLEYDIR "foo" "fie" NIL T) still cause an error.  Bad

rmkaplan:
Actually, the behavior when NOERROR is T and the file/directory exists is also not good.  It returns T instead of the path to the file.

masinter:
should be fixed by PR #721

PR #722, opened 2022-03-05
MEDLEYDIR:  eliminate recursions, fix noerror #715

rmkaplan:
I also took a stab at this


masinter:
Mine's shorter and doesn't rely on FINDFILE (which is found on the file SPELLFILE which comes later in the loadup).

rmkaplan:
OK

PR #730, opened 2022-03-07
moving obsolete lispusers; delete some junk files

masinter:
- lispusers [HIJK]* sort out
- lispusers [LM]* sort out
- more cleanup


rmkaplan (approved):
Looks reasonable

PR #734, opened 2022-03-08
remove fonts/xerox already in displayfonts

masinter:


rmkaplan (approved):
I assume that you checked for byte-equivalence, so this looks good.
Separately, what about all of the .Z files in obsolete/xerox/.  Are those also redundant?

masinter:
the .Z files aren't redundant. Likely useless, but there aren't copies elsewhere.

PR #735, opened 2022-03-08
simplify DEFAULTFILETYPELIST initialization

masinter:
leftover nonsense from old makesys loadup full now gone.
also added WHEELSCROLL to full


rmkaplan (approved):
FWIW, "set" should be changed to "setting" in the message at the end of \UFStoOtherCopyMess.  But I'll approve

PR #751, opened 2022-03-22
Fixed some bad doc's

FultonBrowne:
Some of the tar command sequences where out of order, this patch fixes that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,12 +37,12 @@ From a shell/terminal window:
 Unpack the medley loadups file
 
 *  `cd ` ~parent~
-*  `tar -xvzf medley-`YYMMDD`-loadups.tgz`
+*  `tar xvzf medley-`YYMMDD`-loadups.tgz`
 
 2. Unpack the medley runtime OR clone the Medley repo
    (the "medley runtime" is just a subset of the whole repo)
    
-* `tar -xvzf medley-`YYMMDD`-runtime.tgz` 
+* `tar xvzf medley-`YYMMDD`-runtime.tgz` 
 
    OR
    ```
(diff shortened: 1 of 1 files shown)


nbriggs (commented):
That's a working fix.  I assume we're not going for "maximum portability" which would suggest the bundled flags form, "tar xvfz ..." (with no leading hyphen, and f does not need to be the last  flag)

FultonBrowne:
On Debian f needs to be the last flag, I've never used a build of tar that requires the - for flags

nbriggs:
Which debian?  I haven't run into any build of tar that *requires* the hyphen, but there used to be a number of old systems that didn't permit the hyphen.  The FreeBSD/MacOS behavior is that if you have the hyphen then "f" must be last, but without it any flags in the bundled flags that require arguments get them processed in order from *subsequent* command line arguments. 
image

FultonBrowne:
```fulton@debian:~/code/meledy$ uname -a
Linux debian 5.10.0-12-amd64 #1 SMP Debian 5.10.103-1 (2022-03-07) x86_64 GNU/Linux
fulton@debian:~/code/meledy$ tar -xvfz medley-220309-8400f7be-loadups.tgz 
tar: z: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
fulton@debian:~/code/meledy$ 
`
When I run tar using the instructions in the README

nbriggs:
@FultonBrowne -- I didn't say it worked as written.  I said if you leave off the hyphen neither debian nor any other system I've encountered cares about the order of the flags.

FultonBrowne:
My bad, I'm a bit confused with why you find an issue with my PR?

nbriggs:
I'm suggesting that we go for maximum portability, which is to leave off the hyphen (and then the order doesn't matter), but I certainly don't object to you changing it to "tar xvzf ..." rather than "tar xvfz ..." which is historically what most people would write.

nbriggs (approved):
LGTM.

PR #764, opened 2022-04-29
Rmk37 prc menu shows superset relations

rmkaplan:


masinter (approved):
Some nits for later:
* it would be good to centralize the logic for which files are text vs. binary. vs unknown
* there are some odd formatting issues that might be an artifact of the comparison tool rather than the code? (multiple edit dates on a line)
* i am not sure why LLCHAR had to change... It's more important to example changes to low-level files.
This review should over all rmk35..rmk37

rmkaplan:
The change to LLCHAR was to expose the iteration variables for the inpname inatom instring… i.s.oprs, so that the new UNPACKFILENAME.STRING could operate easily at that level.

rmkaplan:
On the odd printing of dates:  I think there are 2 issues.  One is that all of the dates should be at the same coment level (; ;; or ;;).  They used to all be ;  but dates with comments are typically at a higher level.  The date code should upgrade earlier ones to make them all consistent and give them a shot at all being aligned (It does a better fix-up job in a version that I haven’t yet pushed).

But the other problem is in the way that printdef deals with dates or comments in that first position. The source-compare browser pretty-prints individual definitions to a Tedit stream, then does the comparetext.  I think that printdef doesn’t do the proper alignment on at least the first comment.

It also differs from what you see in an SEDIT display, in that it sticks in extra line-feeds around all comments, so sequences of multiple comments are too spread out.  I’m hoping that PRETTYFILEINDEX, if its loaded, will do a better job (or is easier to fix).

PR #766, opened 2022-05-02
fix #411 scanfilehelp suggested 1100support

### LINKED ISSUE #411: error message 'call 1100support' doesn't give phone number
image

masinter:


rmkaplan:
Looking at this again, why does it call HELP instead of causing an error?

What could the user do in this situation?

masinter:
it's kind of crazy -- but not really worth fixing.
On modern machines just ignore the filemap or fix it.
I wouldn't mind marking #411 as "won't fix".

rmkaplan (approved):

rmkaplan:
This is still  a draft.  Can we incorporate this one way or another, and get rid of this?

masinter:
I think this was  harmless and probably shouldn't have been an issue, but ...

PR #775, opened 2022-05-23
Rmk41  tedit interprets strings as filename

rmkaplan:
This changes TEDIT and OPENTEXTSTREAM so that a string TEXT argument is interpreted as a filename, not a string to be put up in an edit window.

TEDITSTRING is provided as a new interface function that opens a stream on the string.  This should be used for the old behavior when the argument is a variable and its provenance can't be determined.

String calls from system code have been upgraded in this way.

This includes a few other minor updates.


masinter (approved):
probably should extend  to accept (common lisp) pathnames. What I remember but I haven't validated was that Medley pathnames were basically strings with a different datatype number, though, unless that got fixed.

Otherwise LGTM

rmkaplan:
I agree, pathnames should work.  On next round

PR #777, opened 2022-05-23
ADIR, TESTUPF:  New version of UNPACKFILENAME.STRING with test tool

rmkaplan:
See TESTUPF.TXT for testing information


masinter:
Not sure if this is a problem with PACKFILENAME.STRING but it's odd that when I (BREAK PACKFILENAME.STRING) and call (DIRECTORYNAMEP MEDLEYDIR) it has odd behavior
image

masinter (approved):
I don't see any obvious problems and I've been running with this, so I will Approve.
I accidentally pushed a change to MEDLEY-UTILS here, but the change is inconsequential.
I wonder whether the utility to check might belong in "internal" rather than "sources" but it's ok.

rmkaplan:
Actually, I intended the testing utility to be temporary, just in case there was what looked like odd behavior that had to be compared to the previous implementation. I wasn’t sure how to pass along a transitional testing jig.

I could delete it in another commit, or rename it into internal if we want to keep it around.  No need to keep it in sources/.

Sorry for the confusion.

rmkaplan:
I’m not sure that this is a problem.  DIRECTORYNAMEP calls PACKFILENAME.STRING with 2 directories specified, the argument to DIRECTORYNAMEP and the currently connected directory.  I think it is using the \CONNECTED.DIRECTORY to fill in defaults (host ?) that may not be included in the argument.

PR #778, opened 2022-05-24
Move lispusers> strike and c150 fonts to subdirectories

rmkaplan:
Next step will be to move the subdirectories to obsolete/


masinter (approved):
i suppose i should review / merge these two PRs in order?

rmkaplan:
I think there is a superset relation, if you do the second one first, you probably don’t need to do the first one (?)

PR #787, opened 2022-06-04
Rmk45  testupf to internal, tedit pathnames, minor doc changes

rmkaplan:
TESTUPF and TEDIT cleanups as suggested from previous PR comments


masinter (approved):
looks reasonable

PR #789, opened 2022-06-05
Rmk46: Minor changes to comparison functions

rmkaplan:


masinter (approved):
I thought this was covered by rmk27 but it doesn't look that way

Issue #790, opened 2022-06-05
prompt for TEdit window now shows #<Input stream on  .. in prompt window


masinter:
![image](https://user-images.githubusercontent.com/1116587/172072364-4a43eb50-0fc2-41c1-94c0-17c8aabb0c20.png)
an artifact of calls to Teditt with streams?

nbriggs:
Odd... we must have different things loaded: for me it puts up "Please specify a region for the editing window" with no name.

nbriggs:
I wasn't on the latest code in my loadup -- I see the same thing after rebuilding.

rmkaplan:
But is this a bug or a feature?  If you gave it stream on a parrticular file, wouldn’t you want to know which stream it wants to get a window for?  You could have editing sessions open on different streams for the same file, and somehow want to keep track of which is which. But then the bug would be that it doesn’t show the stream address in the title of the window after it has been created.

masinter:
this (bug) happens when I click "Edit" from the filebrowser.
fb medley/rooms/
select OFFICE.TEDIT and cllick "Edit". 

also

select ROOMS-INTRO click on See and you get Failed because NIL is not a NUMBER

![image](https://user-images.githubusercontent.com/1116587/172072692-3bb63ebd-0797-4b10-b631-09b8509f833d.png)
also tracing UNPACKFILENAME.STRING to track down

rmkaplan:
OK, I see that.  But I don’t think it has to do with the internals of TEDIT or UNPACKFILENAME.  And I think I noticed it before and put it on a list.

I made a change to call PFI.MAYBE.SEE.PRETTY for the TEDIT-SEE command, which is what the file browser calls for SEE.  (Also the macro ts in lispusers/TEDIT-PF-SEE.  Before that I was just doing COPY.TEXT.TO.IMAGE, which just converts the byte sequence (including doing font changes) but doesn’t reformat in any other way.

The PFI does its own reformatting, and I thought that had a better look (maybe more sensitive to the window width, better at comments…).

The problem here is that this file has a bitmap, and PFI.PRINT.BITMAP is screwed up in some way.  It seems to avoid this bug when printing to T but not when it is printing  to a text stream.  E.g
     (PRETTYFILEINDEX ‘rooms-intro NIL (OPENTEXTSTREAM) T)
fails

rmkaplan:
On the stream address showing up in the Window-region request string, I see that that happens even if you give a file name and not a stream.  I'll fix that.

[...]

PR #791, opened 2022-06-06
Rmk47: TEDIT, GITFNS, COREIO

rmkaplan:


rmkaplan:
This is when you are clicking on an item in the comparesources window?  (But you aren’t getting the git-command error that you saw before?)

masinter (commented):
is this a fix to #790? I still get NIL is not a number when I see ROOMD-INTRO but not when I See ROOMS-NOTES.

rmkaplan:
No, that’s a completely independent problem (see my previous note):  PRETTYFILEINDEX doesn’t seem to print bitmaps correctly on textstreams, although it doesn’t fail on displaystreams (or maybe it skips printing altogether on displaystreams, I haven’t checked)

masinter (approved):
no more git / gh command error so take these one at a time (approving)

masinter:
@rmkaplan I approved this a week ago, but didn't do the merge. PR#789 (rmk46) included.

Issue #795, opened 2022-06-16
"Build/Push Release & Docker" action picks up incorrect release of Medley when creating Docker image


fghalasz:
The `Build/Push Release & Docker` action just calls the `Build/Push Medley Release` action followed by the `Build/Push Docker Image` action (providing the `Medley Release` action succeeds).

The problem is that the release pushed by `Medley Release` action is marked as "Draft" and has to be manually moved from "Draft" to fully released status on Github before it can be seen by the `Docker Image` action.

When these two actions are called within the `Release & Docker` action, the release created by the `Medley Release` action is not moved out of draft status and hence is not available to the `Docker Image` action.  Instead the `Docker Image` action picks up the previously Medley release.

Obvious fix is to change `Medley Release` action to not mark the release as "Draft".

Is there any reason not to do this?   The "Draft" marking was included in the original buildLoadup.yml workflow, so I just left it in there.  But I am not sure why is was in there in the first place.  Any opinions?

masinter:
Is it possible to automate the addition of "changes since last release" in the release notes.
When you publish a release it will send out a release announcement. 

This was a manual step when there was a bug in the GitHub changelog generation but it seems they fixed that.

masinter:
see https://github.com/Interlisp/medley/issues/516#issuecomment-950519952

fghalasz:
PR#841  Updates Build/Push Medley Release so that it auto-generates release changelog.  With this PR, fully automated Build Release of Medley plus Docker image is working.

PR #796, opened 2022-06-16
Proposed fix for Issue #795:  Remove Draft status from Medley release in buildoadup.yml 

fghalasz:
See discussion in Issue #795.


fghalasz:
Once release is done, can still manually edit the release to generate changelog and add as release message.   Just as did before with Draft status release.

Only issue is that these release notes will not appear in the Medley Docker Image.

Meanwhile, I am working on getting the release message to be automatically generated as part of the buildLoadup.yml workflow (i.e., the workflow that builds the release)

PR #798, opened 2022-06-20
Add loadup-db.sh for loading all, mastersope index

masinter:
loadup-db.sh makes masterscope of everything it can (see OKSOURCES, OKLISPUSERS, OKLIBRARY, OKINTERNAL)
calls MAKE-FULLER-DB on file MEDLEY-UTILS

also fixes some problems with HCFILES on MEDLEY-UTILS which makes PS from TEDIT .

files are made in "tmp" subfolder not for release.


masinter:
restore FILESETS to state before this so that the changes are restricted to loadup-db, HCFILES, and pick

rmkaplan:
I tried scripts/pick, I got a whole bunch of "shuf: command not found"

I tried scripts/loadup-db.sh, I tot TMAX not found.  I think this needs the down-a-level convention.

Why is EQUATIONS and DATEFORMAT-EDITOR in the set of files?  What is the principle for inclusion?

masinter:
I suppose shuf is linux and not MacOS. The "pick a random number" utility in MacOS is 'jot' but not as handy.  I can move "pick" out of the repo and just keep it in my private release, or rewrite it in Lisp (using ShellCommand to get the github issue list, or better still, figure out how to call the GitHub API from lisp).

The list of LispUsers packages was derived from following the instructions in the "ReadMe before you print" file in lispusers/000-docs which said what things had to be loaded to get the lispusers package documents to print (except for the postscript font for Sigma) plus the idle-hacks I was reviving just for fun (did Interlisp invent the "screen saver as art"? predating Berkeley flying pigs?

masinter:
I don't know why TMAX doesn't load for you -- I tried running in a clean clone and didn't have any problems. There might be a problem that lispusers/TMAX was replaced by lispusers/tmax/TMAX one level down, and the way you're instantiating a directory from github might have a problem with the case shift? Just a thought.

rmkaplan:
I connected up to tmp/ and let it go.  So it appears that it is connected to the wrong place when it tries to load EXPORTS.ALL

It appeared to load everything up and do the database pass.

One glitch:  It stops when it loads TABLEBROWSER and asks if it is OK to redeclare its records.

From the transcript, now there is an ambiguous field between VT100SAVE and something in XCCS, also somethings in TMAX-ENDNOTE,  IDLEHAX, LIFE

masinter:
I like making the various files in /tmp/ because if there's a problem it doesn't mess up the working dir. But my idea is to address this by having "loadup-all" call "copy-all" at the end, putting the newly built /tmp/ files into the proper directories too.

then there's no ambiguity of where to load exports.all from.

I'm not sure but I think loadup-db.sh could start from a release (loadups/full.sysout rather than tmp/full.sysout) but I'll leave that for now.

Updated in PR now.

masinter:
The tablebrowser asking about redeclaration is worrysome.
for the ambiguous record fields, I'd like to change it so that ambiguous record fields give priority to declarations in the same file ... or else turn off MSRECORDTRANFLG so it doesn't look in the translation. The only problem is calls from ACCESSFNS records making the database less complete.
maybe another value for MSRECORDTRANFLG?

masinter:
I'm not sure but I think I handled these except for the dribble file(s). I wasn't sure what to call them. I also want to include the GATHER-INFO -1 -2 -3 which report problems.

rmkaplan:
I don’t usually run in the clone, I just use it as a way-station between git and my working medley.  So I had not recently made a loadup.

I did a loadup-all, and now it does find TMAX.

But then it can’t find EXPORTS.ALL, underneath loading SYSEDIT.

The EXPORTS.ALL does exist, as tmp/EXPORTS.ALL.  DIRECTORIES does not have an entry for tmp.  Is that the issue.  When it fails, the connected directory is tmp/logindir, not tmp.

Separately, I will want to run this on my working directory, not the clone.  I assume that that just depends on where the script is run from?

PR #799, opened 2022-06-22
Rmk48: Miscellaneous minor updates: TEDITFILE, GITFNS, EXTERNALFORMAT, EDITINTERFACE

rmkaplan:


masinter:
not quite related to this PR but you need to do a `get fetch` before you call `git log` because git doesn't actually download anything from the server until then.
image


image
image

masinter:
I think in the long run we'll want to separate out the initialization of the git projects from the code that does the work... so we can diverge on what projects we want.
And maybe reset after makesys.

masinter:
minor glitch -- when the 1106:229 on the right is selected, the left side has a section on the left still underlined.

masinter (approved):
The glitches with using prc now in issue #800.
This looks fine

PR #802, opened 2022-06-25
Medley-utils additions pick, hcfiles, new loadup-db

masinter:
MEDLEY-UTILS additions & changes
`HCFILES` fixes to avoid "bad files".
loadup now copies (hard links) from tmp to loadups creating new versions
(no need for copy-all.sh )

PICK function/command chooses something to work on 
`cpv src dest` copies one file to destination, but keeps older versions maintaining maiko conventions.


rmkaplan (commented):
I was able to run the loadup-db without failing.  I'm not sure that it is a good idea to keep the fuller.sysout--it's 55MB compared to about 11MB for a full.  Is it worth using up that much storage in a backup?  It wouldn't take too long to recreate the information, by the gather-info all, as needed.

Looks like the exports.all gets copied moved to library, but the where-hash, rdsys, and the database stay in tmp/

rmkaplan:
Also, do you mean to include BADFILES ?

PR #804, opened 2022-06-26
mark \ fns and SI fns as potentially unsafe to run interpreted

### LINKED ISSUE #441: SYSEDIT should  (or should not) set DFNFLG to PROP
I think the fact that it is NIL is why it has been unsafe to edit system code

masinter:
proposed fix #441


rmkaplan (commented):
Actually, I think warning about every slash function is going to be overkill, quite a nuisance.  Many/most slash functions are slashed because we don't want to guarantee their behavior and don't expect ordinary citizens to invoke them.  But most of them would be OK to break or advise.

I think slashing and safety are related but not closely aligned.  There are another bunch of functions that take you into the funky tty debugger when you break them, even if you break a call inside something that you know is safe.

It would be good to get this straightened out, my bet is that the slash filter is way to coarse

masinter:
newcomers can get into trouble just by flailing around. So I'd rather be cautious "Don't edit anything you don't understand".

 I've had trouble tracing OPENFILE for example, if the font used by the trace window hasn't been loaded -- get a break inside a break.  Some would argue this is a fundamental flaw in the Interlisp development model. But in practice it's just a small matter of keeping the heuristics up to date.

Err on the side of caution -- it just pops up a warning before it redefines the function which is easy to dismiss.

rmkaplan:
Maybe make it sensitive to a level-of-expertise flag, that defaults to idiot?

rmkaplan (approved):
Looks reasonable

PR #805, opened 2022-06-27
assume process streams are UTF-8

masinter:


rmkaplan (approved):
CLIPBOARD asserts its own externalformat on its process stream,also  currently UTF-8.  Maybe that now is needless confusion, that that is not a source of variability above and beyond what all process streams do?

nbriggs:
A process stream is only UTF-8 if the LANG family of environment variables says so.  It's not correct to blindly assert so.

rmkaplan:
Seems like UNIXCOMM should read the UNIX-GETENV and set the external format, that CLIPBOARD should then leave it alone.

Note:  If you are moving UNIXCOMM to sources, delete it from library.

nbriggs:
It’s not a fundamental part of the system, I think it should stay in library… ?

masinter:
it wasn't my intention to move UNIXCOMM, I'll add a check for the LANG environment variable and use it to determine the external format/ charset encoding (this perhaps should be a maiko GETPARM subrcall value but that's another story).

PR #808, opened 2022-06-27
MEDLEY-UILS loadup-db run-medley fixes

masinter:


rmkaplan:
I'm still having trouble with the NEW-SHELL variable, that UNIXCOMM seems to be the same as it was before.

rmkaplan:
Also, the tmp/login> ends up with a 57MB virtual memory image

masinter:
i got rid of the lisp.virtualmem.
There was some problem with the way UNIXCOMM was compiled that caused the DONTEVAL@LOAD DOCOPY expressions.

GITFNS shouldn't set up projects when loaded; aftermakesys / aftersysout forms or better clean up before and re-setup when first called. (will make another issue)

rmkaplan (approved):
loadup-all and loadup-db both seem to work, Process-stream is OK. vmem file is gone.

FMI, why did you switch to dfasl?

masinter:
> FMI, why did you switch to dfasl?
it might have been 'pilot error' (when I look at the file versions it seems likely) but it got rid of the error that (INITIALIZE-NEW-SHELL-DEVICE) wasn't called.
I think we should put in some checks when we do loadups, PR review or whenever to make sure the .LCOMs and .DFASLs are made from the sources we've checked in.

PR #813, opened 2022-07-02
pull more newer library lispusers internal(/library) files from envos

masinter:


rmkaplan (approved):
This looks harmless enough

PR #814, opened 2022-07-03
Rmk51 end game of external format integration

rmkaplan:
What was left a little ragged, a year or so ago, was how external formats get associated with file devices by default.  As part of the transition at that time, it were also fallback variables for the format functions if those fields hadn't been set for a stream or device.

This straightens that out.  When an FDEV is created, if an explicit format isn't specifed as the default for its streams, then the value of a single variable *DEFAULT-EXTERNALFORMAT* (currently :XCCS) is used.  So every device automatically has an external format.  And every stream fills in as default values for the method functions the functions in its device's DEFAULTEXTERNALFORMAT.  The implementation functions (/INCCODE etc) no longer work with missing method-fields.

Apart from changes to EXTERNALFORMAT, FILEIO, XCCS, this required recompiling device-creating functions on other files.


library/UNICODE
```diff
@@ -832,7 +833,7 @@
 (RPAQ? UNICODEDIRECTORIES NIL)
 
 (PUSHNEW UNICODEDIRECTORIES (PACK* (UNIX-GETENV "MEDLEYDIR")
-                                       '/unicode/xerox/))
```
masinter, line 835: `(PUSHNEW UNICODEDIRECTORIES (MEDLEYDIR "unicode/xerox")) `

library/UNIXCOMM
```diff
@@ -197,7 +202,10 @@ Copyright (c) 1988-1990, 2018, 2022 by Venue & Xerox Corporation.
                                     EOFP _ (FUNCTION UNIX-STREAM-EOFP-NEW)
                                     BACKFILEPTR _ (FUNCTION UNIX-BACKFILEPTR-NEW)
                                     GETNEXTBUFFER _ (FUNCTION UNIX-GET-NEXT-BUFFER)
-                                    BLOCKIN _ (FUNCTION \BUFFERED.BINS])
+                                    BLOCKIN _ (FUNCTION \BUFFERED.BINS)
+                                    DEFAULTEXTERNALFORMAT _ (AND (STRPOS ".UTF-8" (UNIX-GETENV "LANG"
```
masinter, line 206: should set it once every time a sysout starts... this is tricky

masinter (approved):
prc is hanging in a long irrelevant git command listing files that aren't part of this PR.
Inspection by viewing git diff from the web i left some comments.
Will merge and test after.

rmkaplan:
Is there a better way of doing this?  Should UNICODEDIRECTORIES be set up in the site init?

rmkaplan:
Probably should be done in a separate function that is put on AFTERSYSOUTFORMS.

BTW, what’s the difference between the shell-device and the new-shell-device?  Do we need both?

rmkaplan:
It was also hanging on BIN, even though the ENDOFSTREAMOP just have been returning NIL, before I put in the PEEK.  I think it is also not detecting fatal errors now, because of some new funny behavior.

And, I think that I now always crash when I close the file browser window after a prc.  Something has gotten a little loose.

nbriggs:
I think it should be set up in the site init, plus I don't like MEDLEYDIR anywhere but in the site init.

rmkaplan:
Even the first \PEEKBIN on the result of the process stream is flaky:  It returns the second character.

rmkaplan:
A further thought:  to fetch a file from the repo, maybe it would be better to avoid the Medley process stream altogether and tell the git show command to direct it to a local tmp file?  Then it could be read as an ordinary unix file?

PR #817, opened 2022-07-04
Rmk52 fix external format for error stream

rmkaplan:


masinter (approved):
hard to review while PRC doesn't work so well

rmkaplan:
There are still cases where it hangs on the process stream, when it is trying to read the final COMPLETED.  I can’t figure it out, maybe someone else will have a better understanding.

One thing I did observe that surprised, and screwed up some of my heuristics:  (EQ 0 (GETFILEINFO file ‘LENGTH)) is never true, even for an empty file.  The reason is that GETFILEINFO - LENGTH always returns a FIXP, even for a small (or empty) file.  You have to use (IEQP 0 ..).

PR #823, opened 2022-07-08
Rmk53 added system externalformat

rmkaplan:
Also, child of rmk52 for testing of prc rework


masinter (approved):
approving hoping this will clean up

PR #824, opened 2022-07-09
LLDISPLAY, WINDOW, BIGBITMAPS:  share common DSPCREATE, ensure formats

rmkaplan:
The common stuff is now all in LLDISPLAY, the variants are covered by arguments to the common


masinter (approved):
approving

PR #826, opened 2022-07-10
Rmk55  gitfns with redirect

rmkaplan:
GITFNS command strategy is the major target, a few other minor adjustments


masinter (commented):
prc works again.

We're gathering a lot of new medley features that need documentation and changes to the IRM? Not sure where they go but we should have a place to accumulate.

masinter (approved):
meant to approve

PR #827, opened 2022-07-10
Rmk56  filepos with external formats #3

rmkaplan:
See discussion at issue #3


masinter (approved):
should fit test cases into the test repo test cases.

PR #833, opened 2022-07-14
Tedit1 move tedit files to separate tedit subdirectory

rmkaplan:
I'm starting to track down Tedit problems.  First organizational step is to move all of the relevant files into a tedit/ subdirectory, like LAFITE, TMAX, with consistent file names TEDIT-xxx so that our 2-level loadup procedure will find them.

This required a minor adjustment  in the LOADUP filenames, so at the same time I've included the changes that enable UNICODE to come at the very beginning of the lisp.sysout loadup, and for the UNICODEDIRECTORIES to be set along with all of the other Medley directory variables.  (Parts of it eventually should go into the MAKEINIT).


masinter:
i wonder if we might have less trouble moving psudohost earlier....
The interactions between loadup-time and makeinit-time and before-sysout time and after-sysout time was hard to debug and get right.


I was working on TEDIT too by pulling TEDIT out of the ENVOS Repo -- I couldn't find anything later that didn't have the main flow INSERT-BRT calls commented out.

I got an older TEDIT to compile and even run interpreted except a RAID call when error in uninterruptible code (which I don't think should be uninterruptible, just protected by an unwind-protect / resetlst).

rmkaplan:
It might be good to move it earlier, because it is really quite useful in managing multiple projects and multipe versions.

But this particular problem, of trashing files when trying to create a directory on the Unix device, didn’t have to do with the loadup sequence.  I wasn’t trapping for an error that had never shown up before.  And with the fix that Nick made to Maiko, this particular problem won’t show up again.

I put out my modified version of TEDIT-FILE with a refactoring of TEDIT.BUILD.PCTB.  It appears that whatever I did here also made the simple rendering bug (ascii after Greek showing up as black boxes in XCCS) go away.  Although I’m not quite sure why.

PR #836, opened 2022-07-15
Tedit2 move tedit files to separate tedit subdirectory  attempt 2

rmkaplan:
Second attempt at reorganizing the files.  GITFNS was silently screwing things up when asked to create a new Unix directory; that was fixed in a Maiko branch I hadn't yet integrated.  These files should be correct.


masinter (changes requested):
I wish you wouldn't do this now. It messes up the ways I have of retrieving old versions. Making restore-versions.sh work with renaming is hard.

I'm not at all certain that we shouldn't roll back the jds changes in the 90s and reapply the few other modernizstions.

if you want them to be together in library, consider renaming pctree textofd and tfbravo so they all start with tedit.

rmkaplan:
I’m generally not in favor of backing up to the 90’s and trying to re-evolve the code.  It had problems then too, perhaps just different problems, and John presumably did make some forward progress.   I wouldn’t want to get old format files to work at the cost of getting new format files to fail.

What problem are you trying to solve?

(This is separate from the question of rationalizing the file locations. Inertia from old code is not by itself a motivation for not moving forward.)

masinter:
grep -lrwi teditdcl .

shows the following source files still reference TEDITDCL

sources/MAIKOLOADUPFNS
internal/MEDLEY-UTILS
lispusers/TEDITDORADOKEYS
library/TEXEC
library/lafite/LAFITETEDIT

masinter (approved):
I'll fix medleydir in a separate PR

rmkaplan:
I don’t see why MAIKOLOADUPFNS mentions TEDITDCL.  It just dummies out a bunch of functions with no reference to anything in TEDIT.  And the loadup goes through.

Maybe I should just remove that reference?

MEDLEY-UTILS just has a bunch of TEDIT files on one of the OK lists, not ust TEDITDCL but PCTREE, TEXTOFD, etc.

masinter:
fixed in PR #842

PR #839, opened 2022-07-16
loadup and run-medley script bug fixes

masinter:
* run-medley put back -n option
* loadup-all.sh put back leaving files in tmp (not calling copy-all)
* loadup-full.sh is now like loadup-all without loadup-aux (making whereis.hash and exports.all)
* run-medley -loadup has common logic for -nogreet (reset LOGINDIR).

### DIFF
--- a/run-medley
+++ b/run-medley
@@ -59,6 +59,10 @@ export LDEKBDTYPE=x
 while [ "$#" -ne 0 ]; do
     case "$1" in
 	"-loadup")
+	    # Keep (GREET) from finding adifferent init file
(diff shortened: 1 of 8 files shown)


rmkaplan (changes requested):
Typo in loadup-all.sh   edho instead of echo

rmkaplan (approved):
I tried as many as I could, all worked well.  The separate load-full and the separate copying are much bgetter.

nbriggs:
The canonical empty directory is `/var/empty`, rather than making your own empty directory, but I didn't get a chance to review this before it was merged.  Sigh.

PR #840, opened 2022-07-16
fix to HCFILES to pick up sketch files; bug fix sketch

masinter:


rmkaplan (commented):
I don't undertand the MEDLEY-UTILS conflict, maybe things got out of step?

The minimal SKETCH looks good to me.

masinter:
not sure why there was a conflict either but I just kept it to the sketch patch.

rmkaplan (approved):

PR #841, opened 2022-07-17
Part of fix to Issue #795.  Build/Push Medley Release workflow now auto-generates release changelog…

fghalasz:
…, allowing for release to be created in non-draft status.


masinter (approved):
don't think I can test this without merging it.

PR #842, opened 2022-07-17
fix oklibrary and use to not fail when file has been renamed

masinter:
the ok* variables list files that are 'ok' to load in to create single masterscope database.
probably a questionable practice ... loading all those files sources is very error-prone.
we should review, but in the meanwhile this pr


rmkaplan:
I see that you removed all the tedit files except TEDIT from the OK list (and put the FINDFILE test in).  Will just EDIT in that list cause the other files to come in (because TEDIT loads them)?

masinter:
> Will just TEDIT in that list cause the other files to come in (because TEDIT loads them)?

yes

>

rmkaplan (approved):
There are a few more references to TEDITDCL in other files, but these files look good to me.

PR #844, opened 2022-07-17
make tmp dir if not there; add README for scripts

masinter:


rmkaplan:
This may relate to the already merged precursor to this one.  loadup-db.sh now fails because it can't find EXPORTS.ALL when it is trying load SYSEDIT.

masinter:
The problem this fixes is with the automated release mechanism using GitHub Actions, which builds a loadup from an empty directory. THe fix was to add a "mkdir -p tmp" . It wouldn't affect anyone.

The problem with the missing exports.all is something else.

If you cleared your tmp directory, then I'd guess that 'loadup-db.sh' assumes you've run 'loadup-aux.sh' (which makes, in tmp, new export.all and whereis.hash).
I suppose loadup-db could just remake exports.all but that will be in a separate pr.

rmkaplan (approved):
LGTM

nbriggs (approved):
Seems fine.

rmkaplan:
The tmp/exports.all exists both in my working-medley and also in the git-medley directories.

loadup-aux/loadup-all work fine.

PR #845, opened 2022-07-17
FONT, EDITFONT: Cleanup strike font reading and writing

rmkaplan:
EDITFONThad its own slightly different version.  Now centralized in FONT with slight adjustments to internal interfaces.
Note that we don't have a way of writing AC font files, as near as I can tell


masinter (approved):
looks ok, not sure what the plan is.

PR #846, opened 2022-07-18
MAIKOCOLOR: Newer version than the one in master?

rmkaplan:
I have this one in my working directory, and it seems to be bigger than the one in the repo and its previous version is later than the version in the repo.  Maybe this is better?  I putting this out as a draft for people to look at.


masinter (approved):
yes, this is newer and much more complete

PR #848, opened 2022-07-18
GITFNS local-origin

rmkaplan:
I hope it's converging


masinter (approved):
I fear there will be a long tail of surprises because we're coming in at the "porcelain" level of interface meant for (geeky) humans, rather than tying into the "plumbing" level of the git/github API.
But the github/lisp/file package interactions are interesting at least.

rmkaplan:
Yes, we are sniffing around the edges.  For one thing, the logic for figuring merged and unmerged branches still isn’t right.  How would we get down to the plumbing—direct network access rather than these shell commands?

masinter:
I imagine modern API interfaces aren't more complicated than courier. But I'm not sure the Common Lisp crowd has gotten up to Python or JS -- not sure.

nbriggs:
git just pulls files from the server and puts them onto your local disk, then you look in the files in the .git directory to do what you need to do.

nbriggs:
(using the git plumbing commands to poke around in the files)
Best to have a look at, for example, https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain to see how things work.

masinter:
I think this PR changes to GITFNS(.LCOM) were included with rmk62 a23ce4272

PR #849, opened 2022-07-20
remove .IMPTR files (byproduct of hardcopy of tedit Interlisp Manual)

masinter:


nbriggs (commented):
These were all explicitly added by Arun on Dec 13, 2020.  Are we *sure* they're unnecessary?

nbriggs:
Reading the IMTEDIT.TEDIT -- I guess if we've made the TOC and INDEX then we don't need the IMPTR files, though they don't seem to be a particular problem.

masinter:
Every time you TEdit hardcopy a file, it regenerates the .IMPTR file.
It corresponds to the page numbers which might depend on the output format (Postscript, Interpress, Press) and the fonts used for that format.

Anzus:
Yup, they can be removed.

nbriggs (approved):
LGTM, then.

PR #852, opened 2022-07-20
unsafe.to.modify only warns if redefining compiled with interpreted (or in original UNSAFE.TO.MODIFY.FNS

masinter:


rmkaplan:
I tried editing a slash function that was initially \CCODEP, when I quit SEDIT it just redefined it, it didn’t ask.

nbriggs:
`UNSAFE.TO.MODIFY(FN OPTION)` won't complain if `OPTION` is non-NIL, and SEDIT passes "redefine" for `OPTION`.

nbriggs:
Actually, `UNSAFE.TO.MODIFY` seems to always be called with an option: "advise", "trace", "break", or "redefine" -- so unless the function is on the UNSAFE.TO.MODIFY.FNS list it's not going to object.

rmkaplan:
So, no warning from editing and redefining a slash function in SEDIT ?

masinter:
There's now OK.TO.MODIFY.FNS which can be T (just use UNSAFE.TO.MODIFY.FNS) or a list -- when you reply Y the first time, it will accumulate.  I think maybe we could possibly run Maiko with logging function calls and basically say it's unsafe to modify anything that is needed by a round-trip through the read-eval-print loop once as "unsafe to modify" but the main reason for the change was to make it OK to set DFNFLG to T in SYSEDIT.

rmkaplan:
After doing a lot of system work (= backslashed functions) with this for a bit, I think it needs an adjustment.  It is good to ask only when the new code is replacing a compiled definition, so the question doesn't get asked again.

But the question is asked in the system prompt window, typically far away from the SEDIT window that I am attempting to exit and that I am paying attention to.  So the SEDIT window hangs, while there is some unnoticed ASKUSER in the prompt window.  The only signal is that the bell rings when I try to type something in the Exec, thinking that the SEDIT is done.

So, can the askuser be put in the prompt window of the SEDIT, and not in the system prompt window?  At least there it will be more obvious.

rmkaplan:
If the promtwindow is hard, still better than now would be to use MOUSECONFIRM instead of ASKUSER.

masinter:
maybe SYSEDIT should set OK.TO.MODIFY.FNS to T -- if you're working at the system level.

PR #853, opened 2022-07-21
Rmk63  backccode peekccode eolc

rmkaplan:
This PR involves a fair number of files, but the main changes are in EXTERNALFORMAT, UNICODE, XCCS.

The main goal is to have name consistency with respect to EOL interpretation for the external-format input functions. \PEEKCCODE was an outlier, it is now \PEEKCCODE.EOLC, and \PEEKCCODE.NOEOLC is now just \PEEKCCODE.  This follows the pattern of \INCCODE(.EOL) and \BACKCCODE(.EOL).

At the same time, \BACKCCODE(.EOL) now returns the character code corresponding to the bytes it backed over, which makes for less error-prone logic.

In doing the renaming, some issues in calling code were also fixed.


masinter (approved):
This PR looks OK -- I don't understand the changes in EDITCALLERS.
I don't remember about the use of *BYTECOUNTER* as a free variable and wish there was something written about it.

rmkaplan:
The change to EDITCALLERS is so that it can call FFILEPOS without a case array.  The search algorithm that now respects character-encoding conventions can run at byte-matching speed if there is no case array but has to do all the character encoding/decoding, perhaps multiples times at each file position, if a case-array is provided.  It runs about 7 times slower in that case, as I recall.

The previous algorithm was using a case array to map all seprs/breaks into the same byte, and one of those bytes was packed onto the pattern when the search target was supposed to be a whole atom (and not just an internal substring of an atom).  I.e. to prevent MARKASCHANGED from being seen as a hit for the search “MARKAS”.  So the actual search would be for a string “ MARKAS “, with spaces on either end, and the spaces would then match any other sepr/break through the casearray.  (The (undocumented?) convention was that an escape (code 27) at one end or the other of the pattern would allow it to match such an atom-internal substring.)

The revised algorithm preprocesses the pattern and determines whether a pre or post sepr is required, and strips off escapes either end.  It searches without packing (e.g. just “MARKAS”) and without a case array (byte speed), but then at each hit, it decodes and examines the character preceding and/or following the match to see if they are seprs or not.  In other words, it only does the character decoding and sepr processing as a confirmation of each hit, after the fact.  (The assumption, BTW, is that CR and LF are in the same sepr class, so it uses the non-EOLC functions for those tests.)

On the BYTECOUNTER, with this revision to get consistent naming, I will now finish and put out the documentation.  (There is a template in Lispusers for documentation, with ENVOS in the header.  Do we want to fix that with Interlisp or Medley ?)

rmkaplan:
However, the original code had another oddity about the treatment of seprs that so far I have carried forward.

The characters that are treated as atom-markers are calculated by the function SEPRCASE, which returns the case array for the characters in a particular readtable.  But it also has an argument CLFLG which if T indicates that the characters in CLISPCHARS (+ should be put into the equivalence class along with the read-table's breaks and seprs.  CLISPCHARS has the characters ^/+=_:'~+~<>@!←↑.  Note in particular that his includes the single quote, which at least in the Interlisp readtable is not a break character--it is a read macro.

In EDITCALLERS, DWIMIFYCOMPFLG is passed as the CLFLG argument, and DWIMIFYCOMPFLG appears to be NIL in my full sysout.  The effect is that EDITCALLERS won't find 'XXX.

So, should DWIMIFYCOMPFLG be T?  Should CLFLG simply be passed as the constant T?  Should single quote be added separately to the equivalence class?  I'm not sure why the dependence on DWIMIFYCOMPFLG.

(Separately, I see also that a case array has to be passed explicitly to FFILEPOS (and thus the slow case) if the readtable is case-insensitive.  I'll fix that.)

PR #863, opened 2022-07-26
Remove bignum test from \INSERTTREE -- was just a debugging aid

### LINKED ISSUE #859: The BIGNUM declaration is not in EXPORTS.ALL
It is defined in LLBIGNUM, used at least by TEDIT.  Should be exported

masinter:
fixes #859 

This change was put in by me as a hack. \PUTBASEFIXP already checks against bignums. It is only because it was under an UNINTERRUPTABLY that it called RAID rather than signalling a Lisp error.

The code here could be improved but putting it back seems like the prudent approach-- something else is going wrong well before we get here.


rmkaplan:
I think Nick suggested keeping the check but testing against MAX.FIXP instead of type BIGNUM

nbriggs:
We should probably not merge debugging hacks into master.  If it's debugging code that others should try, it's easy enough to push a branch to github so they can get it... no PR is necessary, just announce the branch name.

Since `\PUTBASEFIXP` does the check for the right type and calls `\LISPERROR` if it's bad I'm OK with removing the extra check.

rmkaplan (approved):
LGTM

PR #864, opened 2022-07-26
Compile color files to allow color imageobj -- even though color doesn't work yet

masinter:


nbriggs (approved):
LGTM.  Fixes the "bad compiled function ..." from the outdated LCOMs.

PR #865, opened 2022-07-26
Fix HCFILES to skip some directories with large .txt files

masinter:
just a stress test of TEdit hardcopy


masinter:
moved to test repo -- will make change there

PR #867, opened 2022-07-29
recompile & add a .TXT desribing CONDITIONGRAPH in internal(was in internal/library

masinter:


masinter:
I'll combine this with more Lispusers cleanup.

PR #877, opened 2022-08-04
Eliminate duplicate declarations for TABLEBROWSER record, remove TABLEBROWSERDECLS file

masinter:


rmkaplan (commented):
This seems to be OK, but GITFNS is showing me some of the changed files but with no actual changes (e.g. DINFO).  Is GITFNS missing something, or are some of these have character differences that git picks up but that GITFNS properly realizes as having no significance?

masinter:
Just to make sure there wasn't a redefinition of the TABLEBROWSER and TABLEITEM records, I went ahead and did the UNSAVEFNS to mark some functions as changed so as to recompile them.
It would be good if PRC could make sure compiled files match the files they were supposedly compiled on.

rmkaplan:
There are some utilities in COMPAREDIRECTORIES (see the bottom of the .TEDIT) that might help with this. Is the question essentially whether SOURCE-FOR-COMPILED-P is T for file X and X.LCOM?

 But if one of them has changed and not the other, the git-diff would only show one.  Presumably, if there are 2 different source files but no different compiled files, then a compiling got missed. So something is wrong.

If there are two different compiled files but no diferent sources, but they were compiled on the same source, maybe that’s OK.

And, what should the flagging look like?

For starters, I suppose there could be a display mode that doesn’t show compiled files if they are each compiled on their corresponding sources, and the sources are showing up as different.  Then maybe you would not be distracted by a bunch of things that you can’t tell anything about.

rmkaplan (approved):
I think this is OK.  prc shows all of these key changes, plus some other differences that maybe have to do with commit that are ahead or behind.

But I think this looks good

PR #878, opened 2022-08-04
Files have been commited to test repo, remove from medley

masinter:


rmkaplan (approved):
This looks good from the Medley side, the files are deleted. The PR on git doesn't show that the files now show up in the Test repo, I assume that's a PR over there.

GITFNS prc shows other file differences between the different Medley branches, again we still don't have the right commands to find the right file differences (and even, for the "cob" command, the right branches.  But I approve this from the Medley side

PR #882, opened 2022-08-06
IBM EGA fonts renamed

nbriggs:
Rename the IBM EGA font files that @MattHeffron provided from old-style to current style so that they will be recognized by the font machinery.


masinter (approved):
tested using (DSPFONT '(IBM 14) PROMPTWINDOW) and holding menu down.

PR #883, opened 2022-08-09
GIT-INIT called after return to LOGOUT or SYSOUT etc; add option of making subdirectory to repo

masinter:


rmkaplan (approved):
Seems reasonable.  I have separate changes to GITFNS, I'll reconcile after this is merged.

One issue is that some of the "differences" that GITFNS put up were spurious, having to do with different upper/lower casing of CLISP keywords.   I wonder if COMPARESOURCES doesn't normaize those, are whether the COMPARETEXT should know about clispwords

PR #885, opened 2022-08-10
Make /usr/local/bin/medley run the right 'real' medley

tfeb:
The `Dockerfile` creates `/usr/local/bin/run-medley` which really just execs the real `run-medley`.  But it's missing a path component so it would try to run `.../interlisp/run-medley` say rather than `.../interlisp/medley/run-medley`.

This created file is not in fact used in the Docker image, so perhaps it should not even be created.  But if it is created it should probably be right.


nbriggs (approved):
LGTM.

PR #887, opened 2022-08-12
Rmk66:  2 little glitches, plus GITFNS prc dealing better with forking

rmkaplan:


nbriggs (commented):
In the VT100SAVE record, the CHARSET field never seems to be referenced, so does it need to be there?  If it's referenced by some other code outside of VTCHAT then presumably those should be updated in this revision to use the new name.

rmkaplan:
I did a scan and didn’t see it, but maybe there was a rationale.  So, convervatively, I at least left the cell.

library/VTCHAT
```diff
@@ -151,7 +153,7 @@ Copyright (c) 1983-1988, 1990, 1993, 2022 by Venue & Xerox Corporation.
 
 (DECLARE%: EVAL@COMPILE
 
-(RECORD VT100SAVE (CURSORPOS CHARATTR CHARSET))
+(RECORD VT100SAVE (CURSORPOS CHARATTR VTSCHARSET))
```
masinter, line 157: the CHARSET/VTSCHARSET field doesn't seem to be used. In ALLCONTENT.

masinter (approved):
After reallizing that I'd wandered into the second half of the PRC differences, LGTM

Issue #257, opened 2021-03-10
Things in LOADUP-FULL and where should they go


masinter:
Where do these belong (currently part of LOADUP-FULL)? Most came from Ron's init
* commands `cd` `pwd` `ls`
* FIXMETA (changes KEYACTION for BLANK-TOP)
* setting DEFAULTFILETYPE and DEFAULTFILETYPELIST
* pre-loading (some) fonts
* changing idle.profile timeout to 0
* setting `*DEFAULT-CLEANUP-COMPILER*`

Do these files belong in the full.sysout
* CHAT, UNIXCHAT (need Maiko changes) 
* THINFILES
* ISO8859IO (needed by UNICODE?)
*

nbriggs:
Some should be in site or user's greet file -- depending on whether you're doing mostly Interlisp or mostly Common Lisp work, what your idle screen preferences are, what your META key setup is, whether you like Unix-y commands, ..

I think CHAT and UNIXCHAT should be in the full.sysout, THINFILES perhaps up to the user (i.e., in their greet file), and I presume, based on the name, that we'll want ISO8859IO as we're moving to UNICODE.

masinter:
PR #889 handles the first item on the list.
we've not kept the rule about lispusers and library.

PR #889, opened 2022-08-12
misc lispusers changes -- UNIXYCD, lsee

masinter:


rmkaplan (commented):
Seems reasonable to get it out of the loadup files.  But is there a better home for something like this, in one of the source files, rather than a separate little addon file?

masinter:
just didn't belong in the loadup, and there wasn't any other place that came to mind.
The other bit left over is FIXMETA, which seems odd.

Maybe UNIXCOMM?  
I'd like to move ShellCommand renamed to UNIX-COMMAND or something.
GITFN has GIT-COMMAND which isn't quite right.
I could see 'cd' use the project list so I could do 'cd loops/truckin' ? not sure though.

Now that we're getting more users.... 

addresses some but not all of #257 move things in LOADUP-FULL to where they ~should~ go.

rmkaplan:
You can say cd {GLOOPS}/truckin or {WLOOPS}/truckin, i.e. using the pseudohosts that the medley-project defines.  You can also use the cdg and cdw commands, to specify the project without the brackets:  cdw loops trucking, cdg loops truckin.

Maybe there should also be a generic pseudohost LOOPS that is a synonym for GLOOPS, for people that don’t make that git/working distinction, or that gets set up instead when the directory in the GIT-MAKE-PROJECT doesn’t have “git-xxx” or “working-xxx” in the spec.

It is a little awkward, notationally, that this all goes through the host field of a file name, so you need the { }, instead of a pseudo-directory that gets popped off into the pseudo host.  Maybe it could be done instead by fiddling with unpackfilename and packfilename, when a directory doesn’t exist (check to see if a prefix is known as a pseudo directory).

masinter:
will redo this

rmkaplan (approved):

PR #893, opened 2022-08-14
Extend 'man' and other helpsys to (also) look in the Common Lisp HyperSpec

masinter:


rmkaplan (commented):
I don't see that this works.  I tried "man OPEN" and got a bunch of protocol information (attachhed) ending in an error "no such medley directory tmp/clh

Screen Shot 2022-08-14 at 11 53 00 AM
s

masinter:
try it again. It downloads and parses the index of the CLHS only once; from then on it just opens your browser to the right page. Getting it to work on all OSes will be a problem. I think they're touchy about creating derivative works but this is just running a browser.

masinter:
going to close this one and open a new one with fewer commits

PR #895, opened 2022-08-16
helpsys extend

masinter:
- Add Common Lisp HyperSpec and LispUsers/Library lookup
- call REPO.LOOKUP in more cases


masinter:
this still needs more work and depends too much on local configuration.

Although it is touchy I find it useful to remind myself of how things work.
The "opener" for HTML just calls the browser.
The "opener" for LispUsers and Library files works reasonably.
The IRM opener prompts for a window before it does the lookup -- not good.

Need to check notefiles and loops docs

masinter:
subsumed by new PR -- closing this one.

PR #902, opened 2022-08-20
More useful helpsys / other sources

masinter:


masinter:
man symbol should find the symbol in the CLHS (Common Lisp HyperSpec) if it's in the LISP package.
Should show the .TEDIT or .TXT documentation. It no longer opens a Interlisp REference Manual window if there is no match in the index.

rmkaplan (commented):
FILEPKG looks good, getting the silly status messages out of HELPSYS looks good.

But I tried man OPEN and CL:OPEN on the mac, it seemed to load the CLHS index, but then nothing happened.  Should something happen?

masinter:
there are different methods for firing up a browser depending on the host OS and also if you're running on a remote display (Windows with Linux, remote X connected to another machine). I've been trying it on linux and windows and mac with either direct or remote X server.   There are more possibilities (other specs, other access methods) but I just wanted a baseline to look things up that were ANSI Common Lisp rather than old CLTL (1 or 2).

masinter:
I've been making improvements. There are three sources of documentation: CLHS (Common Lisp HyperSpec), IRM (Interlisp Reference Manual) and REPO (a TEdit or TXT file).   If you give a string man "open" it will look in all 3. If you give an atom: if it's in one of the common lisp packages (LISP or XCL) it will look in CLHS; if the atom is in Interlisp it will look in the IRM (which doesn't seem to contain any Common Lisp). For atoms that are in both (RETURN, LIST, CAR ..) it will look in both.

In the repo depends on WHEREIS and FINDFILE ... if the atom is WHEREIS it will look for the filename.tedit or file.tedit for file-name.tedit following the medley conventions.

i did this because i needed it but it's early. A good example of the Interlisp development style.

masinter:
Making this work online will require some channel for the Lisp process to ship a URL for the browser to open.  Maybe something with the clipboard?

nbriggs:
No, it doesn't need that, it just needs a browser in the docker image that can open a window on an X display -- which will come up on the same display that you're viewing the medley window.  For example, in online, (CHAT "SHELL") and in that shell run x-terminal-emulator -- it will pop up a terminal emulator X window overlaying the Medley X window.  Chrome or Firefox, for example, would do the same.

masinter:
replaced by PR #817

PR #904, opened 2022-08-20
IDLE will LOGOUT instead of SAVEVM if ONLINEP

masinter:


masinter:
This PR changes the SAVEVM in the IDLE logic to instead do a LOGOUT if ONLINEP is not NIL.
It goes into IDLE and then N minutes later (via IDLE.PROFILE) it will LOGOUT) saving your VMEM.

I cleaned out some now-useless code for incrementally saving dirty pages; it was a little complicated because if you really have zero dirty pages you don't need to do a savevm.

When this is merged and released we can change the online logic to not advise IDLE so it won't hang.

(The Idler starts a new process)

rmkaplan (commented):
There is a typo in IDLER:  Vvmem should be Vmem or vmem

Otherwise, I think it's good

masinter:
I fixed the typo. There's a PR in the 'online' repo to take advantage of this (marked as draft)
https://github.com/Interlisp/online/pull/22

rmkaplan (approved):

rmkaplan:
On further reflection, I realize that I am unclear about when/whether the SAVEVM happens.  I don't think the default should be that it periodically happens, just because I left my machine idling for awhile.  If I have time machine running on a Mac, won't that always be copying the VM file to the backup, every hour or so, and thus using up a lot of backup storage?   If so, it shouldn't require a sophisticated user to turn off this behavior.

rmkaplan:
I'm now getting errors "can not open sysout file" in SAVEVM under \SAVEVMBACKGROUND, on my mac

masinter:
@rmkaplan I suspect the problem with SAVEVM might have had to do with running two instances of medley at the same time, and they were stepping on each other. If that wasn't the problem, then I'm barking up the wrong tree.

We need a way to SAVEVM or LOGOUT on a timeout for running the Online service. It was probably a mistake to change the default behavior for users who weren't running online, at least without discussion.  In the D-machine era, SAVEVM was being done in the background with the \DIRTYBACKGROUND on BACKGROUNDFNS (which is now a no-op).

rmkaplan:
That wasn’t what I was thinking of.  I just wasn’t sure that there was any value in doing the SAVEVM’s automatically and by default, and I didn’t think it should silently gobble up a lot of back-up disk space for an unsuspecting time-machine user (who hadn’t thought to put the lisp virtualmem files in an unbacked-up place).

It might be nice to have an easy way of switching this on (set a variable in your init file), but it shouldn’t be on by default (except online).  But even online, I don’t quite understand why doing this in the idle is a useful thing, when there are more explicit ways of saving state (SYSOUT, LOGOUT).  Is the scenario that someone might go away and expect to be able to come back, without actually explicitly saying they want to save a state to come back to?

Doing this with several images running simultaneously is a separate important issue.  I do do that, particularly so I can compare the behavior of old and new Tedits, and so trashing the VM’s is definitely a bad thing.

masinter:
It has more to do with the way the Lisp session is lost when the connection to the display is disconnected.  With X, Lisp is the client and your display is the server and nothing is saved.
This happens when my PC goes to sleep -- the WSL process connection to the Windows process breaks because the TCP connection went to sleep. With VNC it's different: the display is the client and Lisp is on the server, and if you get disconnected, you can just reconnect.  BUT we don't want all these lisp processes hanging around for days without a connection. So if there's nothing going on, log out.
How can you tell nothing's going on? That's what IDLE does. Frank was just advising IDLE to LOGOUT but apparently that locked up the mouse waiting for an event that never happened.

Besides, I think the idle hacks are works of art that deserve a rotating idle-hack-at-random virtual gallery display.

Yes, there's a variable you can set in your init, IDLE.PROFILE. `man IDLE.PROFILE`

nbriggs:
I believe it wouldn't be too hard to handle the error that ldex will get when the X server goes away and do a SAVEVM before it exits.  The only issue is that on your local machine,  if you're quitting X because you want to throw away the session, it would save it instead.

rmkaplan:
Although I'm not enthused about having SAVEVM done automaticallly under an idle, there is one situation where an implicit SAVEVM might be very useful.  When the system now puts up the question about unsaving a dangerous function (slash, or on an unsafe list), and the user answers "Yes", it would be nice to save state just before the redefine.  If things fall apart, you can restart and answer No instead.

PR #910, opened 2022-08-22
move doctools from toplevel to internal/im

masinter:


masinter:
broke loadup for unclear reasons -- redoing

PR #911, opened 2022-08-23
Move doctools to be in internal/im and fix references to it

masinter:
- move doctools from toplevel to internal/im
- update medleydir to not markaschanged


rmkaplan (commented):
In PICK, should DOCTOOL be DOCTOOLS ?

Does the \MEDLEYDIR.BEFORE test solve the problem of starting up the sysout with the same user on the same machine?

Should DIRECTORIES in MEDLEYDIR include NIL to indicate the LOGINHOST/DIR ?

masinter:
I think the \MEDLEYDIR.BEFORE test might address the problem, since it doesn't change anything if the value of MEDLEYDIR doesn't change. (And it also leaves it alone if you're restarting from a LOGOUT or SAVEVM since those aren't supposedly moveable.)

about adding NIL ... I'm not sure, I think it might be added in some contexts anyway 
(CONS NIL DIRECTORIES) or something. More of a change than I wanted to do without more investigation.

masinter:
This was a small step toward reducing the overall complexities of directories.
Maybe it's too little with not enough merit to be with the hassle?
I'd like to get to the state where there is just sources, library, lispusers (internal?) but fewer directories and a clear correlation between what goes into the loadups and where files are.

rooms, obsolete, cltl2, notecards, loops, greetfiles? should we just do it all in one or continue piece-by-piece?

masinter:
marking as draft and will resubmit

PR #912, opened 2022-08-24
thie problem only shows when you switch EDITMODE but have some SEDIT windows open

masinter:


rmkaplan (approved):
Seems reasonable.  But it begs the question:  is this the place for EDITDATE to cause the window to be redisplayed with the date-update as the edit session is ending?

masinter:
redisplaying after edit date is updated probably should go here...
if we're keeping edit dates in the same place that we've always kept them.
maybe in 2022 we should put who edited what somewhere else? Then we wouldn't have to worry about this.

PR #913, opened 2022-08-24
Add loadup of masterscope database of the world; fix yml bug

masinter:


masinter:
i added loadup-db.sh to the workflow automation, to include a masterscope database of everything.
it adds a lot to the loadup time (5 minutes rather than 35 seconds to do a loadup.all). I wouldn't think it should run on every push -- just daily if there are multiple pushes to master only.

rmkaplan (commented):
Including the full database is good, worth the extra 5 minutes.

But putting it in the release library when the release library doesn't include all the other library files creates a hassle in constructing a full release environment.  The library files are a separate folder, so you have to be careful if you want to merge the exports and the database with the other files.  You can't just move the folders.  If you are going to put these in the loadup library without all the other files, then you should at least also include them in the separate all-file tar file, so you can just dump one on top of the other.

masinter:
There had been a problem where a file was repeated in both the -runtime and -loadups tar files.
I think that particular problem had been addressed in the latest 220812 release.
However ... I do think we need to move things around a bit to clean up. Here are some ideas for discussion:
* remove inactive files from the repo -- 'obsolete' 'fontsold' 'fonts/altofonts' -- maybe do this last
* keep 'loadups' for all results of doing a loadup -- lisp.sysout whereis.hash exports.all fuller.database *.dribble
* move *.venuesysout out of the repo
* add a new folder to hold starter.sysout
* move rooms to be under library
* separate legacy lispusers (updated but made 20 years ago) from modern stuff

Then we'd have

* sources (used to make lisp.sysout)  
* sources-init (only used by makeinit?)
* library (loaded into full.sysout)
* library-new (modernize et al)
* lispusers (only those that have been tested)
* lispusers-unsorted (to be moved into lispusers or tossed)

greetfiles -- just in a lispusers/init folder, rename MEDLEYDIR-INIT to INIT-MEDLEYDIR

change the loadup process to just make the files where they should wind up, instead of making them in tmp and copying.

I think we should talk about requirements .

I suggest we take up this discussion separately after discussion but for now, this PR should be approved.

masinter:
This PR has comments about the config issue category, but the main thing it does is add another target of "loadup-db.sh" which creates a sysout (fuller.sysout) which gets built and not released and a fuller.database.  I'm OK with not doing this, but ... either approve or request changes or close, don't leave PRs unreviewed for so long.

masinter:
hold off on this until there's more discussion of it...

PR #917, opened 2022-08-27
Update HELPSYS to find CLHS (Common Lisp HyperSpec) and lispusers/library docs

masinter:


masinter:
This adds features to HELPSYS to look for information not just in the IRM but also other places -- for LISPUSERS or LIBRARY features ... to look for a TEDIT file for the name you're looking up, and (for Common Lisp symbols) to look up in the Common Lisp HyperSpec.  It also fixes the problem that "man xyzzy" prompted you to open a window only then to tell you that nothing was found.

I think we might want to mount our own copy of CLHS following their License restrictions just for our own stability? Not sure.

It would be nice to have an SEDIT keystroke that would invoke this on the selection, and the user experience should be better, but i find this useful as it is.

rmkaplan (approved):
FILEPKG just seems to be a comment change, commonlisp HELPSYS worked well

PR #918, opened 2022-08-27
a better fix to the ttystealing problem

masinter:


rmkaplan (commented):
I'm confused.  It seems like the effect of this is to remove the TTY.PROCESS in the RESETLST in the master.  But the comment in the master about this is dated 8/23/22 (as is the red version in the github pr-remove display.  So, which way is this going?

masinter:
I'm not sure how things got this way, but I'll close this one, and we'll look again.

PR #919, opened 2022-08-27
change loadup to make and save fuller.database

masinter:


masinter:
Tried to replace #913 because git complained about it, but this PR has the same problem so I'm closing it

PR #891, opened 2022-08-13
set up list for classifying files without moving them -- for gather-info

masinter:

PR #927, opened 2022-09-06
avoid merge error, use newest MEDLEY-UTILS

masinter:


masinter:
I redid #891 to avoid the MEDLEY-UTILS conflict (which I will close).

masinter:
This conflicts with the 'move-cd' branch which hasn't been merged. I'll just mark it as draft and try to separate out the changes.

masinter:
Don't worry about the conflicts -- the same files are part of another commit; can be cleaned up when the merge is done.

masinter:
The main intent of this PR was to move the system of noting separate lists of files what are "OK" (can be used, has been checked out, loads, compiles etc), with separate lists OKFILES OKLISPUSERS ... to having one list of (root) file names and their associated state.

Maybe there's a better way to do this? This i just a proposal for a method which seemed promising.

Issue #923, opened 2022-08-29
multiple medley instances on the same machine step on each other's lisp.virtualmem


masinter:
it's easy to wind up starting multiple instances of medley. They can really interfere with each other writing lisp.virtualmem at SAVEVM time or LOGOUT etc. This isn't a problem with online (it's harder to get an online) or when it is only invoked manually. 

People complain about Medley cluttering up their login directory...  
also about time-machine backing up a vmem file if you write one every 10 minutes.

I'd like to get more frequent vmems to help us diagnose crashes.

There are some issues about having open files complain if you restart a vmem made when it was loading and you restart in a different directory.

shell commands inherit PWD from the launching shell rather than the lisp one.

masinter:
i'm experimenting with a version of run-medley that takes a -timestamp option.
When you specify -timesstamp, it will generate

lisp.virtualmem.220906222453 instead of plain 'lisp.virtualmem' (or whatever your -vmem was.)

when you start a run-medley without specifying a sysout to run, it will use the most recently written (using `ls -t lisp.vmem`*).

the idea is that each time you run-medley there's a unique lisp.vmem file for that run.
(need to add a 1-second delay in cases where the lisp.vmem.220906... file already exists).

There should be some kind of automatic trim when you exit Lisp normally ? 
In the version-vmem branch.

https://github.com/Interlisp/medley/compare/master...version-vmem

not ready for prime time but wanted some ideas about how to approach the problem.

masinter:
@fghalasz solved the problem anogher way (the --id flag o medley startup).

PR #928, opened 2022-09-08
add a '-timestamp' option to run-medley that timestamps lisp.virtualmem

masinter:
It seemed like we were discussing overlapping concerns with IDLE, LOGOUT and SAVEVM in a PR that had been merged (#904) and then reverted (#922) to address an issue about shutting down online sessions that had been closed, and a related issue #923 that arises when you are NOT online but instead have open multiple sessions on the same machine and them stepping on each other.
 
* Behavior should be configurable simply (e.g. by IDLE.PROFILE)
* don't clog the user's disk or backups
* Distinguish between abandoning a session when falling asleep or called away, vs. closing the X window
* ONLINEP images  (LOGOUT) saving VM if session is abandoned (current behavior)
* Matching D-machine experience (background SAVEVM) is a plus (running two sessions = borrowing one of the pool Dorados)
 
I think some variant of this (draft) PR to timestamp the saved vmems in the file names, using the most recent if starting up with no other qualification.

### DIFF
--- a/run-medley
+++ b/run-medley
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/sh -x
 #               Run Medley
 #
 # Syntax: run-medley [-noscroll]                 #turn off scrollbars
@@ -114,16 +114,32 @@ while [ "$#" -ne 0 ]; do
 	    mem="-m $2 "
 	    shift
 	    ;;
+
         "-vmem" | "--vmem" | "-vmfile" )
             export LDEDESTSYSOUT="$2"
             shift
             ;;
+
+	"-timestamp" )
+	    savedvm="$LDEDESTSYSOUT."`date +%y%m%d%H%M%S`
+	    if [ -f "$LDEDESTSYSOUT" ]; then    # already a lisp.vmem
+		latest=`ls -t ${LDEDESTSYSOUT}* | head -n 1`
+		echo saved  $savedvm    latest   $latest
+		export LDESRCESYSOUT="$latest"
+		ln "$LDEDESTSYSOUT" "$savedvm"  # link it to the dated
+		export LDESRCESYSOUT="$savedvm" # 
+	    fi 
+	    export LDEDESTSYSOUT="$savedvm" 
+	    ;;
+	
         "-full")
             export LDESRCESYSOUT="$MEDLEYDIR/loadups/full.sysout"
             ;;
+	
         "-lisp")
             export LDESRCESYSOUT="$MEDLEYDIR/loadups/lisp.sysout"
             ;;
+
         "-n" | "-new" | "-newfull" )
             export LDESRCESYSOUT="$MEDLEYDIR/tmp/full.sysout"
             ;;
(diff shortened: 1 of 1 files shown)


masinter:
Can more of the logic here can be done in Lisp?
If there are multiple vmem files around and you didn't supply one as your LDESRCEYSOUT (or on the command line to RUN-MEDLEY) pick the latest one? (by write date) choice needs to be done before lisp starts, but it doesn't matter what suffix was used -- could just use version numbers. 

note that the write order might not be version number order: start A start B logout B logout A, then restartshould restart A not B.

fghalasz:
I think we are conflating two issues that should be orthogonal to each other:  1) the issue of auto savevm/logout kicked off by idle and 2) the issue of savevm/logout when there is more more than one instance of Medley running at a time.  Seems like the solution offered here is really about the second issue - which exists whether or not we do auto savevm/logout during idle. So I propose we use this Issue/PR to discuss the multiple simultaneous instances problems/solutions.  And use Issue#904 to discuss the issues surrounding auto savevm/logout during idle.

I don't think the use of of timestamps to distinguish lisp.virtualmems is particularly useful.  Usually when you are running multiple run-medleys, you know why - e.g., one is running master, one is running dev branch and one is for quick tests.  So why not use labels for the various run-medley instances.  For example: `run-medley --vl master ... & run-medley --vl dev ... & run-medley --vl test ...`.  And if you don't specify a --vl, then it would default to, say, --vl default.

run-medley would check to make sure that no other process has _lisp.virtualmem.\<vl value>_ open for write (e.g.. using lsof).  If not it would then immediately open for write (not exactly atomic but good enough for these purposes) and use that as LDEDESTSYSOUT.  If _lisp.virtualmem.\<vl value>_ was already open for write, then run-medley would exit with appropriate error message to the user.

This would mean that with no --vl, users can only run one run-medley at a time.  And with --vl, they can only run one run-medley with a given label at a time.  I think this would cover most common use cases.  And for folks like Ron who never use SAVEVM nor LOGOUT without T, we could have a flag (and corresponding environ variable) that just skips all this checking and happily starts N run-medleys without worrying that they might overwrite their lisp.virtualmems.

On startup, run-medley would just use the _lisp.virtualmem.\<vl value>_ as LDESRCESYSOUT unless LDESRCESYSOUT is otherwise specified.

fghalasz:
Also some quick comments on the code in this PR

1.  You deleted the section that started with `if [ -z "$LDESRCESYSOUT" ] ; then`.  But in the case where no --timeline flag was  given, this significantly changes how run-medley works.  I didn't dig into it deeply, but at the very least this section should go back in, possibly with a !--timeline guard.
2.   Nit:  #!/bin/sh -x

masinter:
@fghalasz 

> I think we are conflating two issues that should be orthogonal to each other: 1) the issue of auto savevm/logout kicked off by idle and 2) the issue of savevm/logout when there is more more than one instance of Medley running at a time. 

The only connection is that 2 was happening because I'd changed the default IDLE.PROFILE in dealing with 1. 

> I don't think the use of of timestamps to distinguish lisp.virtualmems is particularly useful. 

I've been running with it and I agree. Too long and hard to decipher.
The subr that implements savevm (the vmemsave subr) takes a path name so I'll experiment with that.

run-medley adds a lot of complexity that is quite off-putting to those getting started.
All of these environment variables! I'm going to close this PR and try something else (the VMEMSAVE subr takes a filename argument that we're currently not using).

masinter:
that was hasty, but maybe this is an opportunity to redo the run-medley / lde / LDESYSOUT environment variables into something more sensible. When I first saw that, it was awfully complex.
Perhaps a new executable target in addition to the lde / ldex / ldeinit ones that would have a sensible behavior?

nbriggs:
We should think about what it should do… let’s design it before rushing off to change it.  I don’t have the lde source in front of me, but will look at what’s there when I get back. Logically there’s the image you’re starting from, which may be a fresh one or resuming the last state, and the place it’s going to save the current state on exit, which is not going to be on top of a fresh startup  image. Beyond that there are explicit saves to a given name for sysout and makesys operations.

PR #932, opened 2022-09-11
Attempt to address poor 'empirical' algorithm for deciding when to break

masinter:


nbriggs (commented):
Please check that the functions you are changing packages on are not system defined… interesting-frame-p IS a Common Lisp function and should not be only in IL:

masinter:
I wound up not changing the package of any of the functions. I was going to, but thought better of it.

To check CSBROWSER(DEBUGGER DEBUGGER;3) (version numbers from 'restore-version.sh) but essentially the latest vs. master
image

The different packages for symbols in a single file does seem like at least a style error -- it will be harder for common lisp users used to a different set of conventions.

I can find no reference to interesting-frame-p or enter-debugger-p in the Common Lisp HyperSpec.

masinter:
the trigger for calling debugger is a little too sensitive now... cd to a non-existent directory breaks.
I think for newcomers it's better to err on the side of 'more breakpoints', but still...
The PR has an error so I'll mark it as draft and resubmit a cleaner PR

masinter:
The medley debugger has some settings that need reworking.
HELPDEPTH needs setting after fixing up INTERESTING-FRAME-P , as does BAKTRACELST to compensate for changes to how the stack looks in a normal EVAL or CL:EVAL.
It looks like the last time this was updated was Interlisp-10.

PR #936, opened 2022-09-13
Idle program display fixes

masinter:
@rmkaplan 

the savevm in idle is a setting on IDLE.PROFILE: TIMEOUT says "how long idle until it goes into IDLE program" and the "SAVEVM" setting is "How long after that does it SAVEVM" with a value of 0 meaning "never".

The default has been that it's off. I'll turn it on. You can leave it off.
What should be the default for new users?

If IDLE.PROFILE is (TIMEOUT 10 SAVEVM 1) it will start an idle program display after 10 minutes, and, after a minute if you don't type something to stop the idle program, it would sleep and savevm.

This is how Windows and Mac OS and my Amazon Fire TV work now, and how it worked when we were using Dorado's and Dandelions.

Submitting as draft -- need to insure idle functions don't eat CPU for online users

'


fghalasz:
For the sake of clean modularity, we should try not to use ONLINEP within core code.   So far we have treated online as just another "site" and any special things about online are contained in the ONLINE-INIT file.

I propose we add a new entry to IDLE.PROFILE called LOGOUT.  LOGOUT would be like SAVEVM except the idler would logout rather than just savevm after the given number of minutes.  If LOGOUT is 0 or NIL or absent, then it has no effect.

If both SAVEVM and LOGOUT are present and non-0, non-NIL in IDLE.PROFILE, then if SAVEVM < LOGOUT, the two would execute in order else if LOGOUT < SAVEVM, only LOGOUT would be executed else if SAVEVM=LOGOUT, then do LOGOUT.

In this setup, ONLINE-INIT could just set IDLE.PROFILE:LOGOUT and IDLE.PROFILE:SAVEVM as appropriate.  The default INIT-MEDLEYDIR could set IDLE:PROFILE:SAVEVM to whatever we deem appropriate for casual users and :LOGOUT to NIL.  And Ron's personal INIT file could set both to NIL or 0 (or remove both).

fghalasz:
Also, there is an issue of resetting various timers after returning from logout.  Currently with the ADVICE we are using online (and it looks to be the same in the changes in this PR), when you return from the auto-logout within IDLER (i.e., by restarting the saved lisp.vmem), you return into the "same" IDLE process and that IDLE process will never auto-logout again.  I would expect that the IDLE process if left untouched for the SAVEVM timeout, would auto-logout once again.

Bottom line: need to add code to that resets the SAVEVM (and LOGOUT) timers after a return from auto-logout.

masinter:
@fghalasz I agree with your proposal for SAVEVM and LOGOUT on IDLE.PROFILE and will do something about timers on return. 

 There are a couple of ways that the system prevents a LOGOUT if you have files open for a MAKEFILE or LOAD that it's not clear to me should be addressed. Maybe PAGEFULLFN should have a timeout also?

There's a lot of logic around passwords checking that I'm not sure of the continuing utility. Keep it? I suppose why not.

fghalasz:
An optional timeout on PAGEFULLFN would definitely help for online.  Should be a global variable and not an argument on call to PAGEFULLFN, so it can be set in ONLINE-INIT.  Also (and this might be tough) once the timeout is hit, it should be considered hit for all of the subsequent page fulls for the calling operation.  For example, if a page full timeout hits on the first "page" of a LOAD, all subsequent page fulls for that LOAD should just progress immediately.  This way a 1 minute timeout doesn't turn into a 20 minute total timeout for a LOAD with 20 "pages" of output.

Taking out the password checking stuff seems like it might be destabilizing for not much (no?) benefit.

masinter:
@fghalasz  Look at lispusers/PAGEHOLD for some considerations on the subject of paging.
I tried at one time to include that in FULL.SYSOUT to meet with a vigorous objection, but the document at lispusers/PAGEHOLD.TEDIT raises some issues to think about.

masinter:
main functional change: add option to IDLE.PROFILE to SAVEVM or LOGOUT (or both, if SAVEVM is < LOGOUT).
add IDLE.RANDOM: choose at random among other options loaded.

masinter:
replaced by #948

PR #937, opened 2022-09-13
Rmk68:  Fix GITFNS PRC file lists, plus a few minor fixups

rmkaplan:


masinter (approved):
i don't know how useful it is to leave the code "wjem nil" rather than just deleting it.getting the warning is irritating so LGTM.

lispusers/GITFNS
```diff
@@ -1571,31 +1607,41 @@
...
+                             (* ;; "Let the directories figure it out")
+
+                             (AND NIL (IF (EQ (CADDR GFILE)
+                                              100)
+                                          THEN 
 
                                  (* ;; "A little tricky to figure out what corresponds to the real file in the mapping, which directory it belongs to.  Maybe the first one should always be one that exists, the second may just be a useful name.  But we have to know whether to match against INFO1 or INFO2")
 
-                                      (PUSH MAPPINGS (LIST (FULLNAME F1)
+                                               (HELP GFILE 100)
```
masinter, line 1622: really?

sources/BREAK-AND-TRACE
```diff
@@ -124,46 +124,45 @@ IL:|{DSK}<usr>local>lde>lispcore>sources>BREAK-AND-TRACE.;1|)
...
-                                                                        XCL:*TRACE-LENGTH*))
-                                     IL:DO (PRINT-TRACED-ARGUMENT $ARG-COUNTER$ $ARGUMENT$ 
-                                                      $$INDENT$$))
-                                  (PRINT-TRACED-ARGUMENT ',ARG-LIST ,ARG-LIST $$INDENT$$))))))
+                                  (IL:FOR $ARGUMENT$ IL:IN ,ARG-LIST IL:AS $ARG-COUNTER$ IL:FROM
+                                                                                         1
+                                     IL:WHILE (OR (NULL XCL:*TRACE-LENGTH*)
+                                                  (<= $ARG-COUNTER$ XCL:*TRACE-LENGTH*))
+                                     IL:DO (PRINT-TRACED-ARGUMENT $ARG-COUNTER$ $ARGUMENT$ $$INDENT$$
+                                                  ))
```
masinter, line 157: ??? what is this fixing?

sources/BREAK-AND-TRACE
```diff
@@ -708,25 +699,23 @@ IL:|{DSK}<usr>local>lde>lispcore>sources>BREAK-AND-TRACE.;1|)
...
+                                   (UNWIND-PROTECT
+                                       (XCL:DEBUGGER
+                                        :FORM
+                                        `(FUNCALL ',#'$BRKEXP$)
+                                        :ENVIRONMENT NIL :STACK-POSITION $POS$ :CONDITION
+                                        ,(OR XCL:CONDITION `(IL:LOADTIMECONSTANT
+                                                             (XCL:MAKE-CONDITION 'BREAKPOINT 
+                                                                    :FUNCTION ',IL:FN))))
+                                       (IL:RELSTK $POS$)))
+                                IL:ELSE ($BRKEXP$))))
```
masinter, line 719: I thought the main thing to fix in BREAK-AND-TRACE was that when you said (BREAK (MUMBLE :IN FOO)) that it was FOO that might be unsafe to break or advise and not MUMBLE.

sources/DWIM
```diff
@@ -944,35 +945,39 @@ Copyright (c) 1982-1986, 1988, 1990-1991, 2021 by Venue & Xerox Corporation.
...
-                                             ((LISTP VAR)
-                                              (CAR VAR))
-                                             (T VAR)))
-                                   (PROG (TEM)
-                                         (LISPXPRIN1 "****Warning: the iterative statement:
+  [LAMBDA (X)
+
+    (* ;; "Edited 13-Sep-2022 09:16 by rmk:  Disabled this warning about a historical transition.  The EDITFINDP overshot anyway--would find occurrences other than VARS (e.g. field names).")
+                                                             (* wt%: "24-MAR-80 08:23")
+    (CL:WHEN NIL
```
masinter, line 952: EDITFINDP?

sources/INSPECT
```diff
@@ -111,22 +117,21 @@ Copyright (c) 1982-1987, 1990-1991, 1993, 1995, 1999, 2018, 2021 by Venue & Xero
...
-                      (SETQ VALUEMARGIN (\INSPECTW.VALUE.MARGIN
-                                         (COND
-                                            (PROPPRINTFN (for PROP in PROPERTIESLST
-                                                            collect (APPLY* PROPPRINTFN PROP 
-                                                                               DATUM)))
-                                            (T PROPERTIESLST))
-                                         IWFONT))
+                      (SETQ VALUEMARGIN (\INSPECTW.VALUE.MARGIN (COND
+                                                                   (PROPPRINTFN
+                                                                    (for PROP in PROPERTIESLST
```
masinter, line 124: it's distubing to see the pretty printer mangle things so badly that looked OK before.
No chance to review these using lisp's PRC, just saying

sources/INSPECT
```diff
@@ -1303,13 +1360,14 @@ Copyright (c) 1982-1987, 1990-1991, 1993, 1995, 1999, 2018, 2021 by Venue & Xero
     (for PROP in (PROPNAMES ATM) when (NOT (FMEMB PROP SYSPROPS)) collect PROP])
 
 (INSPECT/LISTP
-  [LAMBDA (LST WHERE)                                    (* ; "Edited  2-Feb-87 17:05 by jop")
+  [LAMBDA (LST WHERE TAG)                                    (* ; "Edited 12-Sep-2022 20:49 by rmk")
```
masinter, line 1363: is there an example of how TAG is used?
Should there be some 'release note' that explains new capabilities so others can explore?

masinter (changes requested):
only reviewed one file, too soon

rmkaplan:
> i don't know how useful it is to leave the code "wjem nil" rather than just deleting it.getting the warning is irritating so LGTM.

I can find the caller and rip out the call, then take out the warning function.  I was just a little lazy

sources/INSPECT
rmkaplan: > is there an example of how TAG is used?
> Should there be some 'release note' that explains new capabilities so others can explore?

We should figure out a release-note strategy, to track extensions and revisions in general.  A single document, a collection of notes in a well-defined place?

As for tag, there are 2 ways:  give a tag as the optional argument in the call to INSPECT, or use the title button-function to add a tag to an existing inspect window.

A separate issue:  INSPECT current returns NIL, so that there is no way of manipulating the inspect window after the fact (e.g. to have a program that move it around in a systematically helpful way).  Perhaps INSPECT should return its window?

sources/BREAK-AND-TRACE
rmkaplan: > I thought the main thing to fix in BREAK-AND-TRACE was that when you said (BREAK (MUMBLE :IN FOO)) that it was FOO that might be unsafe to break or advise and not MUMBLE.

There are more things to fix there, but break and debug are really very opaque now.  I wanted to do a simple thing:  put the name of the broken function in the title of the debugger window.  I found where the title appeared to be asserted, but I couldn't figure out how to do it.

In this case I think I was just trying to get the filemap to be accurate

rmkaplan:
I removed WARNUSER and its calls (DWIM and DWIMIFY)

rmkaplan:
Revised INSPECT so it returns the window (IRM says it should, that got lost along the way)

rmkaplan:
> > i don't know how useful it is to leave the code "wjem nil" rather than just deleting it.getting the warning is irritating so LGTM.
> 
> I can find the caller and rip out the call, then take out the warning function. I was just a little lazy

I did this

masinter (approved):

Issue #58, opened 2020-10-27
set up key bindings to match Mac, Windows, Linux, Emacs (at user's preference


masinter:
"help wanted" is different than "good first time project". In this case, what we need is someone to research what key bindings are now and what they should be. Also for window actions, menus, etc.

I don't think the fact that Interlisp pre-dated Mac and Windows and Linux UI guidelines is important to preserve

_Originally posted by @masinter in https://github.com/Interlisp/medley/issues/50#issuecomment-713083301_

masinter:
Between TTYIN, sedit, tedit compared to Linux, Mac, Windows, emacs set up user GREET settiable (undoably) consistent set of key bindings. It is impossible to be simultaneously consistent with multiple inconsistent things, but settable key handling should allow these:
* arrow keys (left, right, up/previous, down/next)
* DEL as "delete selection" or (if nothing selected) delete forward
* CR/Enter as (submit line) even if not at end
* control-A as "select all" or "go to beginning of line"
* control-K as "delete rest of line"

For extra credit: allow the input stream to be a stdin / pipe and output stream to be a terminal with character IO

masinter:
* This issue has some overlap with other keyboard issues (European keyboards) but hopefully separable
* everyone has opinions about what their fingers are trained for. We need to make reconfiguration easy (even though the underlying mechanisms are complex.
* Some things are not implemented (deleting forward, arrow keys) and it isn't just a question of key bindings
* we should mainly look at what the browsers do because
* online users are running in a browser
   * browsers face the same developer dilemma of needing to support and document multiple different user preferences
https://www.groovypost.com/tips/50-universal-web-browser-hotkeys/
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent
https://learn.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/customize/shortcuts
https://support.apple.com/en-us/HT201236
https://support.google.com/chrome/answer/157179?hl=en&co=GENIE.Platform%3DDesktop

masinter:
those links were pretty useless. Here's the beginning of a chart
https://docs.google.com/spreadsheets/d/1FOkrr62TtEhhY49m9U0T_6bvqSGRQt9fBRtMDw0YKtY/edit#gid=642562870

masinter:
I made good progress on the keystroke binding table:
[...]

PR #942, opened 2022-09-16
Fixes and cleanup in HISTMENU, CLIPBOARD, and MANAGER

MattHeffron:
Update HISTMENU with Exec target check & cleanup
Add to CLIPBOARD support for M-x (Extract) in SEdit 
Fix the file "See" commands in MANAGER


rmkaplan (changes requested):
Arming meta-X (or meta-x) as "extract" in SEDIT conflicts with the existing behavior wherein meta-x means "expand"--substitute the code for a given function in place of the call.

It would be nice if we could synchronize the meta-keys in Tedit and Sedit with what we now see in modern interfaces, but we have to figure out where to move the existing bindings.

I added meta-X in Tedit (and also meta-z (undo), meta-a (all), and meta-g (find again)) because they did not conflict.  I held off on Sedit, because we have to decide where to move the current meta-x behavior.  Similarly, Sedit meta-a means abort, not all, and meta-z means mutate, not undo.  meta-g is safe, I just haven't gotten around to it.

But we have to make a decision about how to shift the other key bindings.  Meanwhile, and unhappily, I think this should be saved for later.

masinter:
https://github.com/Interlisp/medley/issues/58 Want consistent keybindings (as part of Modernize)

I had started a table of key bindings https://drive.google.com/file/d/1jFxxEkZW3yKY5m7Bp8pjzkWeF3CYevp2/view
but I'm having trouble opening it.

fghalasz:
@rmkaplan  > I added meta-X in Tedit (and also meta-z (undo), meta-a (all), and meta-g (find again)) because they did not conflict.

Adding meta-x, meta-z etc to to TEdit is modernizing it only if you consider the MacOS modern.  For the vast majority of the modern world (i.e., Windows users and - to some extent - Linux users) the modernized bindings should be ctrl-x, ctrl-z, etc.

rmkaplan:
There may be a different set of conflicts in the Windows world then in the Mac, but still some issues.  ctrl-X “extract" also conflicts with the Sedit “exit”, for example.

Would we do any better if we tried to map Tedit and Sedit onto emacs key bindings, and piggy-back on however conflicts have been resolved there?

nbriggs:
> Would we do any better if we tried to map Tedit and Sedit onto emacs key bindings, and piggy-back on however conflicts have been resolved there?

No.   For example, in emacs (in text mode) to copy a region:  move to a point, ctrl-space to set mark, move to another point, esc-w to copy to mark-to-point region to kill-ring, move to another point, esc-w to paste.
In emacs in window mode with mouse integration (on a mac) you can use meta-x, meta-c, meta-v; or you can switch to "CUA mode" and use ctrl-x, ctrl-c, ctrl-v for cut/copy/paste-- and then it gets a little weird, because if you have some text selected ctrl-x copies it, but if there is no text selected then ctrl-x is a command prefix (and ctrl-x, ctrl-c is still exit... as long as you don't have text selected).   Yuk.

In Linux shells, it's ctrl-shift-x/ctrl-shift-c/ctrl-shift-v but in other apps it's usually ctrl-x/ etc.

masinter:
Issue #58 is the "consistent keybindings" issue. I'll start another chart of existing keybindings (with and without 'modernize') against different Lisp SEdit TEdit and TTYIN editing.

In the meanwhile, the function belongs where this PR puts it. I'd rather see it bound to some keystroke rather than unbound. If there was an easy option to move 'expand' from 'meta-X' to 'control-meta-X' and usurp meta-X for this function, I'd put that in my init until we have a more general solution.

masinter:
See Discussion https://github.com/Interlisp/medley/discussions/637 on bundling different edits into one PR or not.

MattHeffron:
I created a new Pull Request that excludes the changes to Clipboard due to the issues of conflicting key bindings.

PR #944, opened 2022-09-20
Histmenu+Manager+Clipboard branch WITHOUT the changes to Clipboard

MattHeffron:
(since there were issues with conflicting key bindings)


masinter (approved):
Thanks, looks good

PR #948, opened 2022-09-20
IDLE.PROFILE has LOGOUT and SAVEVM options; add some delays in IDLE hacks

masinter:
This PR replaces #936 with some patches to idlers that were too fast these days.


rmkaplan (commented):
I loaded all these files on top of an existing master system.  When I clicked IDLE in the background menu, I got "Venue" floating around on the screen.  I wasn't sure whether randomly I should see other effects, but I didn't.  When I slid-off to see the profile, it showed up far away, in the system promptwindow, where I didn't notice it until some time later.  I assumed I would see some sort of inspector-like pop-up.

masinter:
I updated the PR but mainly just to reduce the differences.
IDLER is in the loadup and it sets up various defaults with an (INITVARS ...). So loading a new version on top won't change things.

The default idler now bounces around your user name. To see other idle hacks you can load IDLEDEMO (which in turn loades all the lispusers packages that fiddle with IDLE.PROFILE

fghalasz (approved):
LGTM.  Reviewed code changes, loaded idle-patches on top of latest master., and ran it thru it paces.  IDLER changes all work as expected.  

Lots of the various idler packages (Hanoi, Kinetic, etc) have rough edges .  And some (not included in this PR) lead to breaks - e.g., GlidingBox.  But this should definitely not hold up this PR.  Fixing all these idler hacks should be very low priority IMHO.

masinter:
The Idle hacks amuse me, but it's a good demo of what was fun about Interlisp-D...
the easy access to raw bitmap graphics and ... were there any screensavers before them?

fghalasz:
Didn't mean to step on your fun.  Only meant to say that I didn't spend much time fully testing and evaluating the operation of the many idle hacks.

PR #949, opened 2022-09-21
better breakcheck

masinter:
- Attempt to address poor 'empirical' algorithm for deciding when to break
- Add EXTERNAL-SYMBOL-P for INTERESTING-FRAME-P
- CONTROL-T called XCL::INTERESTING-FRAME-P which was moved to IL package
- tweak function name packages, INTERESTING-FRAME-P
- more tweaking for backtrace and interesting frames


masinter:
This still needs tweaking. There are lots of stack frames and perhaps counting them isn't the right useful data.  HELPTIME is hampered because most of the time HELPTIME = 0 is appropriate; setting it to 1 and most things complete sooner.  We could possibly add a microseond clock rather than a millisecond one, but is the time elapsed the right thing to be measuring?

There are lots of stack frames that don't make sense to me... OLDFAULT1 NEWFAULT1...

masinter:
#953 redoes the PR

Issue #497, opened 2021-09-22
change Lisp source & compiled file names in github to lower case


masinter:
### Discussed in https://github.com/Interlisp/medley/discussions/15

Moving back to an issue

<div type='discussions-op-text'>

<sup>Originally posted by **masinter** August 17, 2020</sup>
Filenames for Interlisp files are mainly upper case.
@nbriggs: except for the ones that people started writing from XCL, where they tended to use lowercase.

Unix and Windows default to lower case.
> Or at least their users do.

I think Interlisp has a model where it will match either.
 
> It's messier than you might think -- Unix file systems (outside of MacOS) are
> usually case-sensitive.  MacOS is case-insensitive and case-preserving (by
> default), but you _can_ mount file systems case-sensitive.
 
Ron has warned about some kind of whereis, filepackage problem, but
that could be solved by changing the files since case shift for ascii
characters in filenames doesn't change the lengths.

Second, we should have .tedit and .sketch as well as .interlisp or
.il .lisp (save for common lisp) .ilisp .medley are all possibilities.
With filetypes that know about medley.

> Isn't there a bunch of code that knows that Interlisp source files don't have an
> extension?

I don't think there's a bunch. 

</div>

masinter:
closing combining duplicate issues in  #265

PR #957, opened 2022-09-26
Add .gitattributes for binary files.

MattHeffron:
Add .gitattributes so *.TEDIT, *.LCOM, *.DFASL, and *.SKETCH are always treated as binary (and the lowercase versions).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,9 @@
+# Denote all files that are truly binary and should not be modified.
+*.tedit binary
+*.lcom binary
+*.sketch binary
+*.dfasl binary
+*.TEDIT binary
+*.LCOM binary
+*.SKETCH binary
+*.DFASL binary


nbriggs:
Should we add .wd, .strike, .displayfont, .sysout ?

masinter:
at some point I did a 
`git config core.autocrlf false` and that made some problems go away.
Lisp source files shouldn't be autocrlf'd to anything. The Interlisp source code has embedded ctrl-M as well as embedded ctrl-J LF codes.

masinter:
i think mainly 

git config -global core.autocrlf

will eliminate a number of git problems. This was a problem we had early on with Lisp sources using CR instead of LF, but the autocrlf repair made things worse, not better, and so... do these two:

git config --local core.autocrlf false
git config --global core.autocrlf false

fghalasz:
@masinter  "git config" is always per user (or per machine in case of --system).  It doesn't follow the repo as you clone it, etc.
So you autocrlf:false trick only works in every contributor has set these it their git environment(s).

It seems that it would make more sense to use .gitattributes to set all lisp source files as "not text" so git doesn't mess with any cr/lf substitutions on them - no matter what the contributors git config settings are.

Unfortunately lisp source files in our case are marked by the absence of an extension and gitattribute file pattern matching is a restricted form of globbing, not regex.  Saying the absence of an extension is not possible in gitattributes.

Instead we would add the lines: `*  -text` followed by `*.* text` at the top of the .gitattributes file. 

This would set all files with an extension to "text" (which I think is the default), leaving all files without an extension to "-text".  Then the following lines that Matt is proposing would turn most of them to binary, leaving .TXT and .md (to name two) marked as text.

masinter:
We went around on the issues of file names several times...

https://github.com/Interlisp/medley/issues/497
https://github.com/Interlisp/medley/discussions/15
https://github.com/Interlisp/medley/issues?q=is%3Aissue+is%3Aopen+file+extension+

fghalasz:
@masinter Not sure why you bring up the file name discussions.  My point was to handle CR-LF for lisp source files - as they are currently named without extensions - in .gitattributes rather than in .gitconfig.

.gitattributes is a property of the repo while .gitconfig is a property of the user/machine.  To ensure consistent treatment of lisp source files by git across all users, its better to enforce the cr-lf specifications at the repo level with git attributes.

The issue is that .gitattributes doesn't allow negative assertions (like absence of an extension), so you need to be slightly - and only slightly - tricky about how you get files without extensions covered properly.

masinter:
Sorry for the leap: one of the original motivations for adding an extension to Lisp source files was precisely this problem -- we couldn't use the extension in .gitattributes. 

But there are conflicting desires -- for the purpose of CRLF mangling, Lisp source files should be left alone. But for GitHub web site comparisons -- which we were stuck with until GITFNS -- it was better to leave them as text. 

I'm thinking of a repo of PDFs of Lisp sources (PRETTYFILEINDEX) and TEDIT => PDF conversions too.

masinter (approved):
there are other issues brought up in the discussion, but this PR is fine.

PR #963, opened 2022-10-01
WIP changes for dodo or other 10MB XNS

masinter:


sources/LLNS
```diff
@@ -475,12 +472,12 @@ Copyright (c) 1983, 1984, 1985, 1986, 1988, 1989, 1990, 1991, 1992, 1993 by Venu
...
 (STOPNS
-  [LAMBDA NIL                                            (* bvm%: "17-FEB-83 15:57")
+  [LAMBDA NIL                                                (* ; "Edited  1-Oct-2022 16:28 by lmm")
+                                                             (* bvm%: "17-FEB-83 15:57")
     (\DEL.PACKET.FILTER (FUNCTION \HANDLE.RAW.XIP))
     (\DEL.PACKET.FILTER (FUNCTION \HANDLE.RAW.10TO3))
     (DEL.PROCESS '\NSGATELISTENER)
-    (CLOSENSOCKET T)
-    (SETQ \NSFLG NIL])
+    (CLOSENSOCKET T])
```
nbriggs, line 480: Why is it not right to set \NSFLG NIL when you stop NS?

masinter:
There are two variables: whether XNS is available and whether it is wanted. Also for PUP.
There are two other variables, whether '10mb' ethernet is available and also 3mb.

I think the subr call reports available but not wanted. This (draft) PR uses \NSFLG as "wanted" and \10mbflg for 'available' but that's a change that i'd want to look at what uses (Lafite, chatserver, etc) expect.  The code we started with in 2020 wasn't a good guide -- it was done as a (broken) turn-off-xns patch after delaying startup waiting for a server to reply to a 'what's my pup number' to reply.

nbriggs:
Perhaps you need to look back to the 1990 or so versions of the code before it got hacked to disable the ethernet.  The deal was, if you had an ethernet enabled in the code (that is, you started ldeether instead of lde, and it would build up the interface to the /dev/nit or the streams interface) then \ether.available was true.   If you don't start it with either a raw ethernet or the nethub code turned on then \ether.available should be false and it should not try to start the network.

We also need the \10MBWATCHER (or was it the NS routing watcher?) back, and the dodo code should be sending periodic routing packets so that the Lisp code picks up a network number automatically.   I need to check if dodo has a time service as well.

Even with the modified network code you have on this branch things don't work well at startup.   The first thing I tried was (NS.ECHOUSER BROADCASTNSHOSTNUMBER NIL NIL 10) and it hung, or got no response.

masinter:
I don't think we have sources between the Interlisp-10 1984 SUMEX sources and the Envos/Venue sources of ~1993.   There is a file LLETHER with \ETHEREVENTFN calling (COND (\GATEWAYFLG (\INIT.GATEWAY))) and several others that use \GATEWAYFLG to decide whether to forward packets, but I can't find anything in the Interlisp/history or Interlisp/envos repositories that set \GATEWAYFLG or define \INIT.GATEWAY (at least by grep -lri init.gateway ... Similarly for \STASH.PASSWORDS for a definition other than MOVD?(NILL \STASH.PASSWORDS) which was done in the lisp.venuesysout.

are there other snapshots that aren't in the 'history' repo that I should be looking for?

nbriggs:
LLETHER of 19-Jan-1993 sets \GATEWAYFLG to NIL, which is what it should be, in an INITVARS.

One should only set it to T if one wants one's machine to forward XNS and PUP packets -- which is only going to be in really specialized cases.  I believe that the code in LLETHER provides a hook for a gateway function, but does not provide an implementation because you really don't want someone who doesn't know what they're doing to start forwarding packets.   It is still a good hook to have, though.

masinter:
There's <lispusers>RS232cnetwork by someone named Briggs with a copy of \ETHEREVENTFN with the same line (COND (\GATEWAYFLG (\INIT.GATEWAY))) but no definition of \INIT.GATEWAY.
😄

nbriggs:
Yes.  That's what I would expect.

masinter:
trying again going back

PR #965, opened 2022-10-03
COMPAREDIRECTORIES:  upper-case-file-names

rmkaplan:
Plus typo fixes in the TEDIT file.

Note that the MATCHNAME has always been uppercase, and that the directory matching has been filtered using the FILEDIRCASEARRAY


fghalasz (approved):
LGTM.    Works as expected.

PR #970, opened 2022-10-06
dodo4

masinter:
- move Maiko versions of functions into place vs MAIKO.MOVDS
- pick up a few MAIKO.MOVDS that were missed


nbriggs (commented):
Why is a change to IDLER in this branch?

sources/10MBDRIVER
```diff
@@ -74,47 +73,41 @@ Copyright (c) 1982, 1983, 1984, 1985, 1986, 1987, 1988, 1990, 1991 by Venue & Xe
              PACKET])
 
 (\10MBSENDPACKET
-  [LAMBDA (NDB PACKET)                                   (* ; "Edited 14-Apr-88 15:30 by bvm")
```
nbriggs, line 77: ETHERLIGHTNING is really handy for debugging behavior in flaky network situations.   Please don't take this out.

sources/10MBDRIVER
```diff
@@ -74,47 +73,41 @@ Copyright (c) 1982, 1983, 1984, 1985, 1986, 1987, 1988, 1990, 1991 by Venue & Xe
...
-    (PROG ([DROPIT (AND \ETHERLIGHTNING (EQ 0 (RAND 0 \ETHERLIGHTNING]
-           IOCB BUFLENGTH)
+  [LAMBDA (NDB PACKET)                                       (* ; "Edited 31-Oct-89 14:10 by bvm")
+    (PROG NIL
           [COND
              (\RAWTRACING (\MAYBEPRINTPACKET PACKET 'RAWPUT]
           [COND
-             ([AND (NOT (fetch NDBCANHEARSELF of NDB))
```
nbriggs, line 83: Note that Dodo isn't the ONLY way to send and receive ethernet packets -- it still matters whether the "real" ethernet interface (/dev/nit, /dev/bpf, DLPI) "can hear itself".

nbriggs (commented):
I'm not seeing any explanation of why each change is the right thing to do.  It's really difficult to reason through the changes with no explanation.  Is this really just taking functions that were previously Maiko-specific (and conditionally moved over the old definitions) and replacing the old code directly in the source?

masinter:
It wasn't conditionaly... I was trying to simplify by doing this

nbriggs:
OK.  I would have explained it for the less knowledgeable person who reads the commit log as

"Various functions that are specific to the D-machine ethernet implementation were replaced by Maiko-specific versions during loadup (?) via the "MAIKO.MOVDS" mechanism [see MAIKO.MOVDS in file xxxx].  This commit replaces the the D-machine implementation by the Maiko implementation in the source."

then nobody is going to be left wondering why COMPARESOURCES and other tools are reporting large changes without there being any explanation.

nbriggs:
You could even provide the list of names of Maiko-specific functions that replaced the D-machine functions...

masinter:
subsumed by #975 and others

PR #973, opened 2022-10-08
notify \IDLING.OVER in \IDLE.OUT

masinter:
For reasons not well understood, sometimes after returning from LOGOUT, the RESETSAVED notify to \IDLING.OVER doesn't happen. This change insures that it does. \IDLE.OUT is a backgrround function. 
(found during testing of LOGOUT/return from LOGOUT with changing ethernet enalbing)


rmkaplan (approved):
Looks like all it does is to insert the notify.  I don't know how to test this, but it seems reasonable.

PR #975, opened 2022-10-08
replace definitions of original D-machine versions

masinter:
"Various functions that are specific to the D-machine ethernet implementation were replaced by Maiko-specific versions during loadup via the "MAIKO.MOVDS" mechanism [see MAIKO.MOVDS in file MAIKOETHER. This commit replaces the the D-machine implementation by the Maiko implementation in the source." See
image


masinter:
This is a PR reworking #970

masinter:
i'll try this some other way -- perhaps separate commit for each MOVD

nbriggs:
If you do it with a separate commit for each MOVD, when you go to merge it... make sure you do it as a "merge" and NOT a "squash-and-merge", or the separate commits will all go away and be turned into a a comment line in the commit log message for one single big commit.  There are times when squash-and-merge is the right thing to do, and others when just plain merge is right.

Issue #972, opened 2022-10-07
Looks like a typo in MODERNWINDOW


masinter:
![image](https://user-images.githubusercontent.com/1116587/194638137-c975ee96-1e50-41d9-8fd1-ef7a81aa7dff.png)
the comma doesn't seem right %,TITLEPROPORTIONS 3rd line from bottom.

PR #976, opened 2022-10-08
MODERNIZE:  Fix bug in MODERNWINDOW  (Issue #972 )

rmkaplan:


masinter (changes requested):
The "right" fix is not to construct a lambda expression using backquotes, but use CL:LAMBDA and compile with the common lisp compiler. In Common Lisp, you have real lexical scoping, so the function object of a lambda expression will 'capture' the storage for the titleproportion.

rmkaplan:
I don't want to get into compiling with the commonlisp compiler, if that would make this dependent on fake-compile-file or cl:compile-file.

masinter (approved):
ok  then

PR #977, opened 2022-10-08
run-medley should not quote patterns in 'case' command

masinter:
"quote" around pattern is either useless or broken


nbriggs (approved):
LGTM.  I don't believe it matters either way, since no patterns we're matching on have spaces or other characters that need quoting.

PR #978, opened 2022-10-08
run-medley has a -NF option in caps used by loadup, means no fork

masinter:
bug fix -- it was introduced by accident.  loadups will fail without this fix.


nbriggs (approved):
LGTM.

PR #979, opened 2022-10-09
Fix simple typo in UNIXCOMM

masinter:
simple typo prevents new unix shells; introduced 6/28/2022 by me.
Found while looking for something else.


rmkaplan (approved):
LGTM

PR #984, opened 2022-10-10
when constructing a stream, it might not have a FDEV -- allow it to print

masinter:


sources/FILEIO
```diff
@@ -668,22 +666,29 @@ Copyright (c) 1981-1993, 1999, 2020-2022 by Venue & Xerox Corporation.
...
+                                                                                            )
+                                                                                 of STRM)))
+                                                  ((TYPE? FDEV (FETCH DEVICE OF STRM))
+                                                             (* ; "Name the device")
+                                                   (CONCAT TYPE " "
+                                                          [CL:STRING-CAPITALIZE
+                                                           (STRING (fetch (FDEV DEVICENAME)
+                                                                      of (fetch DEVICE of STRM]
+                                                          " Stream"))
+                                                  (T (CONCAT TYPE " unknown stream"])
```
nbriggs, line 691: perhaps " unknown Stream" (capital S).  But is "unknown" the right word, or should it just be " Stream"?

nbriggs (changes requested):
Other than resolving the comment,  it looks good.

sources/FILEIO
masinter: just Stream

nbriggs (approved):

nbriggs (approved):
LGTM.

PR #987, opened 2022-10-11
Change SYSTEM-EXTERNALFORMAT to more accurately guess the external format

masinter:
More importantly, it cannot return NIL.
This problem was detected trying to CHAT(SHELL) on Cygwin.
This change is incomplete (need BEFORE and AFTER events when possibly moving from one platform to another) but that will be done separately.


rmkaplan (approved):
Seems as reasonable as anything.  Separately, the events should smash the new system default into the file device.

PR #989, opened 2022-10-11
Unixcomm make *SHELL-DEVICE* the shell stream used, delete pre-Medley 1.x code

masinter:
UNIXCOMM had two file devices *SHELL-DEVICE* and *NEW-SHELL-DEVICE* where "NEW" was before Medley 2.0
Consolodate on *SHELL-DEVICE* using "NEW" definitions and remove unused functions and variables.


masinter:
will try approaching this some other way

masinter:
no matter what direction we will go, this isn't on path, so closing this after all.

PR #994, opened 2022-10-13
Fix Issue#985 HELPSYS/CLHS.LOOKUP fails when MEDLEYDIR not writeable.

fghalasz:
This fixes  #Issue#985: HELPSYS/CLHS.LOOKUP fails when MEDLEYDIR not writeable.
In CLHS.LOOKUP, output of CLHS.OPENER shell command is now "/tmp/clhs-warnings-$$.txt".
It was "$MEDLEYDIR/tmp/clhs/warnings.txt".


masinter (approved):
Thanks 👍🏽

PR #996, opened 2022-10-14
make SETQ and typed in calls undoable (first steps)

masinter:
A small amount of investigation about the state of UNDO..  originally UNDO worked as a step in the  Exec (aka REPL / command line / shell) -- it would substitute operations starting with / which were undoable versions... RPLACA would become /RPLACA when it ran. Changes didn't descend into functions ... just in the expression that was typed -- unless those functions were carefully written to be undoable.  Somewhere in the IRM (I think) there was an explanation.

In Medley 3.5 there are two ways to make something undoable -- both of which are broken. There is a macro UNDOABLY which is used by Interlisp execs which attempts to do the substitution using a code walker . There is a function UNDOHOOK which is used by the (Common Lisp CLTL1-only EVALHOOK) to make the substitution . (Note that CL:SETQ and IL:SETQ  are different, (CL:SETQ will CL:EVAL, handles multiple pairs while IL:SETQ doesn't know about lexical scoping and probably ignores expressions after the first); while CL:SETF (which can set other things -- more like 'replace') is preferred modern usage.
The brokenness was minor -- both of UNDOABLY and UNDOHOOK have a IL:SELECTQ (or CL:CASE) that select on (CL:SETQ CL:SETQ CL:SETF) -- like it was intended to work on both CL and IL versions of SETQ but the packages got mixed up. Changing one of the CL:SETQs to IL:SETQ makes those undoable.
The other breakages were calls to MEMBER instead of ASSOC when looking things up in LISPXFNS.

A lot of this environmental code was important for newcomers especially those coming from rigid declaration systems -- authored by wt: in 1978 or earlier and not updated or reviewed since.
The effort to bring medley up to speed with both newer Interlisp and Common Lisp will have a long tail.


masinter:
image
THis is pretty obscurbe -- when typing into the TELETYPE editor, entering ?= <return> gives an odd message. Interlisp macros were intended more like 'optimizers' than macros... and expanding Interlisp macros might have been an error recovery mechanism.

I'd like to "clean up" the EVAL stack so that macros don't go through FAULTEVAL and not get confused with DWIM.

rmkaplan (approved):
I included it in a loadup, SETQ is now undoable.  So, it is an advance

PR #997, opened 2022-10-16
Update \SENDMESSAGE.RESTARTABLE usage of OPENSTRINGSTREAM

nbriggs:
\SENDMESSAGE.RESTARTABLE unconditionally used OPENSTRINGSTREAM on its argument, which is not usually a string. Now, only use OPENSTRINGSTREAM if the argument is a string, otherwise pass it unchanged to TEDIT.


rmkaplan (approved):
LGTM

nbriggs:
Thanks. I’m away from computer, and this will make it possible for the dodo testing to press on.

PR #998, opened 2022-10-18
Rework MEDLEYDIR before/after logout to substitute instead of reset

masinter:
This PR modifies the handling of MEDLEYDIR initialization, save and restore over LOGOUT, SAVEVM, SYSOUT and MAKESYS to more often do the "right" thing, in the situation where directories and environment variables change.
LOADUP-INIT.LISP is loaded into the "starter.sysout" to run MAKEINIT. It needs to load the new version of MEDLEYDIR even though starter.sysout had already loaded an earlier one.

MEDLEYDIR uses a helper MEDLEY-INIT-VARS that was previously a function called during loadup and greet. It's been put into service as a EVENTFN where the events are one of `BEFORE*` `AFTER*` and `AFTERDO*` for * in SYSOUT, LOGOUT, SAVEVM, and MAKESYS (`AFTER*` returns from a restart, while `AFTERDO*` happen when continuing -- there is no `AFTERDOLOGOUT` and `AFTERDOMAKESYS` is unexpected.

Things should continue to work as expected when doing a LOADUP, starting from a loadup, a SAVEVM, a LOGOUT, changing MEDLEYDIR environment variable, changing the LOGINDIR environment. Allowing other directory setting variables to participate in the scheme. Very little of this was automated in the "old" days.

The previous method was to just reset variables to new values based on the initial values in the variable MEDLEY-INIT-VARS (same name as the function that did the processing).

This PR reworks it so that MEDLEY-INIT-VARS is a EVENTFN, and, in appropriate events, substitutes the new value of MEDLEYDIR for the old within strings and atoms for variables listed in MEDLEY-INIT-VARS.

 `MEDLEY-INIT-VARS` is (now) a list of (`var` `defaultvalue` [ RESET ]) tuples. When starting up or first loading, each `var` is set to (EVAL `defaultvalue`).  But the previous variable values are saved in BEFORE and restored with string substitution of initial strings (or atoms) matching MEDLEYDIR to the new value. (The RESET token means always re-evaluate, which happens with LOGINDIR-dependent variables). 

There are also hacks to deal with adding tmp to DIRECTORIES and WHEREIS cache files, which are retained for now.


masinter:
I looked at supporting GITFNS... but it doesn't handle {UNIX} files with / in the names -- I only tested for {DSK} and > delimited. But testing is hard -- I'm trying renaming the medley repo and passing the file between WSL Linux /home/larry/medley and cygwin /cygdrive/c/Users/larry/ilisp/medley while changing my LOGINDIR in various configurations.

rmkaplan (changes requested):
MEDLEYDIR has a redundant GLOBALVARS declaration at the top.
You might use STRING.EQUAL instead of STREQUAL.  It uses a casearray, defaults to UPPERCASEARRAY, but FILEDIRCASEARRAY is case-insensitive and also equivalences < > and /.

I think MEDLEYSUBSTDIR has a bug, you should ADD1 to NCHARS OLD.

rmkaplan (approved):
This looks good

rmkaplan:
Actually, I made a fresh loadup with after this was updated, and it still doesn't do the right thing.

In particular, now it wipes out my directory initialization when I start from the fresh makesys.  Before, it only did that when starting a sysout that was made later.

So, essentially, after having started up the makesys and having been greeted, I immediately have to reload my init file.  DIRECTORIES has all the medley directories, but it doesn't actually have my loginhost/dir, or anything else that I specfically set up.

masinter:
@rmkaplan please check out PR #1003 -- I can't quite tell from your description what you're trying to do.

rmkaplan:
I ran scripts/loadup-all.sh to create a full.sysout makesys (in tmp).

I then started that up in the ordinary way, specifying my ordinary greeting file. 

My greeting/init file loaded as it should.

But: at the starting exec, DIRECTORIES was set back to its initial MEDLEYDIR state.  It did not have the NIL (for my logindir) or the LFG directory that my init file explicitly establishes—and has been able to establish in earlier configurations.

So, something is happening after my personal greeting that causes at least DIRECTORIES to revert.  If I explicitly (GREET) again after starting up, the directories are correct.

This is starting a makesys, separate from the issue that nothing should happen when a sysout resumes (same user same machine).  In the makesys case we want the medley directories to be installed as part of the system greeting, but the effect of the user greeting (after the system greeting) should be preserved.

(Separately, I think the initial DIRECTORIES should always include the loginhost/dir (NIL)).

Issue #931, opened 2022-09-11
HARDCOPYIMAGEW.TOPRINTER in HARDCOPY breaks on list entries in DEFAULTPRINTINGHOST


D-Van-Buer:
With DEFAULTPRINTINGHOST  (FOOP (POSTSCRIPT BARP POSTSCRIPT)),
right click on a plain window (e.g. the Logo window), select Hardcopy > to a printer.  When printer menu pops up, it will offer FOOP and "BARP POSTSCRIPT", select "BARP POSTSCRIPT". This causes a break because the printer's name is BARP, not "BARP POSTSCRIPT".

The error is in file HARDCOPY, function HARDCOPYIMAGEW.TOPRINTER, which unlike HARDCOPYREGION.TOPRINTER does not correctly handle the menu for the entry (POSTSCRIPT BARP POSTSCRIPT).

The user documentation also does not explain the list entry on DEFAULTPRINTINGHOST, it's only in comments in HARDCOPY.
The format is (printertype name preferredimageformat).  printertype needs to be in global PRINTERTYPES such as POSTSCRIPT or INTERPRESS.

 - OS: Linux
 - OS Version: Fedora 35
 - Host arch: x86_64
 - Maiko version: [e.g. commit ID from `git log | head`]
 - IL:MAKESYSDATE: 12-Aug-2022 23:46:38

I've tested the obvious fix.
Time to get out my old Hawley mouse and USB adapter I made for it!

masinter:
it's been so long since I've had a printer or printed anything, I'm not sure what "the obvious fix" you had in mind. But I copied the initial COND from HARDCOPYREGION into HARDCOPYIMAGEW.TOPRINTER. Set DEFAULTPRINTINGHOWT TO ((INTERPRESS "POSEIDEN:" INTERPRESS)) 
and I was able to print.
[job_0183_EDF8_714E.pdf](https://github.com/Interlisp/medley/files/9816312/job_0183_EDF8_714E.pdf)
[job_0183_EDF8_714D.pdf](https://github.com/Interlisp/medley/files/9816313/job_0183_EDF8_714D.pdf)
[job_0183_EDF8_714C.pdf](https://github.com/Interlisp/medley/files/9816320/job_0183_EDF8_714C.pdf)

PR #999, opened 2022-10-19
DEFAULTPRINTINGHOST can have list members as per comments

masinter:
As per #931 the error is in file HARDCOPY, function HARDCOPYIMAGEW.TOPRINTER, which unlike HARDCOPYREGION.TOPRINTER does not correctly handle the menu for the entry (POSTSCRIPT BARP POSTSCRIPT).

The user documentation also does not explain the list entry on DEFAULTPRINTINGHOST, it's only in comments in HARDCOPY.
The format is (printertype name preferredimageformat). printertype needs to be in global PRINTERTYPES such as POSTSCRIPT or INTERPRESS.

I copied the initial COND from HARDCOPYREGION.TOPRINTER into HARDCOPYIMAGEW.TOPRINTER. Set DEFAULTPRINTINGHOWT TO ((INTERPRESS "POSEIDEN:" INTERPRESS))
and I was able to print an image window. There were problems with clipping and scaling the image, but that is a different issue (with dodo's IP interpreter, I imagine)


D-Van-Buer (approved):
This fixes the issue.

rmkaplan (approved):
LGTM

PR #1003, opened 2022-10-25
another pass at variable initialization after logout savevm sysout makesys

masinter:
This is complex... the theory is that MAKESYS starts out with no initialization -- you could start it up with a completely different setup. 

If you want to set DIRECTORIES and then make a MAKESYS, the way to do that is to change the variable MEDLEY-INIT-VARS which has the initial values.


masinter:
There was some ugly hackery around getting loadups in the tmp/ directory to load WHEREIS.HASH and EXPORTS.ALL from where they are expected in a release. It was too difficult to get that to work, so I added to the "loadup-all.sh" script a final copy-all.sh that puts the components where they "belong" (using hard links rather than copying).

masinter:
One more try:  MEDLEY-INIT-VARS was on AFTERSYSOUTFORMS and got called twice; now more careful and doesn't change anything if called a second time without an intervening BEFORE event.

rmkaplan (approved):
I tested this by doing a makesys full, starting it up, observing that my initialization was still intact.

Did a sysout, started it up, and directories will still there.

So, I think this is good, finally.

rmkaplan:
Well, still not quite right.  After running the load-up script, exports.all and the whereis-hash are on tmp/.  But tmp is not on DIRECTORIES.

When I push that on, I can load EXPORTS.ALL from other places.  But I still can't get whereis to work.

PR #1005, opened 2022-10-25
whereis fix

masinter:
- copy files already done by loadup-all.sh
- whereis init had problems


rmkaplan:
Still no whereis or exports.all.  whereis says "File not found #.(PATHNAME "NIL")" on everything.  If you are backing away from tmp/ in the medley-dir setup, I can put that in my init.  Not sure why whereis stopped working, since I explicitly point to the tmp/ file in my init

PR #1007, opened 2022-10-26
Another pass on initialization of directory names over VMEM saves

masinter:
Hard to debug or test -- lots of cases. There are 4 'save state' operations with differing balance of how much to save or reset.  SAVEVM is meant as a 'save state right now in case I crash'. 
LOGOUT implies 'save state for later'. SAVEVM should be returnable with the same files open to the closest match... LOGOUT perhaps not. LOGOUT(T) doesn't save anything.  SYSOUT is a way of saving state but possibly sharing it, moving to another machine. While MAKESYS is an operation you only do for wide distribution, and a MAKESYS shouldn't be personalized.

The problem is that there are changes to the file system -- version numbers change, for example, and people move directories around. Some of these changes might be compensated for by using pseudohosts, but even then the pseudohosts have state.    Testing on multiple machines the 'same' code (or the same machine but running WSL or Cygwin) also requires adapting.

The variable MEDLEY-INIT-VARS (and the function of the same name) are used to control variable values and substitution. MEDLEY-INIT-VARS has tuples (<variable-name> <expression> [RESET]) of 2 or 3 elements of variables that might need to be adapted over a memory save, with 11 possible events.
BEFORE/AFTER/AFTERDO for SAVEVM/LOGOUT/SYSOUT/MAKESYS (except no AFTERDOLOGOUT).

The problem this PR addresses is that the location of the file 'whereis.hash' which is built during the loadup process needed to be updated even when the directory didn't change, because the variable wasn't just a bare file name, it named an explicit version ("whereis.hash;3"). But version numbers change often during the GItHub workflow.


rmkaplan:
tmp/ still missing from DIRECTORIES, WHEREIS still not working

masinter:
@rmkaplan how are you testing?  tmp is no longer needed as part of DIRECTORIES -- it was a broken hack that got worse.
It instead makes a hard link from the tmp copy to the place in 'loadups' and 'library' where they are expected.  I suppose we could move export.all into loadups but that is part of the longer discussion about what goes where, so I didn't move it.

rmkaplan:
The loadup scripts I have been using for quite some time put all the build artifacts in my working-medley/tmp directory.  Sysouts and their dribbles, exports.all, whereis, fuller database.

Then with working-medley/tmp on directories, the exports.all, fuller.database, and whereis-hash were accessible.

There was a time when the scripts were separately copied into loadups/ but we stopped that quite some time ago (at that time we were confused also about when and where an actual release was done).  

The scripts that I think I just pulled down and used also still seem to build in tmp/.  So in my working configuration nothing ever looks at loadups/. And the only thing in loadups after running the scripts is the starter sysout that is always there.

My model is that loadups/ is only populated by downloading and unzipping an official “release”.  I have another directory where I keep the official unzipped releases, for comparison purposes.  But I don’t work there.  (And separately I have the git-medley  folder that’s synchronized with the repo and changes from branch to branch, which I only use for communication staging.

So what I am seeing now is that everything gets built where I expect it to be (working-medley/tmp). That contains a consistent set of build artifacts and that’s where FINDFILE etc. expects to find them (by DIRECTORIES).  Nothing is copied or linked into loadups/ or library/.  (And I don’t particularly want export.all, fuller.database, or whereis-hash to migrate over to library, because then they don’t clear out consistently when I clear the old versions of other artifacts.)

If this isn’t the proper model going forward, what do I need to do?

masinter:
I understand now the problem, my model was wrong. The solution is not to always do the copy and instead only do the copy when making a release -- different kinds of releases might copy files to different places, depending.

Closing this (which accidentally includes UNIXCOMM changes)

PR #1008, opened 2022-10-26
Put tmp/ back in DIRECTORIES for MEDLEYDIR users

masinter:
replacement for PR #1007 -- 

> I understand now the problem, my model was wrong. The solution is not to always do the copy and instead only do the copy when making a release -- different kinds of releases might copy files to different places, depending.
>


rmkaplan:
This fixes the tmp problem, it now finds exports.all, fulller.database, and whereis.

But:  my add-ons to DIRECTORIES are still wiped out when I start up the sysout.  I still have to explicitly GREET() again.

masinter:
rewinding init changes

PR #1009, opened 2022-10-27
greet update

masinter:
See issue #1010


masinter:
The problem as I've been thinking about it is over-constrained.
This PR is going in the wrong direction, though... will close it

PR #1012, opened 2022-10-30
Rmk73  new tedit

rmkaplan:
This is the first upload of  the major revision of Tedit that I have been working on. Clearly a draft with possibly some odd behaviors and possibly some breaks, but it has reached the stage where it would be useful for other people to poke at it. I believe that the core functionality of formatting/display/hardcopy/selection/editing and file/stream manipulation are working on top of the new data structures and new code abstractions. I hope I have eliminated some of the crashes that we have attributed to Tedit, and I have fixed number of bugs.

If you build a full sysout in this branch and start to use it, please report any breaks or crashes, and anything that seems strange.  I have not had it trash any edit-files, but you should operate with caution.

There is still work to do (e.g. for some reason the key actions in LFG are not being triggered), but it would now be helpful to get some alpha-user testing.


fghalasz:
New TEdit and Notecards:  NoteCards defines (and depends on) a function TEDIT.LIST.OF.OBJECTS that takes a TEXTOBJ and returns a list of the ImageObjects it contains.  TEDIT.LIST.OF.OBJECTS uses TEDIT.MAPPIECES (from TEDIT) to extract this list.  Due to changes in TEDIT.MAPPIECES, TEDIT.LIST.OF.OBJECTS no longer works.  Here are the issues:

1.  TEDIT.LIST.OF.OBJECTS uses the (PIECE POBJ) field to extract ImageObjects from the pieces in the piece table.  But the POBJ field has been removed from the PIECE datatype.    So either we need to restore the POBJ field or figure out another way to extract the ImageObjects (and their character locations) from the TEXTOBJ.

2.  TEDIT.MAPPIECES takes a FN argument and APPLY*'s this FN to every piece in the piece table.  In old TEdit this FN was called with 4 arguments (CH# PC PC# FNARG).  In new TEdit, it is called with 3 arguments (CH# PC FNARG).    TEDIT.LIST.OF.OBJECTS is based on the 4 argument signature.  It doesn't really use the "extra" argument but it seems like an unnecessary change to TEDIT.MAPPIECES that may break apps beyond NoteCards.

3.  In new TEdit TEDIT.MAPPIECES breaks with "NIL is not a NUMBER" when it reaches the end of the piece table.  This is due to the `as CH# from 1 by (PLEN PC)` clause in the `for` loop.    At the end of the piece table, PC takes the value NIL, which results in the break when `(PLEN PC)` returns NIL instead of a number.  

If interested, TEDIT.LIST.OF.OBJECTS can be found in the file .../notecards/system/NCUTILITIES.

rmkaplan:
Thanks for trying it out, you are the first one in the pool.

> 1. TEDIT.LIST.OF.OBJECTS uses the (PIECE POBJ) field to extract ImageObjects from the pieces in the piece table.  But the POBJ field has been removed from the PIECE datatype.    So either we need to restore the POBJ field or figure out another way to extract the ImageObjects (and their character locations) from the TEXTOBJ.

I removed the piece-specific fields  POBJ, PSTR and PFILE in favor of a generic PCONTENTS field whose meaning is determined by the value of PTYPE.  This unscrambled a lot of code, made it easy to add a few new types and lays the groundwork for more (e.g. UTF-8).  The equivalent of POBJ would now be either (IMAGEOBJP (PCONTENTS PC)) or (AND (EQ OBJECT.PTYPE (PTYPE PC))(PCONTENTS PC)).  For convenience, I can add an access function to the piece declaration that preserves this interface.

> 2. TEDIT.MAPPIECES takes a FN argument and APPLY*'s this FN to every piece in the piece table.  In old TEdit this FN was called with 4 arguments (CH# PC PC# FNARG).  In new TEdit, it is called with 3 arguments (CH# PC FNARG).    TEDIT.LIST.OF.OBJECTS is based on the 4 argument signature.  It doesn't really use the "extra" argument but it seems like an unnecessary change to TEDIT.MAPPIECES that may break apps beyond NoteCards.

PC#, here and in random other places, was a vestige of a very early implementation, when pieces were referenced from an array.  I think John discovered that maintaining the order in the course of editing was a disaster, and went to the current B-TREE arrangement.  He didn't really clean things up, piece numbers were still being passed around in frivolous parameters and then either ignored, commented out, or checked for some kind of meaningless consistency.  This loop could count out the pieces too ("as PCNO from 1), but I think it would be better just to completely get rid of this notion.  (Note also that piece sequence numbers are an overlay of the edit history on top of the mapping to the character-sequences of the text:  if you insert a character and then delete it, all the later piece sequencenumbers are different, but the text hasn't changed.)

I would prefer moving this interface forward, and fixing the callers (I think Notecards, and essentially the same object function in TMAX.)

> 3. In new TEdit TEDIT.MAPPIECES breaks with "NIL is not a NUMBER" when it reaches the end of the piece table.  This is due to the `as CH# from 1 by (PLEN PC)` clause in the `for` loop.    At the end of the piece table, PC takes the value NIL, which results in the break when `(PLEN PC)` returns NIL instead of a number.

My bug, I'll fix it.  When I simplified this loop, I got confused about the order of evaluation of the "as" clause and the "while".  But I think I will also add an "inpieces" iterative statement for mapping through a piece sequence.  The ca;; tp TEDIT.MAPPIECES in the Notecards function could then be replaced by (for PC inpieces ... as CH# from 1 when  (POBJ PC) collect ...).  You wouldn't have top deal with your own TCONCing, and the funny interface for passing extra arguments.

Note also:  the atom LASTPIECE no longer exists.  When it runs off the end, you get to NIL.  (Interally there is an actual last piece that holds some other boundary-condition information.  But the NEXTPIECE field of the actual final piece is NIL.)
> 
> If interested, TEDIT.LIST.OF.OBJECTS can be found in the file .../notecards/system/NCUTILITIES.

In sum:  I would prefer not to maintain the PC# fiction just in this mapping function, although I could also just manufacture the numbers on the fly, with another counter. I can add a POBJ access function for backward compatibility. And I will fix the loop logic here, although I think the interative statement is a cleaner and more flexible solution.

What else is broken?

fghalasz:
Another New TEdit bug:   TEDIT the attached file (which contains just a single bitmap object).  It should come up OK.  Reshape the TEdit window so that it's width is less than the width of the bitmap in the file.,  Should get a break with "Help! UNRECOGNIZED CHARW"  (or maybe just an error msg - depending on how BREAKFLAG is set I guess).

Works OK in the old TEdit.

Spend some time digging around to isolate this bug from NoteCards and discovered that the break is caused in \DISPLAYLINE because the NEXTAVAILABLECHARSLOT field of the THISLINE field/object in the TEXTOBJ is set incorrectly causing the iteration through the CHARSLOTS to run off the end of the "good" data (to where CHAR and CHARW are both NIL).  Didn't track any further how it got into this state.

Note: Attached file is zipped because Github won't accept a TEdit file as an attachment.

[BM-CARD-BUG.zip](https://github.com/Interlisp/medley/files/9946846/BM-CARD-BUG.zip)

fghalasz:
Another bug discovered in TEdit while testing Notecards.  But this one is in both Old and New TEdit - although it appears to have been introduced in "modern times".

Bug is in TEDIT.DEACTIVATE.WINDOW.    In some cases (at least in Notecards) this function is called on a window whose TEXTOBJ has already been NIL'd out.   This results in an error/break "ARG NOT TEXTOBJ  NIL".

Fix is to add `(AND TEXTOBJ (replace ...))` around the `(replace (TEXTOBJ EDITFINISHEDFLG) of TEXTOBJ with T)` at the beginning of the function.

Figured I'd just let you add this fix to the new TEdit rather than do a whole new pull request just for this one small fix.

rmkaplan:
I think I already saw this one, in the context of bugs (old and new) in window-unsplitting (which was generally broken in the old).  I hope this is fixed for this case in the files I just put out.

rmkaplan:
The line overflow logic is complicated.  In this case, the old code (\FORMATLINE) apparently let’s it run over the margin, and let’s the window clip it. For characters it sends run-overs to the next line.

But this reveals another glitch in both old and new:  If instead of reducing the width, if you raise the bottom above the bottom of the bitmap, they both don’t just clip the bottom, the bitmap completely disappears.  What’s worse, if I then lower the bottom, it can’t make it come back, even if I scroll or redisplay

Another nonfeature in both old and new:  if you don’t reduce the window but simply click in the window arbitrarily far to the right of the bitmap image, it brings up the bitmap menu, as if it thinks that the mouse is still in the bitmap.  Even if the pop-up is confusing, at least with both you can type new characters at the end of the bitmap (although the new code doesn’t blink the caret there to tell you that you can).

The old code (TEDIT.SCAN.LINE) had complicated heuristics to determine what happens if you click past the visible end of the line (as determined by \FORMATLINE). I rewrote it to disentangle the cases but tried to replicate the behavior, apparently a mistake in this case.  I think the object-menu action is isolated to a particular subfunction, so it should be (relatively) easy to fix.  (There is a separate geographical problem in the Tedit menus, old and new, where you can click to a modifier on the right of a fill-in field (like “picas” in the paragraph-looks menu), and counterintuitively the caret goes to the following unrelated field.)

rmkaplan:
Fixed, hopefully, in commit just pushed.

nbriggs:
I just made a new full.sysout with this, loaded Lafite, and tried to compose a message: if you start typing over the blue-pending-delete Subject:, when you type the first character it is displayed with the caret blinking to the right of a blank space at the end of the line.  It continues to be to the right of the characters typed as you type additional characters.  Attached: before and after typing one character.

Screen Shot 2022-11-08 at 10 18 32 AM
Screen Shot 2022-11-08 at 10 22 35 AM

nbriggs:
If you start up an empty TEDIT window `(TEDIT)` and then type any character(s) followed by <return> it does not place the blinking caret on the next line but at the right of an extra space on the current line.  If you then type another character it places it correctly at the beginning of the next line with the caret at the right edge of the character, and continues to display correctly as you type additional characters (or you type <return> at which point the mis-display repeats).

nbriggs:
Also note that the blue-pending-delete highlighting on the Subject: field extends one character beyond the `<<`.

rmkaplan:
Actually, there isn’t a real space in the piece table, I think it is putting the caret to the right of how it is displaying the EOL.  Should be to the left of that.

The empty-window example is probably the same issue, one part of the code is trying to display the EOL when you click to the right, so you can delete it, but this is a different situation.

Any maybe event the blue-pending delete extension over to the EOL.

Thanks

PR #1016, opened 2022-11-02
apply ethernet related \MAIKO.MOVDs and rename \MAIKO fns and vars

masinter:
step 1 of medley ethernet cleanup


nbriggs:
Why take out the setting of time from the network?  If the Dodo code provides the correct begin/end DST settings in a time packet it means we can stop worrying about those values being correctly set up in the site-init file.

masinter:
The \MAIKO.MOVD version of settime doesn't use the network at all -- it gets the time zone from a SUBRCALL.

nbriggs:
If it's getting time from the SUBR then it's getting just *time* and not the begin/end DST and time zone name.  The network version when you have NS is more useful.

masinter:
Going through the changes.. mostly taking functions defined originally for D-machine networking and redefined on MAIKOETHER and (a) deleting the D-machine definitions and (b) renaming from \MAIKO.XXX to \XXX when necessary. Specifically:
10MBDRIVER: \10MB functions redefined on MAIKOETHER
AINTERRUPT one function \MAIKO.ETHER-INTERRUPT renamed
LLETHER removed \CHECKSUM later defined by MAIKOETHER
LLNS:   \NS.SETTIME definition for D-machines was overridden by a SUBRCALL definition on MAIKOETHER. The old (but non-functional) definition is now retained in a (IF NIL THEN ....) because it might be useful to use it? 
MAIKOETHER: 
  \\10MB.STARTDRIVER:  additional chanes unclear -- looks like I started with a later version
PUP remove non-MAIKO option

masinter:
raised an issue to redo this more carefully,closing this PR

PR #1017, opened 2022-11-06
Rmk74  run with pseudohosts

rmkaplan:
A few refinements to complete the integration of pseudohosts.  I have been running for a week or so with all of the paths on directories converted to pseudohosts (just {WMEDLEY}/lispusers/) instead of /User/kaplan/local/medley3.5/working-medley/lispusers/). (SETQ DIRECTORIES (MAPCAR DIRECTORIES (FUNCTION PSEUDOFILENAME))).

You can run in pseudohost mode if you cd to the pseudohost directory constructed from MEDLEYDIR instead of to the directory's TRULEFILENAME.  Then the pseudohost names also will appear in FILECREATED expressions.  (e.g. {WMEDLEY}/lispusers/GITFNS if you make the file in a working-medley.  Or {GMEDLEY}... otherwise.


masinter (approved):
I want to review the Pseudohost interface but no point in holding back this PR

PR #1020, opened 2022-11-14
Use -title rather than -t to specify window title

nbriggs:
When when lde is compiled for SDL, without the X11 parsing being engaged the "-t" option is used for setting a timer and the title can only be given with "-title".  Always specify it using "-title" in run-medley script.

### DIFF
--- a/run-medley
+++ b/run-medley
@@ -197,5 +197,5 @@ echo "greet: $LDEINIT"
 
 export INMEDLEY=1
 
-"$prog" $noscroll $geometry $screensize $mem -t "Medley Interlisp" $pass "$LDESRCESYSOUT"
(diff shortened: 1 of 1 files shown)


masinter (approved):
change the default in Maiko so this isn't necessary?
but this PR is fine

rmkaplan:
There was a time, a week or so ago, when I could put -title as a parameter to run-medley, and it would put that string in the window title.  Very handy, since I may have 3 or 4 Medleys going at once as I compare how Tedit behaves.

But that seems to have reverted, the Medley Interlisp default overrides whatever the caller might have specified.

nbriggs:
I'll fix that.

PR #1023, opened 2022-11-17
The macroexpansion of UNDOABLY shouldn't depend on runtime rebinding of LISPXHIST

masinter:


rmkaplan (approved):
Is that test to guard against behavior in the primordial state when before the very first event?  I presume that if its empty, no harm no foul

PR #1026, opened 2022-11-20
Allow user override of -title option

nbriggs:
The switch from `-t` to `-title` to specify the default title had the unintended side-effect of preventing a user override of `-t` by `-title`.  Only set the default title (Medley Interlisp) if there is none specified on the command line.


rmkaplan (approved):
Works for me

PR #1027, opened 2022-11-22
Fix to Issue#1022 "Error during greet"

fghalasz:
Problem with greet is need to reset MEDLEYDIR before greet process otherwise greet uses the "old" MEDLEYDIR set during the loadup MAKESYS.

Lots of places could put this rest, but I put it directly in MEDLEYDIR-INIT so as to hopefully avoid any issues with other uses of MEDLEY-INIT-VARS.


fghalasz:
Turns out we need to reset MEDLEYDIR for AFTERMAKESYS as well.  Made this change in the MEDLEY-INIT-VARS function. 

Discovered this trying to do a loadup for online with no GREET file.  In this case, MEDLEY-INIT-VARS is called by AFTERMAKESYS.

Probably this case eliminates the need to reset MEDLEYDIR  in the MEDLEYDIR-INIT file.  But leaving it in there just in case since there is no damage (I think) to doing it twice.  And its just to hard to figure out who all is calling MEDLEY-INIT-VARS from where and when.

rmkaplan:
It starts up after makesys now and does my user init, but when the greet completes DIRECTORIES and friends have been reset back to what the system greet file wanted, not what my personal init specified.

rmkaplan:
Maybe the problem is that MEDLEY-INIT-VARS is on AROUNDEXITFNS.  I think this means that it gets executed no matter what the event is (BEFORE/AFTER LOGOUT/MAKESYS/SYSOUT…)

masinter:
There are three events for each kind of SAVEVM/LOGOUT/MAKESYS/SYSOUT:
BEFORE, AFTER, and AFTERDO.   After is run when you return restarting.
There is no AFTERDOLOGOUT since LOGOUT doesn't continue.

rmkaplan:
MEDLEY-INIT-VARS runs no matter what the event happens to be.

nbriggs:
What is it that we expect to have happen, and when exactly does it need to be happening?
Do we need the various directories set up *before* the GREET?  We did that in the site init file at PARC.
The code in APUTDQ for GREETFILENAME looks for the site init in either the UNIX environment variable LDEINIT, or
"{DSK}/usr/local/lde/site-init.lisp"

Below is how greet gets things done (from HIST).

```
        (COMS (FNS GREET GREET0)
              (ADDVARS (PREGREETFORMS (DREMOVE GREETFORM RESETFORMS)
                              (SETQ CONSOLETIME (SETQ CPUTIME (SETQ EDITIME 0)))
                              (SETQ CONSOLETIME0 (CLOCK 0))
                              (SETQ CPUTIME0 (CLOCK 2)))
                     (POSTGREETFORMS (SETINITIALS)
                            (AND EDITCHARACTERS (APPLY 'SETTERMCHARS EDITCHARACTERS))))
              (DECLARE\: DONTEVAL@LOAD DOCOPY (VARS (GREETHIST)
                                                    (SYSTEMTYPE)
                                                    (GREETFORM '(LISPXEVAL '(GREET)
                                                                       '_))
                                                    (CUTEFLG)
                                                    (GREETDATES '((" 1-JAN" . "Happy new year")
[...]
                                                                  ("25-DEC" . "Merry Christmas")))
                                                    (USERNAME)
                                                    (HOSTNAME)
                                                    (CONSOLETIME 0)
                                                    (CONSOLETIME0 0)
                                                    (CPUTIME 0)
                                                    (CPUTIME0 0)
                                                    (EDITIME 0)
                                                    (FIRSTNAME))
                     (ADDVARS (BEFOREMAKESYSFORMS (SETQ RESETFORMS (CONS GREETFORM RESETFORMS))
                                     (SETQ MAKESYSDATE (DATE))))
                     (ADDVARS (AFTERMAKESYSFORMS (LISPXEVAL '(GREET)
                                                        '_)))))
```

rmkaplan:
I think the problem is that MEDLEY-INIT-VARS is being run AFTER the user’s greet file, as well as before.

fghalasz:
So what is happening is that in the case of AFTERSAVEVM, first the greet file (MEDLEYDIR-INIT) is run (which in turn calls MEDLEY-INIT-VARS) and then the AFTERSAVEVM events are fired - which calls MEDLEY-INIT-VARS again.

So the right thing to do is to get rid of the AFTERSAVEVM call to MEDLEY-INIT-VARS and just depend on the greet file to do its work.

The one complication is if a  "savevm" is run with -nogreet or in a loadup situation where the greet file is suppressed.  I think for these cases, we will just have to assume that the LOADUP.CM makes a call to MEDLEY-INIT-VARS (if necessary) or that the nogreet user has some other way of setting up their environment.

I will update this pull request to remove the AFTERSAVEVM call to MEDLEY-INIT-VARS (and change the LOADUP.CM for online to call MEDLEY-INIT-VARS directly).

masinter:
I was pursuing the idea that calls to MEDLEY-INIT-VARS(*AFTER) were safe to repeat.

fghalasz:
@masinter   I don't think the issue is repeating AFTER* calls to MEDLEY-INIT-VARS.  Rather it's the issue of calling  MEDLEY-INIT-VARS in an AFTER* context after already having called it as part of GREET - which may modify the various variables in a non -standard way.

fghalasz:
OK. Latest iteration of this PR now does NOT do take any action in MEDLEY-INIT-VARS when event is AFTERMAKESYS.  After a MAKESYS, MEDLEY-INIT-VARS will be called only once - from the MEDLEYDIR-INIT greet file.

All of the BEFORE and AFER actions in MEDLEY-INIT-VARS were left as is.

Cleaned up the greet call to MEDELY-INIT-VARS a bit as well.

masinter (approved):
I'm not sure about (MEDLEY-INIT-VARS 'GREET) being necessary but it shouldn't hurt. @rmkaplan does this work for you?

rmkaplan:
I tried a few scenarios, does seem to work.

PR #1033, opened 2022-12-17
Revert "Lmm cleanup new shell device"

masinter:
Reverts Interlisp/medley#1006 will redo


masinter:
this should remove the bogus changes and let me redo the necessary parts.

nbriggs:
Yes, this gets you a working [for (CHAT 'SHELL)] loadup.

nbriggs (approved):
LGTM.

rmkaplan:
What are the important test cases for this?  I have a loadup based on the later version that I have, which includes a round of Larry’s cleanups, and a little bit more.

CHAT(SHELL) seems to work, although I see some odd characters on the prompt lines, as in the screen shot below.  But that is also what I see in Master.

Anything else to test before I do a PR?  (And, does it matter whether or not the UNIX-PTY FDEV has FDBINABLE T or NIL ?)

masinter:
@rmkaplan I wanted to get to a stable state -- I decided to roll back rather than forward because I couldn't find the version you have, and I wanted to make sure we had a working master before fidding with the new. I think I should make more of a priority of getting the test repo together enough to put test cases in.

I think fixing up things like MAIKO.MOVVDs and SHELL FDEV is a rabbit hole

PR #1034, opened 2022-12-19
UNIXCOMM:  Eliminated the new shell device in favor of a single shell…

rmkaplan:
… device

Also removed unused functions labeled as "obsolete" after Medley 2

Device is also now not FDBINABLE


masinter (approved):
thanks

PR #1035, opened 2022-12-24
loadup-db.sh fix -- don't rely on loadups

masinter:


rmkaplan (approved):
This just changes the parameter -full to -n, which run-medley appears to map to the full sysout.  So a cleanup, with no change to behavior?

nbriggs:
@rmkaplan `-n` uses `tmp/full.sysout` instead of `-full` giving you `loadups/full.sysout`

masinter:
this fixes a problem that only occurred when I ran loadup-db in a clean clone.

PR #1048, opened 2023-01-17
masterscope extensions doc (internal) + tweak helpsys

masinter:


rmkaplan (approved):
It seems that this just moves a documentation file into internal (a little obscure but...) and renames REPO.TYPES.  LGTM

masinter:
This fixed the error trying to add MSX to WHEREIS.
It also sets up so if you say `man DECL` it will show you DECL.TEDIT even though there isn't a DECL in wheereis.

PR #1051, opened 2023-01-21
Create UNIXUTILS in library to go along with UNIXPRINTER, et al

fghalasz:
This PR creates a UNIXUTILS library entrant to go along with the other UNIX... files in library.  The following functions are defined in the UNIXUTILS file:
`(ShellWhich CMD)`  - performs a Linux 'which' on CMD to show where the executable of CMD is stored in the underlying file system.
`(ShellBrowser)` - based on Larry's CLHS.OPENER with some enhancements.  Determines which external browser is available to open html files / websites. 
`(ShellBrowse URL)` - opens URL (which must start with http:// or https://) is tghe external browser determined by ShellBrowser.

This PR also moves the `ShellCommand` function from UNIXPRINT to UNIXUTILS - since it doesn't really belong in UNIXPRINT and makes moere sense in UNIXUTILS.

This PR also updates LOADUP-FULL to include the new UNIXUTILS file in the full.sysout.

This code is used in the upcoming apps.sysout (separate PR coming soon) that mimics the fully loaded online.sysout used in online.   I will also update online to use these functions instead of similar functions used there.  And might want to update CLHS functions in HELPSYS to use these common functions as well..


masinter (approved):
I'm uncertain where we're going with this mode of printing.  This isn't how people print if they print at all. Some way to direct printing to produce a pdf?

Issue #1064, opened 2023-02-02
Better defaulting for GITFNS projects and pseudohosts


rmkaplan:
This is what I recall from discussions a few weeks ago:

The GITFNS projects for the default clones (MEDLEY, LOOPS, etc.) should have pseudohosts {MEDLEY} {LOOOPS} etc. and not {GMEDLEY}..., whether or not someone is running (like I do) with a separate working environment.

And the search  probes for the clone (based on MEDLEYDIR, NOTECARDSDIR...) should be clarified.

Separately, if there does happen to be a working directory as a sister to the clone directory (working-medley, working-loops etc.), then it should silently build the {WMEDLEY} etc. pseudohosts.  But not cause an error if a working directory can't be found.

It should do those probes if the CLONEPATH and/or WORKINGPATH directories are NIL (or T for backward compatibility).

Thus, without a working directory for Loops (GIT-PROJECT 'LOOPS) should produce {LOOPS} (if it can be found offset from the various unix-var possibilities).

PR #1065, opened 2023-02-02
GITFNS:  Better default directory-finding, better pseudohosts  (#1064)

rmkaplan:
Also updated documentation


masinter (approved):
been pending too long. There are likely some configurations this won't work for, but the number of people working on medley using github is limited

PR #1069, opened 2023-02-03
Update PICK to choose demo topics

masinter:


masinter:
needs more planning than PICKing topics

PR #1074, opened 2023-02-09
add :EDIT-BY initials to FILECREATED

masinter:


masinter:
We should discuss if this is an improvement or worse.

rmkaplan (commented):
This will be useful particularly if people start working in the Medley/Loops etc. pseudohosts that no longer have user-identifying local-machine paths.  But sometimes even the initials are obscure, so I wonder whether it should actually put out the actual name of the user (USERNAME) instead of just initials, or perhaps the first name too (value of FIRSTNAME).  E.g. :EDIT-BY "Ron Kaplan"  (CONCAT FIRSTNAME " " (USERNAME))

rmkaplan (approved):
LGTM

nbriggs:
Works for me.  Not sure why it's :EDIT-BY instead of :EDITED-BY, but it doesn't matter.

PR #1077, opened 2023-02-17
Windows installer and medley script for running Medley in Docker on Windows

fghalasz:
This PR includes:

1. Single step installer for Medley running in Docker on Windows.  Installer includes everything needed to run Medley in Docker except for Docker Desktop.

2.  Single step `medley` Powershell script for running the Medley docker container, starting the vncviewer, and "connecting" the two.  Medley.ps1 is a companion to and eventually calls the medley.sh script used in Linux and WSL.

3.  New Medley docker container built from the linux .deb files rather than using a redundant Dockerfile build process.

4.  Updated medley.sh script on Linux to accommodate changes needed by Windows medley.ps1 script.

5.  Updated man page.

6.  Linux medley script now works on WSL1 as well as WSL2.

A companion PR is over in the interlisp.github.io repo to update the Running On Windows page on interlisp.org.


installers/win/editpath/EditPath.iss
```diff
@@ -0,0 +1,165 @@
+; Copyright (C) 2021-2023 by Bill Stewart (bstewart at iname.com)
```
masinter, line 1: do we need to worry about this?

masinter (approved):
not sure how to review this without merging, so...

installers/win/editpath/EditPath.iss
fghalasz: No.  The EditPath project is LGPL licensed.  That particular file is just an example of how to use EditPath in a setup script.  Not needed at all by Medley, so I will delete it.

But we should probably add an open-source software attribution "page" to my medley scripts and to online.  I will put that on my list of things to do.

PR #1082, opened 2023-02-23
WIP git diff using difftool

masinter:


masinter:
this is a Work In Progress, incomplete and probably buggy, and requires a recent addition to git.

On the other hand it is harmless, doesn't redefine anything and might give an alternative when GITFNS fails.

masinter:
No interest in reviewing. The mechanism for calling back to Interlisp is an awful hack. Closing this for now.

PR #1084, opened 2023-02-24
SPY ignores samples from background process; increase yield time when in backgroound

masinter:
Background yield had the bad effect that SPY stack samples were more likely to hit when in the BACKGROUND-YIELD function.
This commit fixes the interaction by introducing \BACKGROUND which is NIL top level but bound to T in the background process, and turning off (ignoring) SPY samples.

Included are two minor changes: BACKGROUND-YIELD increases yield time when \IDLE (for online users) and eliminate code in SPY for two-byte atoms (simplifies code for the actual case now).


nbriggs:
Hmm... perhaps the maiko code should be noticing that it was woken from the yield subr (os sleep()) by a timer interrupt and not generating a Lisp level interrupt that spy would notice?  I think it's not that helpful to ignore samples that occur in functions other than the yield subr if they're executing in the background process.

masinter:
I convinced myself that in most cases, the background process actions have nothing to do with what you want to spy on, since it's just a global variable of functions to call when there's nothing better to do.

I think fixing the YIELD subr to not inject extra non-periodic interrupts. The SPY trees I get when I run with this installed are quite similar to what I get when I (DREMOVE 'BACKGROUND-YIELD BACKGROUNDFNS) turning off the yield feature.

masinter:
redone as PR #1293

PR #1086, opened 2023-02-28
Revert changes in \DRAWLINE.DISPLAY and add BIGBITMAPS to loadup & exports

masinter:


rmkaplan (approved):
This seems to revert to a previous good state, given that BIGBITMAPS are now included in the loadup.

PR #1100, opened 2023-03-08
add rooms manual (scanned, no tedit found)

Anzus:
add rooms manual The lafite manual (.tedit sources) added by another commit 

Derived files (PDFs made from sources, or scanned images) go in the documentation folder of the Interlisp shared Google drive. Only upload scanned PDFs if the original source is lost and we should reconstruct.


masinter:
I added a category to the Interlisp Zotero called Interlisp Documentation.
The idea is to collect together documents, with appropriate metadata (title, author, date, etc.)
An we have a google drive with PDFs. I'm reluctant to put scanned PDFs in the mrdley repo because they are so large. The history  repo? There aren't TEDIT files for these in the envos repo?

masinter (approved):
we should also put these in zotero

masinter (changes requested):
3/27/23 "do we really want to put the PDFs in with the sources"?
Where else? Zoterro is currently not a good answer I don't think. Static pages of website?

masinter:
Interlisp/history/1990s/1993-envos/OTHER/lafite/Doc/Manual/LAFITEMANUAL-* 
seems to have "A USER'S GUIDE TO LAFITE" in TEdit format.
I got this by scanning All-Envos-PDFs.pdf 
for "GUIDE TO LAFITE" pp2783-2991 

I found a lot of material about ROOMS but not the manual.

rmkaplan (approved):
Perhaps we'll find a better home later, but let's get this into the repo

masinter (approved):
will need to clean up later, look for ROOMS in PARC gackups

PR #1103, opened 2023-03-12
Tedit updates--next round

rmkaplan:
Rework of selections and display, merge of display and hard-copy line formatting,


rmkaplan:
This is the next round draft, with lots of changes from the one from a few months ago, lots of code thrown out and reduced.  Still needs lots of shake-down testing--glitches that you observe if just playing around would be helpful.

nbriggs:
I made a new full.sysout, started it, used the background menu to open and place a new TEdit window, and without clicking in that window typed a single character -- at which point it took down the cursor went into an infinite loop involving TEDIT.\INSERT, TEDIT.INSERT.UPDATESCREEN, SCROLLW, \TEDIT.SCROLLFN

rmkaplan:
Thanks, I think I have also been working on that this morning.  It happens when you insert after the end of the file.  Next update.

rmkaplan:
This is a next pretty much complete revision.  It may still be unstable, but it is worth loading and testing.  There may even be some obvious failures, despite all the testing I have done.  I have tried to centralize as many pathways as possible, thrown out lets of code, but still there may be interactions I haven't encountered.

If you try it, look for odd things in the display: do the lines look right and scroll, is the selection and highlighting what you expect, the caret in the right place, etc.

Does history work, at least the way it used to:  undo (but maybe not redo at this point)

Subwindows may or may not work right--next on the list.

Etc.

rmkaplan:
Actually, subwindows do appear to work--but would be useful to know if there is any bad behavior.  Implementation is still needs cleaning and restructuring.

nbriggs (commented):
You probably expect this, but: Middle-button "Expanded Menu" produces a non-functional menu with odd layout, and the same is true for all the other TEdit-based menus.

rmkaplan:
Actually, I didn’t expect this.  Those worked fine in my last tests.  But then I must have improved something else...

rmkaplan:
Comment from the last commit:

I fixed the menu thing--another piece of left-over debugging stuff.  With respect to hardcopy crashes:

OPENSTRINGSTREAM always seemed to be in the neightborhood of the hard hardcopy crash, since it was being used to insert page numbers.  I changed that, and also changed the TEDITSTRING entry point to eliminate the dependency on OPENSTRINGSTREAM.  Instead, it opens an empty text stream, and then does TEDIT.INSERT of the desired string.

But note:  I neve clearly identified the problem with hardcopy, it seems to have gone away in my tests, but still tread lightly.

nbriggs (commented):
It's looking really good now.

I took a look at the Postscript output from hardcopying the same file under old and new TEdit, and while they *look* indistinguishable at first glance (same line breaks, etc.) there are some subtle differences in the Postscript that is generated, and I don't know why.  It seems to be mostly Y-positioning differences - for example in library/MAIKOCOLOR.TEDIT:
```
< 25800 36482 M (COLOR) S
< 25800 35482 M (LLCOLOR) S
< 55801 30099 5401 30099 200 L
---
> 25800 37470 M (COLOR) S
> 25800 36470 M (LLCOLOR) S
> 55800 31087 5400 31087 200 L
```
(it's doing `x y M[oveto]`, `(string) S[how]`, and drawing a line (L))

rmkaplan:
Earlier I noticed one case of a slight white-space difference between old and new in a paragraph heading, by printing and holding up to the light.  

But I couldn’t determine the source of the difference, and it wasn’t clear which one was correct.  But the margins, page-number position, headers, tabs, and other things seem to match.  (Line breaks in hardcopy display mode also match up.)

It would be helpful to get a noticeable difference where it is obvious which one is correct.

Now I’m turning to another problem: dealing with windows that are not entirely on the screen.  Some time ago Larry suggested a hack in the scroll function (which Sedit could also use) of simply repainting the whole thing whenever any part of  the window is off the screen.  That seemed to catch the obvious cases in the old code, in scrolling and with deletions.  But my updates try take advantage of the screen bits in more situations, so I have to pay more attention to the region boundaries.

nbriggs:
BTW - I used the new TEdit to open an old file from the CHM release of PARC file server archives.  Matt's ADD-SEDIT-COMMANDS.TEDIT (probably from the Lyric release) - and it seemed to work just fine.

MattHeffron:
@nbriggs 
~~Two~~ Three things:
1. I don't seem to have a copy of that file! (I have the source code, but not a TEDIT file.
2. I don't know _why_ that file would **_ever_** have been on a PARC server. That was part of the infrastructure of Beckman Instruments' Expert System development environment, which **shouldn't have ever escaped from Beckman!** (There's nothing particularly proprietary about _that_, but I'm wondering how it got to PARC.)
3. Are there any other files that you have access to, for which I am an author, or with Beckman copyright, that aren't already in one of the repositories?

nbriggs:
@MattHeffron you can sign up for access to the PARC archive here:  https://info.computerhistory.org/xerox-parc-archive

I was looking at https://xeroxparcarchive.computerhistory.org/phylum/lisp/lyric/lispusers/.index.html -- where I'm recorded as the last person who touched the files in Dec 1987.

I don't think the CHM has indexed the content of the files so finding you or Beckman would involve automating a download of all the files and grepping through them.

masinter:
i downloaded the whole archive to a portable drive, but i could move it (or the Lisp parts) to something more acessible ... but where?

PR #1111, opened 2023-03-19
fix Y2K problem in greet choices for good morning/evening/you're working late

masinter:


sources/HIST
```diff
@@ -1,12 +1,15 @@
...
PACKAGE "INTERLISP" 
READTABLE "XCL" 
BASE 10)
-(FILECREATED "10-Jul-91 12:07:43" |{PELE:MV:ENVOS}<LISPCORE>SOURCES>HIST.;3| 152184 
 
-      |changes| |to:|  (VARS HISTCOMS)
+(FILECREATED "19-Mar-2023 10:09:08" |{DSK}<home>larry>il>medley>sources>HIST.;2| 152115 
 
-      |previous| |date:| "16-May-90 18:10:04" |{PELE:MV:ENVOS}<LISPCORE>SOURCES>HIST.;2|)
+      :EDIT-BY "0"
```
nbriggs, line 5: Eh?

masinter: fixed EDIT-BY

rmkaplan (changes requested):
I think there is a more general approach to this and other remaining occurrences of the Y2K problem, instead of hacking string positions in each case.  When I first started on this round, in 2018, I added a heuristic to IDATE to recognize 2-digit years.  So (GDATE(IDATE XXX) guarantees the post-Y2K string format for XXX.

masinter:
@rmkaplan You're requesting a change but I'm not sure what change you're requesting.
If there's a better implementation that works...  feel free to update the PR.
I was leaving the calculation (which is problematic for online users -- what time zone are they in?) for nostalgia value.

rmkaplan:
I’m suggesting that you use the general functions for fixing this up, a little less second guessing of character-strings.

For example,  (GDATE (IDATE xxx) (DATEFORMAT NO.TIME)) would give you  "19-Apr 1993” no matter the input format of XXX.

Even better for this purpose would be to add an extra option NO.YEAR to \OUTDATE in IOCHAR, and then
   (GDATE (IDATE xxx)(DATEFORMAT NO.TIME NO.YEAR)) would give you just “19-Apr”.

masinter:
What it was looking for was just the hour. Is there a (DATEFORMAT HOUR.ONLY)? 

Counting characters from the end rather than the beginning seemed like a way to be compatible with both 2-digit and 4-digit results. I think there's a common lisp function that will decode the date in universal time, but not in "current time zone time".

In retrospect, I would have left the default to generate 2-digit dates with year < 70 meaning 20xx and year >= 70 to mean 19xx (the Symbolics hack), and added another parameter to ask for 4-digit years.

That would have some different compatibility problems but fewer I believe. Too late now.

rmkaplan:
I thought you were also worrying about April Fools etc not just working late.  DATEFORMAT (as per \OUTDATE) has a NO.DATE key that gives you time only, and you can add NO.SECONDS to get only hours and minutes.   I’m not sure the DATEFORMAT keys are documented.

masinter:
i rewrote it to use common lisp get-decoded-date which provides the date fields as numbers.
I didn't change the format of GREETDATES.

masinter:
redone branch greet-y2k-rewrite

PR #1123, opened 2023-03-30
GITNFS:  Doesn't error if clonepath defaults but is not required

rmkaplan:
If CLONEPATH is NIL, no error if the clone can't be found, just prints a note.  If CLONEPATH is T, will error.


masinter (approved):
is git-put-project-field exteraal?

rmkaplan:
WRT git-put-project-field, I wanted to have a way of setting a project property without exposing all the record fields.  I was aiming for the ability to temporarily change the exclusions from what gets inherited from .git or is otherwise defined in general (e.g. don't pay attention to sysouts).  But then I realized that there s a prior problem that COMPARESOURCES needs more work.  It doesn't filter by all the desired exclusion-matching criteria, even if specified.  It also doesn't sort the browser display in a useful way (puts  all TEDIT-XXX between TEDIT and TEDIT.LCOM).  So this got added to my list.

PR #1150, opened 2023-04-10
need a better heuristic, but 1 millisecond is better for newcomers. Never write copyright info

masinter:


masinter:
replaced by #1154

Issue #1052, opened 2023-01-21
Who line Rdtbl doesn't update when *READTABLE* is set, either externally or in WHO-LINE drop down menu


masinter:

nbriggs:
The names of the readtables that WHO-LINE presents are the names of the readtables defined in `\READTABLEHASH`.   Likewise, the packages are collected from `(CL:LIST-ALL-PACKAGES)`. In both cases, alphabetized.  I wouldn't object to filtering out some from the menus, as long as it's a variable I can set to `NIL` to turn it off -- the only times I want to use the package and readtable setting features are when I'm doing something that is likely using exactly the weird packages, and looking at the menu for a complete list is really handy vs. remembering that it's a maphash over the readtable hashtable or cl:list-all-packages...

masinter:
I left out the main part of the complaint (except in the Subject Line): If I select RdTbl -> XCL it actually switches *READTABLE* to be XCL but it doesn't update the who-line entry, which still reads INTERLISP. Similarly, if I switch (SETQ *READTABLE* (FINDREADTABLE "XCL")) it doesn't update.

nbriggs:
It updates as expected for me in both the menu and the `SETQ` case.  I wonder what who-line controls you've got set differently from the default -- try it in a clean full.sysout without your init file?

PR #1153, opened 2023-04-13
tweak WHO-LINE

masinter:
tweaks to hostname (don't use PUP), cndir (avoid directories that don't exist), package (sort so likely choices are at top)

Raised in #1052


fghalasz (approved):
LGTM.  Tested and works as expected.

PR #1158, opened 2023-04-16
Add BROWSERMAX to limit depth of paths graph

masinter:


masinter:
I was going to combine this with another "simple" fix to make 'show paths from a to b' do something useful, but that's a much harder problem... save for another day....

MattHeffron (approved):

rmkaplan (approved):
Seems reasonable to limit the dept--the deeper it goes the less interpretable it gets to be--if that eliminates this particular route to failure.  But we still have pesky uraid's to deal with as a separate matter.

PR #1159, opened 2023-04-17
Fix \MAPMDS, compile STORAGE lispusers, fixes HARDRESET problem too

masinter:


nbriggs:
Can you say more about "fixes HARDRESET problem too"?  Is there an issue number for that?

masinter:
There isn't an issue on the HARDRESET problem, it's just something I noticed when doing (EDITCALLERS '\MAPMDS SYSFILES). When you do a HARDRESET it clears all the stack pointers because they are invalid because the stack has blown away. And it relies on \MAPMDS to find all the STACKP pages and clear them.
For example, go into a break window (HELP). In the break window (SETQ FOO (STKPOS 'EXEC)).
you can use FOO to do STKEVAL: (STKEVAL FOO 'FOO) for example. THen (HARDRESET). If you look, FOO is not 'released'.  If you do (STKEVAL FOO 'FOO) again you'll get an error -- I just did it and got a RAID call.

nbriggs:
It looks as though the storage switching is concerned about the 8 MB/32 MB boundary, see LLDATATYPE: \CHECKSTORAGEFULL and other related functions.  `\MAPMDS` seems to be trying to figure out whether MDS is in one contiguous region, or two non-contiguous regions -- the bit of code that is doing essentially `(for I from 0 to (cond (... 0) (T 1)) do ...)` should execute once or twice (not zero or once)?

masinter:
will resubmit once I find out how memory pages are allocated and thin through the cases where the storage allocations changewile \MAPMDS is iterating.

masinter:
we went over this at the Monday meeting. The test case I used was 
`(for I from 1 to 10 do (ADD.PROCESS (LIST 'HELP I)))` to make a bunch of processes that I could type into if needed. I edited (just for diagnostics) CLEARSTK to have a PRINT option that doesn't clear anything but instead just PRINTs the stack pointer. CLEARSTK(PRINT) prints nothing before this PR, prints stack pointers for each process. 

(STORAGE) seems not to work correctly in master, prints out much more data after.  See also the STORAGE LispUser package which shows a bar chart view.

rmkaplan:
I'm running with this LLDATATYPE.  I'm getting stack overflows with the following printout (I don't know whether these are real--the backtrace in URAID is empty).  

WARN: freestackblock:StackFull MP9319 

*Error* Stack Overflow, MUST HARDRESET!
Enter the URaid
CL:NIL

< lu
BXBXBXFBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBFBXBXBFBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBFBXBXBFBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBXBFBXBXBFBXBXBXBXBXBXBXBXBXBXBXBX  etc.

Stack Check done

lu after this doesn't show anything useful.

Some times h (hardreset works), sometimes I get 

*Error* Segmentation fault: 11 at address 0x303da804f.
Please record the signal and code information
and do a 'v' before trying anything else.
Enter the URaid
CL:NIL

There may be something in my code that is causing the overflow, I'm looking. It appears to be happening when I try to create the TEDIT process on textstream that has already been produced.  But URAID doesn't provide anything useful.

rmkaplan:
I determined that the stack overflow itself was my fault (an implicit recursion through the BIN opcode).  But the hard-reset still is flaky.

masinter:
this PR doesn't purport to fix all HARDRESET problems -- just the ones caused by leaving around STACKPs that no longer valid because the page numbering was hidden.  The only reason I hesitate is that I still haven't figured out (don't remember) how memory is laid out before, during and after the 'switch'.

MattHeffron (approved):
"Group review" in Zoom meeting March 4, 2024.
Approved.

PR #1161, opened 2023-04-18
minor usage tweaks

masinter:
- Init file set HELPTIME to 1 more break windows; copyright NEVER
- SYSEDIT also sets copyright flag
- add internal/MAINTAIN.TXT to explain what MAINTAIN is


fghalasz (approved):
LGTM.

PR #1164, opened 2023-04-19
Allow (CLOSEALL) to work after all

masinter:


rmkaplan (approved):
LGTM.  Even if multiple-streams, CLOSEALL shouod close them all.

rmkaplan:
Actually, this fix wasn't quite right.  It looked good, but on testing I realize that the variable \OPENFILES is no longer maintained.  The loop should iterate over the list returned by (OPENP)

PR #1166, opened 2023-04-20
fix HPRINT compilation warnings, share GETFN check

masinter:


masinter (commented):
Do not merge, will resubmit

masinter:
@rmkaplan I'm having trouble understanding what's going on with the TEdit files in docs/dlisp; the Image Object code gets called but then the TEdit file appears with a "WARNING: Document contains unknown image objects" 

image

I thought i was able to view these before.

masinter:
if this is still a problem I willresubmit

PR #1172, opened 2023-04-22
Change cash-file initialization to avoid #.(pathname "NIL") error

masinter:
and recompile HPRINT with exports.all loaded


nbriggs (commented):
Why is a recompile of HPRINT.LCOM in the cash-file-init PR?

While this fixes the problem of there being a `#.(PATHNAME "NIL")` entry on the list by smashing it, how did it get there in the first place?  I think the list *is* supposed to be managed by the `XCL::ADD-WHERE-IS-DATABASE` fn, as it was previously.

fghalasz:
@nbriggs See my comment on the original issue filed by Paolo.  This was an online (and Medley Docker image) configuration issue caused by Notecards calling `(XCL::ADD-WHERE-IS-DATABASE NIL)`.

This pull request won't fix because Notecards called after MEDLEYDIR-INIT.  Need to put a guard into `XCL::ADD-WHERE-IS-DATABASE` to prevent NIL pathnames.

Have fixed in online already.  Will fix in apps.sysout in Medley Docker image soon.  And will eventually fix Notecards so that it finds the right hash file always - so it never puts NIL in this call..

nbriggs:
OK - so I think this PR should go away, a recompiled HPRINT should get its own PR.
The fix isn't as simple as ensuring that we don't put NIL pathnames on the `XCL::*WHERE-IS-CASH-FILES*` list -- it will have exactly the same behavior if the pathname of any non-existent file is on the list.

masinter:
CALLING (PATHNAME NIL) is an error... now signals.
(XCL::ADD-WHERE-IS-DATABASE ...) no longer will add a (pathname "NIL") even if it tried.

sources/CMLPATHNAME
```diff
@@ -213,8 +209,13 @@ Copyright (c) 1986-1988, 1990 by Venue & Xerox Corporation.
...
 
 (PATHNAME
-(CL:LAMBDA (THING) (* hdj " 2-Apr-86 11:01") (* ;; "Turns Thing into a pathname.  Thing may be a string, symbol, stream, or pathname.") (CL:VALUES (CL:PARSE-NAMESTRING THING)))
-)
+  (CL:LAMBDA (THING)                                         (* ; "Edited 30-Apr-2023 13:59 by lmm")
+                                                             (* hdj " 2-Apr-86 11:01")
+
+         (* ;; "Turns Thing into a pathname.  Thing may be a string, symbol, stream, or pathname.")
+
+         (CL:UNLESS THING (CL:ERROR "~S not valid pathname" THING))
```
nbriggs, line 217: I think perhaps this should be
```
(CL:CHECK-TYPE THING (AND (NOT NULL)
                          (OR STRING STREAM PATHNAME SYMBOL)))
```

SBCL says that the argument to PATHNAME should be `(OR STRING PATHNAME SYNONYM-STREAM FILE-STREAM)` but that doesn't seem quite right for us.
It's also not clear to me whether this check should be in `PATHNAME` or `PARSE-NAMESTRING`...

masinter:
redid PR #1188

PR #1180, opened 2023-04-25
move 'other' fonts into displayfonts directory

masinter:
As per LispCore discussion, just move the fonts from 'other' into the appropriate displayfont folders.


masinter:
These are mainly bogus copies of TITAN at display sizes. This isn't the right fix.
Maybe move SIGMA ?

PR #1182, opened 2023-04-27
Another pass on CLOSEALL simplify by calling (OPENP)

masinter:
Perhaps this is better?


nbriggs:
Seriously bad idea to, for example, close your LAFITE mail folder (which is (OPENP)) while it is being compacted.

rmkaplan:
There seem to be USERVISIBLE, USERCLOSEABLE, and CLOSEALL fields/properties on streams which presumably are supposed to guard against this in some way.  I don’t think they are documented, and I don’t remember them.

But LA.OPENTEMPFILE for example has a (WHENCLOSE STREAM ‘CLOSEALL ’NO) expression.

Maybe CLOSEALL (or maybe CLOSEF) should filter the ones that are not uservisible and userclosable

nbriggs:
`CLOSEALL` used to handle that -- see the differences between the old CLOSEALL and the proposed one.

masinter:
just to be safe, i put the checks back in.

masinter:
Removed all the never-used mentions of \OPENFILES everywhere...

rmkaplan (approved):
LGTM

PR #1185, opened 2023-04-30
add back library tcp sources

masinter:


nbriggs:
That doesn't include TCPOPS, which is the interface to the TCP subrs giving access to the host OS TCP implementation.  I think what we're looking for is https://github.com/Interlisp/envos/blob/main/xd1d/users/sybalsky/savoir/TCPOPS - but having the older files present is useful too.

masinter:
TCPOPS  added too

rmkaplan (changes requested):
I think all of these files should be down a level, in a TCP subdirectory, and conform to our hyphenated subdirectory naming convention.  So, TCP/TCP but TCP/TCP-CHAT...

masinter:
they don't work because the SUBR calls were written for SunOS TCP. I'm just copying into the Medley repo but perhaps they belong in medley/obsolete/library/tcp/.  I'm reluctant to change the structure without understadsing what will work when.   These are mainly independent packages not meant to allways loading all the TCP/layer packages.

masinter:
replaced by PR #1283 ane #1284. The first one copies the TCP files to  obsolete/tcp (rather than library) and #1284 does a MAKEFILE NEW on all or them.

PR #1193, opened 2023-05-03
don't declare most interpress constants 

masinter:
- rewrite INTERPRESS to not pollute pool of CL:CONSTANTP variables in common use
- update some functions that are (inappropriately) tied to INTERPRESS

Previously, if you had LOADFROM(INTERPRESS) or LOADCOMP(INTERPRESS), you'd run a problem...

A "constant" declaration in Medley is forever -- the variable is compiled as being the value at time of compilation forever after (in the same image). Common Lisp has defconstant that does the same, but the problem is mainly avoided by the package system.

I could have moved INTERPRESS into it's own package, but I just introduced a macro \IPC for "Interpress Constant" which pulled the constant value out of an alist.

. WHO @ CL:CONSTANTP IS BOUND BY WHOM

now is clear, even in fuller.sysout (which has all source files loaded and Masterscope analyzed.


library/SKETCHOPS
```diff
@@ -600,29 +598,31 @@ Copyright (c) 1986, 1987, 1990, 1992, 1993 by Venue & Xerox Corporation.  All ri
...
 (DRAW.LOCAL.SKETCH
-  [LAMBDA (LOCALSPECS STREAM STREAMREGION SCALE)         (* ; "Edited 24-Mar-92 14:00 by jds")
+  [LAMBDA (LOCALSPECS STREAM STREAMREGION SCALE)             (* ; "Edited  2-May-2023 13:28 by lmm")
+                                                             (* ; "Edited 24-Mar-92 14:00 by jds")
 
     (* ;; "draws the local specs on a stream")
 
     (* ;; "set priority of the stream in case mode is set to REPLACE or ERASE --- would be better to scan list looking for an element that actually has one of these.")
 
+    (MOVD? 'NILL 'SET.PRIORITYIMPORTANT)
```
nbriggs, line 608: Why does this make sense when SET.PRIORITYIMPORTANT is defined in this file?  If there's a reason for it, how about a comment explaining?

masinter: The MOVD? was unnecessary. I mistakenly thought the function was on INTERPRESS
Now gone.

nbriggs (commented):
not sure what the change to sources/WINDOW (declaring \SCREENS and \SCREENBITMAPS global) has to do with the Interpress constants.

masinter:
Sigh -- idle function SCREENPAPER gets a breakpoint in SHEDSCAN trying to do a (CDR 'NOBIND).   
image
ORIG? Where does ORIG come from?
image

sigh -- there was a conflict in records: the record LINE became the reccord XXLINE. *EXCEPT* there was a macro that had (WITH LINE .. stuff ...) which used fields from XXLINE but not from LINE.

Now I need to go bac through all the record conflicts and make sure they haven't infected other files.
Ugh.

sources/XXFILL
```diff
@@ -572,35 +588,35 @@ Copyright (c) 1985-1990, 1993, 2021 by Venue & Xerox Corporation.
...
-           ELSE (SETQ XTOX 0)                            (* ; "transpose & scale")
-                 (SETQ XTOY MicasToDev))
-         [IF (NUMBERP (CAAR POINTS))
-             THEN (SETQ POINTS (IMLTLIST POINTS XTOX XTOY 0 XTOY XTOX 0))
-           ELSE (SETQ POINTS (for I in POINTS
-                                    collect (IMLTLIST I XTOX XTOY 0 XTOY XTOX 0]
-         (APPENDOP.IP STREAM DOSAVESIMPLEBODY)
-         (APPENDOP.IP STREAM {)
+         (if (EQ SCANDIRECTION 'X)
+             then (SETQ XTOX MicasToDev)                     (* ; "just scale")
```
nbriggs, line 600: You turned MicasToDev into a \IPC constant, no?

sources/XXFILL
```diff
@@ -572,35 +588,35 @@ Copyright (c) 1985-1990, 1993, 2021 by Venue & Xerox Corporation.
 
 (POLYSHADE.SCAN.IP
   [LAMBDA (STREAM POINTS TEXTURE OPERATION WINDNUMBER SCANDIRECTION)
+                                                             (* ; "Edited  2-May-2023 09:12 by lmm")
                                                              (* ; "Edited  1-Feb-89 18:53 by FS")
 
     (* ;; "Convert micas to device units, and transpose, tell Interpress to take dev units back to micas, convert knot list into internal data structures")
 
     (LET (ILIST LLIST (MicasToDev 0.1181102)
```
nbriggs, line 596: See comment below...

sources/XXFILL
```diff
@@ -572,35 +588,35 @@ Copyright (c) 1985-1990, 1993, 2021 by Venue & Xerox Corporation.
...
-             THEN (SETQ POINTS (IMLTLIST POINTS XTOX XTOY 0 XTOY XTOX 0))
-           ELSE (SETQ POINTS (for I in POINTS
-                                    collect (IMLTLIST I XTOX XTOY 0 XTOY XTOX 0]
-         (APPENDOP.IP STREAM DOSAVESIMPLEBODY)
-         (APPENDOP.IP STREAM {)
+         (if (EQ SCANDIRECTION 'X)
+             then (SETQ XTOX MicasToDev)                     (* ; "just scale")
+                  (SETQ XTOY 0)
+           else (SETQ XTOX 0)                                (* ; "transpose & scale")
+                (SETQ XTOY MicasToDev))
```
nbriggs, line 603: and here...

sources/XXFILL
```diff
@@ -572,35 +588,35 @@ Copyright (c) 1985-1990, 1993, 2021 by Venue & Xerox Corporation.
...
+         [if (NUMBERP (CAAR POINTS))
+             then (SETQ POINTS (IMLTLIST POINTS XTOX XTOY 0 XTOY XTOX 0))
+           else (SETQ POINTS (for I in POINTS collect (IMLTLIST I XTOX XTOY 0 XTOY XTOX 0]
+         (APPENDOP.IP STREAM (\IPC DOSAVESIMPLEBODY))
+         (APPENDOP.IP STREAM (\IPC {))
          (SCALE.IP STREAM 8.466666)                          (* ; "2540micaspi / 300dpi")
```
nbriggs, line 609: Feels like this should be a constant we have...

masinter: I think I'll try a different approach

nbriggs:
Not sure why XXFILL declares (in the COMS) `(SPECVAR \FILL.WRULE)` but never uses it (nor does it appear anywhere else in the system sources), and `(DECLARE (SPECVAR FILL.WRULE))` (no leading backslash) in a couple of functions.

rmkaplan (approved):
It looks like the various requested changes have been installed.

PR #1194, opened 2023-05-05
include in release fuller.database .sysout and *.dribble for better records

masinter:
we don't do loadups that often, and including the dribble files, the masterscope database and (somewhat controersial?) the sysout with all the sources loaded etc. 
"might come in handy"


masinter:
redone with PR #1280

masinter:
replaced by #1280

PR #1204, opened 2023-05-08
CALENDAR, AOFD:  Calendar had a Y2K problem, AOFD did not allow string-streams to reopen

rmkaplan:
There still may be some Y2K issues, I fixed the obvious ones, including how it searched for the next field in Tedit.  It still does a lot of date hacking, which may still be goofy.  It would benefit from using more general functions.

I extended the open function for basebyte streams so they now reopen and set the fileptr to 0


lispusers/CALENDAR
```diff
@@ -1165,7 +1167,7 @@ Copyright (c) 1985-1990 by Xerox Corporation.
                                                                (CHARACTER 13)
```
nbriggs, line 1167: Is `(CHARACTER 13)` what it should be doing here?  I would have thought since we have EOLCHARCODE (from BOOTSTRAP) that (CHARACTER EOLCHARCODE) would be  clearer about its intention.   Perhaps that's a change for another time though.

rmkaplan:
Didn’t know that there was such a thing as EOLCHARCODE.  It should be (CHARCODE EOL) everywhere.  Maybe Bootstrap needs something earlier in the loadup, before CHARCODE is defined?

masinter (approved):
reviewed 7/7/23 zoom call

PR #1205, opened 2023-05-10
Prettyfileindex knows about :EDIT-BY

masinter:


rmkaplan (approved):
I think I'm now seeing what I expected

PR #1212, opened 2023-05-16
Fix places where rename of record LINE to XXLINE was incomplete

masinter:


rmkaplan (approved):
I actually don't know what this file does, but the LINE → XXLINE looks like just a good book-keeping correction

PR #1213, opened 2023-05-17
Pass --display/-d argument as -display not by setting DISPLAY env variable

nbriggs:
The way it was originally coded, `-display foo` would get passed as-is (in the passthrough arguments) but `-d` and `--display` would get transformed into setting the `DISPLAY` environment variable to `foo`.  Let's not promote confusion by having them work differently, and not prevent the user from setting `DISPLAY` (outside run-medley) when they want to use `-d` or `--display`.

### DIFF
--- a/run-medley
+++ b/run-medley
@@ -41,6 +41,7 @@ pass=""
 mem="-m 256"
 scroll=22
 noscroll=""
+display=""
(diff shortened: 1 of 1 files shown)


masinter (approved):
lgtm

PR #1219, opened 2023-05-24
fix GATHER-INFO for compiled code -- globalvars vs.bounc

masinter:


rmkaplan (approved):
Seem just to add BOUND-BY, looks reasonable.

PR #1220, opened 2023-05-25
Recompile PRETTY

masinter:


rmkaplan (approved):
It appears that this is just a recompile of the pre-existing version.  I'm not sure what the issue was, but seems reasonable.

masinter:
I'm not sure what happened here either.... I was having some trouble dealing with overlaps in PRs.

I think at a minimum PRC should check to see if the result of applying the PR is consistent (source file is same filedate as what was compiled) and other consistency checks. 

With a little more work, we could make a "compiled function fingerprint" that would show up changes to variable/macro definitions.

rmkaplan:
When I first started to unscramble all of the archives, I wrote some little subfunctions in COMPAREDIRECTORIES for trying to match up versions in a number of different ways.  Take a look at the end of COMPAREDIRECTORIES.TEDIT, maybe some of the other tools there could be integrated into GITFNS.

PR #1225, opened 2023-06-05
Add on:schedule to buildRelease workflow so that Medley builds happen automatically on Mondays at 1am

fghalasz:


masinter (approved):
not sure what it's doing or how to test it without merging it, but we'll know soon enough.

PR #1246, opened 2023-06-15
Updated Tedit files for a few Lispusers packages--formatting and typos

rmkaplan:
Mostly fixing heading and tabs to be consistent with other files


masinter (approved):
SGTM  (Sounds good to me)

PR #1247, opened 2023-06-15
INSPECT:  Sort datatype fields alphabetically, fixed a small bug

rmkaplan:
The fields of a datatype with more than 5 fields will be sorted alphabetically.  This is controlled by the variable INSPECTDONTSORTFIELDS:

If INSPECTDONTSORTFIELDS is T, fields for all datatypes will not be sorted.
Otherwise, the (more than 5) fields of a datatype will be sorted, unless INSPECTDONTSORTFIELDS is a list containing the name of the datatype.  Thus, if the list contains FDEV, filedevice fields won't be sorted--the alignment with Maiko would be seen in the ordering.

For now INSPECTDONTSORTFIELDS defaults to T, to preserve old behavior; set it to NIL or add a datatype name with ADDVARS in your INIT.  If this is generally attractive, the default can be changed to NIL.


masinter:
I think this is dangerous sllways and don’t see what problem it is solvingBlock records overlay datatypes for example with different field names If you mean only inspect window ?--LarryMasinter.netOn Jun 15, 2023, at 12:00 AM, rmkaplan ***@***.***> wrote:﻿The fields of a datatype with more than 5 fields will be sorted alphabetically.  This is controlled by the variable INSPECTDONTSORTFIELDS:
If INSPECTDONTSORTFIELDS is T, fields for all datatypes will not be sorted.
Otherwise, the (more than 5) fields of a datatype will be sorted, unless INSPECTDONTSORTFIELDS is a list containing the name of the datatype.  Thus, if the list contains FDEV, filedevice fields won't be sorted--the alignment with Maiko would be seen in the ordering.
For now INSPECTDONTSORTFIELDS defaults to T, to preserve old behavior; set it to NIL or add a datatype name with ADDVARS in your INIT.  If this is generally attractive, the default can be changed to NIL.



You can view, comment on, or merge this pull request online at:
  https://github.com/Interlisp/medley/pull/1247

Commit Summary

  4be6bc4  INSPECT:  Sort datatype fields alphabetically, fixed a small bug


File Changes  (2 files)

  
    M
    sources/INSPECT
    (132)
  
  
    M
    sources/INSPECT.LCOM
    (0)
  


Patch Links:

  https://github.com/Interlisp/medley/pull/1247.patch
  https://github.com/Interlisp/medley/pull/1247.diff

rmkaplan:
BLOCKRECORDS are not DATATYPES (but I see that I didn’t expect BLOCKRECORDS to go through the same internal function—I’ll put in an explicit DATATYPE test).

It is solving the problem of trying to find a something in particular when you are inspecting an instance of a datatype with 20 or so fields.  Just like the bad filtering in backtrace, the randomness makes debugging much more tedious and difficult, for no good reason.

masinter:
but rather than sorting the fields of the DATATYPE at the time of declaration (which is what it sounds like) and a global variable, why not just make it a property of the inspector. Fields sorted or not.
Rather than 5, just make it a "Sort" button.  Just two different views.

rmkaplan:
It sorts at the time of inspection, not at the time of declaration—this is underneath INSPECT, not part of the record package.

Another embellishment would be to add an item to the inspect-title menu, alongside the refetch, IT gets selection:  "Sort fields" if the fields are currently unsorted or "Unsort fields” if the fields are current sorted.

The global setting then would just define the defaults for instances of particular datatypes, but it could be dynamically switched if the default for a given instances is not what is desired at that point.  Even with a menu control, a global default would still be helpful

As you might suspect, this is a distillation of my frustrated experience with the large Tedit structures (but also external formats, file-devices, streams, etc.).

masinter (changes requested):
There are multiple sort orders that might make sense. As listed in the original declaration, as laid out using the datatype data layout algorithm (whcih won't split a field over a field-type boundary, alphabetically using original datatype declarations... it's also possible to have datatypes initialized and inspectable without knowing the field names (initrecords don't use field names).

S

rmkaplan:
True, but what was important to me, for this purpose, was to get an intuitively understandable order that would let me rapidly find the value of a particular field. Without me having to know any of the other considerations that might have been relevant to the actual layout of fields, either in the declaration or in the instances. If I want to know or manipulate more of those implementation details, I can look at the declaration.

Sorting alphabetically is quick and obvious. Other possibilities would involve some research.

masinter:
@lispcore would like other opinions

masinter:
> Another embellishment would be to add an item to the inspect-title menu, alongside the refetch, IT gets selection: "Sort fields" if the fields are currently unsorted or "Unsort fields” if the fields are current sorted. The global setting then would just define the defaults for instances of particular datatypes, but it could be dynamically switched if the default for a given instances is not what is desired at that point. Even with a menu control, a global default would still be helpful

I wish we had a clearer policy about what kinds of enhancements we were making to Medley.
Adaptation to modern IO devices (keyboard, 1-button mouse, touch screen, higher resolution, different fonts etc), fixing bugs and completing work that was in progress were some of the categories.

What is the policy?

masinter (approved):
discussed 3/4/24 implementation weekly, let's go for it.

PR #1251, opened 2023-06-16
put back in IRM.SMART.LOOKUP needed for helpsys from menu

masinter:
It was incorrect to remove IRM.SMART.LOOKUP because it is called (via an APPLY) when entering something in the menu to look up. This PR adds back the function.


rmkaplan (approved):
LGTM

PR #1256, opened 2023-06-18
Minor updates to llread and sedit

rmkaplan:
LLREAD is more careful about CHARACTERNAMES and CHARSETNAMES.

SEDIT formats SELECTC like SELECTQ, copy-select in the title bar does bksysbuf of the thing being edited.  (although BKSYSBUF may not get the package)


masinter (approved):
reviewd 7/7/23 zoom

PR #1260, opened 2023-06-24
PDFSTREAM:  First implementation

rmkaplan:
This introduces library/PDFSTREAM that piggy-backs on POSTSCRIPTSTREAM to allow Medley to (appear to) create PDF hardcopy files.  It depends on the availability of a Unix Shell utility (ps2pdf from ghostscript, pstopdf built-in on the mac) to do the PS to PDF conversion.

The user has to provide the name of the utility (the Medley variable PDFCONVERTER or the GETENV PDFCONVERT) and also a template for the way the unix Shell command should be called (the alist PDF-CONVERT-TEMPLATES).  If we have a list of candidate converters, e.g. for windows, we can include the relevant templates in the package.

Currently this uses the function PROCESS-COMMAND from GITFNS and SLASHIT from PSEUDOHOSTS.  I don't remember now why I didn't use ShellCommand in GITFNS and rolled a different interface--it would be worth Nick/Larry/?? taking a look.

SLASHIT converts <> to // in file names.  Maybe there is a standard function for that someplace.

Currently Tedit puts the name of the /tmp PS file in its promptwindow.  If it is harmless to smash the fullname field of the stream, I can easily clean that up, otherwise more work.

POSTSCRIPTSTREAM is slightly modified here, to add an extra field to its image data to hold the eventual pdf filename.

HARDCOPY is modified just to improve the printer menu that it puts up, I don't think there are any changes there that have to do with PDFSTREAM.

I have marked this as draft for now, it needs some eyeballing and a longer work-out.

Note that when this is loaded, it sets the DEFAULTPRINTERTYPE to PDF.  Maybe that should be contingent on whether PDFCONVERT is defined.


masinter (dismissed):
I hate to see a lot of effort going into this hack, rather than the "right" alternative: a true PDF image stream. There are free open source Common Lisp PDF "drivers" that one could start from.  And it isn't as portable (ps2pdf not available or tested, not part of installer, etc. e.g., in Cygwin-based Windows installs.)

 Shouldn't this go into "LispUsers" or some other collection of "recent hacks"? Is there a .TEDIT documentaion?

rmkaplan:
A true PDF image stream would obviously be much better, but I assume that that is a much bigger project that somebody (else) would have to take on.  Short of that, this seems to work at least in environments that do have a postscript-to-pdf converter.  (BTW, I imagine that ps2pdf has been more tested than any of the available Common Lisp drivers).

In terms of effort for this, the hardest part was figuring out how to add a new streamtype to the hardcopy interfaces that would be needed even if we had a native implementation of pdf streams.  (Which is not to say that there are still some interface issues—useful error reporting from the converter when the postscript file is invalid, and getting the pdf stream before it is closed to exhibit its destination file name instead of the name of the underlying postscript temp file.)

The documentation for this is trivial: Use PDF instead of POSTSCRIPT as the imagestream/printer type, then see POSTSCRIPTSTREAM.TEDIT.  If we had a native PDF image stream, then documentation would be an issue.  It presumably it would require its own specificagtion of fonts, regions, etc.

Could be moved to lispusers.

masinter:
@MattHeffron your thoughts?

masinter:
i'd like to withdraw my request for changes. Why is this PR marked Draft?

rmkaplan:
I marked it draft because it was a first implementation and I wasn’t sure about some of the interface issues—and the close bug that Nick figure out.

Nick also suggested a better way of setting up the postscript interaction (using the AFTERCLOSE streamprop) that doesn’t involve changing the postscript image objects.

So I will update this with these improvements.

rmkaplan:
I committed a new version of PDFSTREAM that doesn't depend on changes to POSTSCRIPTSTREAM.  Instead, it uses the AFTERCLOSE stream method that Nick suggested.  PDFSTREAM also depended on subfunctions PROCESS-STREAM (on GITFNS) and SLASHIT (on PSEUDOHOSTS).  Those are general Unix utilities (like ShellCommand), so I have moved those to UNIXUTILS and removed them from GITNFS and PSEUDOHOSTS>

There were other, mostly minor cleanup, PRs for GITFNS and PSEUDOHOSTS that have been languishing for several months.  The versions of GITFNS and PSEUDOHOSTS committed here include those previous changes.  I am going to remove the earlier PRs to avoid confusion.

I have also removed the draft status from this PR, it is ready to be reviewed, tested, merged.

(I'll note, though, that Nick's fix to UFSCLOSEF has not yet shown up as a PR, so there may be some device-error goofiness with pdfstream until that gets promulgated.)

nbriggs:
I tried to do a loadup in the PDFSTREAM branch, and it failed with:

Screen Shot 2023-09-24 at 11 36 00 AM

while trying to do loadup-aux.sh

rmkaplan:
Something in the loadup scripts?  I usually do loadup-full.sh or loadup-all.sh.  loadup-full.sh gives a message in the copy phase that whereis.hash and exports.all don’t exist in the tmp directory and can’t be copied, but otherwise everything goes through.

I don’t think that the files here would impact the loadup.

nbriggs:
I've tried loading it into a different full.sysout, and it failed because `(TRUEFILENAME <closed ps file>)` is the closed stream name not the name of the file, and `(INFILEP ...)` of that returned NIL.  Not sure exactly what's going on.

nbriggs:
Re: the failed loadup - it gets through the loadup-full.sh parts but fails in the loadup-aux.sh part -- so probably something that you/others load in your init file but I don't, and isn't expressed as a dependency in the script.

rmkaplan:
> I've tried loading it into a different full.sysout, and it failed because `(TRUEFILENAME <closed ps file>)` is the closed stream name not the name of the file, and `(INFILEP ...)` of that returned NIL. Not sure exactly what's going on.

I haven't seen that.  Is PSEUDOHOSTS loaded in that sysout?  TRUEFILENAME is dummied out to IDENTITY if PSEUDOHOSTS isn't loaded.

nbriggs:
Nope, I don't use PSEUDOHOSTS by default - not clear why that should prevent being able to PDF a file.

nbriggs:
So it seems odd that INFILEP return NIL on the closed stream when `(OPENSTREAM <closed stream> 'INPUT 'OLD)` is OK, and the manual says that `(INFILEP foo)` is equivalent to `(FULLNAME foo 'OLD)` which works correctly on the closed stream.

rmkaplan:
If you pass PS-TO-PDF a postscript file with a pseudohost name (e.g.  {MEDLEY}/sources/ADIR.PS), it has to figure out where that is on the underlying unix file system, so it can pass down its unix name.  (And if itdoesn’t already exist in Unix, then it has to be copied on to Unix—I relaize that I haven’t yet done that case.)

The dummy TRUEFILENAME on ADIR may not be sufficient if it just passes thru the stream without pulling out the underlying name.  Maybe it should have been doing INFILEP instead of EVQ, but apparently that doesn’t work yet either. Of just do its own FULLNAME.

I will try to run with TRUEFILENAME backed off to EVQ and see if I get your error.

(Independent of this, almost all of the few remaining calls to OPENFILE need to be changed to OPENSTREAM.  OUTFILE and INFILE, for example, always break withn a “Litatom streams are not supported” error.  I’ve gone through and fixed most of those, will gather the changes into another PR.  But maybe we should just get rid of OPENFILE in favor of (MOVD ‘OPENSTREAM ‘OPENFILE)).

rmkaplan:
Is there a predicate that will tell me whether a file resides in the unix file system, or just assume that the answer is no if the host is not DSK or UNIX?

rmkaplan:
> Re: the failed loadup - it gets through the loadup-full.sh parts but fails in the loadup-aux.sh part -- so probably something that you/others load in your init file but I don't, and isn't expressed as a dependency in the script.

Does the loadup-aux depend in any way on the local init file?

nbriggs:
The embedded init file is missing a load of `where-is` --
```
diff --git a/scripts/loadup-aux.sh b/scripts/loadup-aux.sh
index 598f6955..e0be53d4 100755
--- a/scripts/loadup-aux.sh
+++ b/scripts/loadup-aux.sh
@@ -15,6 +15,7 @@ cat >"${cmfile}" <<"EOF"
   ((WORKDIR (IL:CONCAT (QUOTE {DSK}) (IL:UNIX-GETENV (QUOTE LOADUP_WORKDIR)) (QUOTE /))))
   (IL:MEDLEY-INIT-VARS)
   (IL:LOAD(QUOTE MEDLEY-UTILS))
+  (IL:LOAD(QUOTE WHERE-IS))
   (IL:MAKE-EXPORTS-ALL (IL:CONCAT WORKDIR (IL:L-CASE (QUOTE exports.all))))
   (IL:MAKE-WHEREIS-HASH
     (IL:CONCAT WORKDIR (IL:L-CASE (QUOTE whereis.dribble)))
```
and there may be other files it should ensure are present.

rmkaplan:
I tried loadup-all on my local machine in the pdf branch that I pushed, also tried it in a fresh clone master, neither failed.

So somebody must be loading WHERE-IS.

Note, however, that there is also a file internal/WHEREIS (no hyphen),  Maybe that should be removed, to avoid confusion?

fghalasz:
I get the same loadup failure as Nick does - in loadup-aux.sh with the error that XCL::WHERE-IS-NOTICE is not defined.

The cause is actually in loadup-full-from-lisp.sh - even though it shows up first in the subsequent loadup-aux.sh.

The cause can be found in the function LOADUP-FULL contained in the eponymous file internal/loadups/LOADUP-FULL(.LCOM).

LOADUP-FULL (the function) does a call to (LOADUP '(POSTSCRIPTSTREAM)) which fails because it appears the file library/POSTSCRIPTSTREAM.(.LCOM) has been removed in the PDFSTREAM branch/PR.

The result is that subsequent loadups (including library/WHERE-IS(.DFASL)) supposed to be done by LOADUP-FULL (the function) do not get done.

According to the comment POSTSCRIPTSTREAM is being loaded at this point in the loadup sequence to provide the function PSCFONT.READFONT, which in turn is used by a later call to LOADFULLFONTS in LOADUP-FULL (the function).

I can only presume that @rmkaplan has POSTSCRIPTSTREAM file lurking somehwre that is being picked up in his loadups.

So @rmkaplan I leave it to you how to fix LOADUP-FULL.

And note: a secondary issue is why the loadup-full-from lisp script doesn't report a failure in this situation.  I will open an issue on this.

-- Frank

rmkaplan:
That’s my fault.  An update to POSTSCRIPTSTREAM was needed in the previous version of PDFSTREAM but it was no longer needed with the revision using the AFTERCLOSE function.  So I thought that removing it from the branch was the way to back it out—as opposed to somehow restoring the older version.

I think what I will do, to clean this up, is create a new PR branch under master for PDFSTREAM with just the proper files, and close the existing branch.

Thanks for tracking this down, sorry for the confusion.

rmkaplan:
I decided on another approach:  I added the preexisting POSTSCRIPTSTREAM back into this branch, and pushed it.

I hope that has the desired effect.

Separately, I am going to create a new PR with Nick's fix to UFSCLOSEF (he doesn't have , which does benefit the PDF stream. Nick doesn't have mobile data right now, but recommended getting this done.

masinter (approved):
as per 10/2/23 meeting

PR #1262, opened 2023-06-27
VIRTUALKEYBOARD:  fix loadup

rmkaplan:
Just initializes CURRENTKEYBOARDCONFIG


masinter:
it might be helpful to change VIRTUALKEYBOARDS to put the keyboard files into the same subdirectory.
That way they're not spread out

possibly rename VIRTUAL.TEDIT to vrtualkeyboards/VIRTUALKEYBOARDS.TEDIT

rmkaplan:
That the MAIKOKEYBOARDS, DORADOKEYBOARDS, DANDELIONKEYBOARDS… files?

Are those only relevant to VIRTUALKEYBOARDS or used more generally?

If only for virtual, putting up a separate virtual directory is a good idea.

rmkaplan:
This has the changes described in #1267

rmkaplan:
The keyboadfiles (e.g. DANDELIONKEYBOARDS) are not referenced anywhere else (Loops, Notecards, Medley, Rooms), and are relevant only as inputs to VIRTUALKEYBOARDS

rmkaplan:
I have removed the DRAFT status of this.  This fixes problems with loading/using VIRTUALKEYBOARDS, and moves all other files that are relevant only to VIRTUALKEYBOARDS into the same library/virtualkeyboards subdirectory.

It also makes the keyboard files (MAIKOKEYBOARDS etc.) properly managed files, not with their own special format.

KEYBOARDEDITOR didn't work before and still doesn't work, still to be looked at.  But this at least puts it in a more reasonable place.

masinter (changes requested):
starting with current master 
`git checkout master` `git pull`
create a test branch
`git checkout -b testNN`
then merge this branch
`git merge VIRTUALKEYBOARDS`
then do a loadup
`./scripts/loadup-all.sh`
run it
`./run-medley -full` 
then `(FILESLOAD VIRTUALKEYBOARDS)` yields error: `INVALID KEYBOARD` `DEFAULT`

image

masinter (approved):

masinter (changes requested):
same problem: I updated the VIRTUALKEYBOARDS PR.
```
git checkout VIRTUALKEYBOARDS
git pull
./scripts/loadup-all.sh
medley --full -x -
```
if you check out this branch and do a loadup (just to make sure) and start up the new full.sysout with no user greet file and then try to
```
(FILESLOAD VIRTUALKEYBOARD)
```
you get the same error "INVALID KEYBOARD" DEFAULT.
On the stack you see a call to 
(VKBD.CREATE-KEYACTION-TABLE 'DEFAULT ...)
which calls (FINDVIRTUALKEYBOARD 'DEFAULT) 
So NEW_KEYBOARD is the atom DEFAULT, and the (MEMB NEW-KEYBOARD VKBD.KNOWN-KEYBOARDS) will always fail, because VKBD.KNOWN-KEYBOARDS is a list of lists (looks like an alist) rather than a list of keyboard names.
This in turn is called by (VKBD.ADD-DEFAULT-KEYBOARD)

The logic here is hard to trace... and the entire keyboard handling logic doesn't really handle the cases we want (modern keyboards). 
I'm not sure what got edited when.

rmkaplan:
OK, this is something to go on, I'll look, hopefully also figure out why I didn't see it.

rmkaplan:
I still can’t replicate this.  I checkedout virtualkeyboards, did a loadup-all, and ran in that sysout with arguments -x and -nogreet.

I then did (FILESLOAD VIRTUALKEYBOARDS), it loaded fine.  (But note:  in your comment, you had VIRTUALKEYBOARD without an S.  Is that another file that you have hanging around?)

Here’s is what I see (on a Mac):

￼

masinter:
@Interlisp/core could someone else try this? Starting with a new clone of Interlisp/medley and checkout VIRTUALKEYBOARDS and do ./scripts/loadup-full.sh
and then try and (FILESLOAD VIRTUALKEYBOARDS). I get an error INVALID KEYBOARD / DEFAULT while Ron doesn't.

masinter:
I can now load VIRTUALKEYBOARDS if, before calling run-medley, I do
```
export LDEKBDTYPE=X
```
but I don't understand why there is a linkage of the actual attached keyboard to what happens when trying to load.   The attached keyboard can change.

I also noticed that if I choose Keyboards > GERMAN that I get a keyboard where the U key has been replaced by the number 6. (There remains a key upper right for U with umlaut, but no unadorned U.)

masinter (approved):
I added a new issue #1366 to collect VIRTUALKEYBOARDS remaining problems, and will approve this shuffling.

PR #1263, opened 2023-06-29
ADIR:  \COPYSYS should close the sysout file

rmkaplan:
ADIR: \COPYSYS should close the sysout file, otherwise, when the sysout starts up it complains (and sometimes fails) that the tmp file can't be found.

So I put in (CLOSEF? --) before the RENAMEFILE, and that seems to eliminate that.  But I don't understand why:  that happens AFTER the \FLUSHVM, so how does the starting sysout know about that closing?

(SAVEVM) doesn't go through this code, but it doesn't work for me--\FLUSHVM says that it can't make the file.  Do I need to set a unix variable someplace for that?

Another observation: When making the sysout, SYSOUT calls \DEVICEEVENT with the tag AFTERDOSYSOUT, and with the tag AFTERSYSOUT when restarting.  But that is the only appearance of AFTERDOSYSOUT in sources/.  Is that a placeholder for some future distinction, or just a mistake?


nbriggs:
> Another observation: When making the sysout, SYSOUT calls \DEVICEEVENT with the tag AFTERDOSYSOUT, and with the tag AFTERSYSOUT when restarting. But that is the only appearance of AFTERDOSYSOUT in sources/. Is that a placeholder for some future distinction, or just a mistake?

I believe that is NOT a mistake.  There are `AFTERDOxxx` for MAKESYS, SYSOUT, and SAVEVM.  Per the comments, it's a chance for the `\DEVICEEVENT` and `\USEREVENT` code to get control in the original running code as it continues vs. the other `AFTERxxx` events which happen in the new resumed image.

This didn't use to be an issue but I think that the changes around the whole versioned file thing in this area have introduced new problems that we're still discovering.
Are you sure that a CLOSEF? here is really the right thing to do?

rmkaplan:
> Are you sure that a CLOSEF? here is really the right thing to do?

No, I’m not sure, and it doesn’t make sense.

But empirically, when I put that in, it stopped warning that it couldn’t find the temp file in the sysouts that I usually work in (and that I make locally), and it stopped the error when I started a sysout made in the release loadup.

But other things (that used to work) are also now screwed up.  If I make the sysout when connected to a pseudohost directory (e.g. {WMEDLEY}/library/tedit), then MEDLEYDIR is NIL when it restarts, and GIT-INIT complains that it can’t find the clones.    This even though the shell variable MEDLEYDIR hasn’t changed:  /Users/kaplan/…/working-medley or git-medley.

These are both relatively new symptoms—I noticed the temp-named sysout warning a few months ago, and I noticed the MEDLEYDIR problem just today.

Maybe the MEDLEYDIR problem would go away if I pack {DSK} onto the shell variable, and then we wouldn’t have to solve this.  But I haven’t been able figure out how to escape { and } in my .cshrc (a prehistoric but generally working file that I have always put off modernizing).

nbriggs:
Perhaps we can figure out what (else) is going wrong and/or why this change seems to work before committing it to master...

> But I haven’t been able figure out how to escape { and } in my .cshrc

Lots of double quotes to ensure that the braces don't get reinterpreted except where desired...
```
% setenv FOO "{dsk}foo"
% env | grep foo
FOO={dsk}foo
% setenv BAR "ABC${FOO}DEF"
% env | grep BAR
BAR=ABC{dsk}fooDEF
```

rmkaplan:
I now see (actually, I remember) that it is more complicated, because MEDLEYDIR is doing double duty.

MEDLEYDIR is used as a shell variable in the run-medley script, so it can’t have {DSK} attached.  But inside Medley, if it doesn’t have the {DSK} (or some other host), then it appears to be susceptible in some contexts to defaulting to the host of the currently connected directory.  The new behavior is, if that’s a pseudohost, then things are screwed up.

I tried enclosing it in (TRUEFILENAME…), but that also inherits the currently connected host.  

Is it safe to assume that DSK is the implicitly defaulted host of otherwise unhosted MEDLEYDIR?

nbriggs:
> Is it safe to assume that DSK is the implicitly defaulted host of otherwise unhosted MEDLEYDIR?

That would be a better default assumption than anything else I can think of.

rmkaplan:
I updated MEDLEY and added it to this PR, it now adds {DSK} to the value of (UNIX-GETENV "MEDLEYDIR"), as the default.

In terms of understanding the CLOSEF?, could it be that the \FLUSHVM closes the file AFTER it has dumped all the pages, and the restart then sees that the file is still open?  The CLOSEF? is a no-op after the sysout is made, because the file has been closed.  But actually does some work when the \FLUSHVM returns at the restart.

masinter (commented):
t'his is a patch on a patch on a patch.
There was some behavior on D-machines which was replaced by Maiko.
At one point the Maiko code was explicitly changed to not make new versions.
Then that was patched to make new versions by writing out a -temp and then copying it.

THis is all conflated by the two contexts of "ordinary user doing loadedup for themselves" vs "loadup done by github action".

There is another PR/Issue about -loadup and also about AFTERDOSYSOUT.

* There should be a place to document this stuff, but I don't know where it should be.
*  -loadup was added to mean "never GREET" to avoid having releases pollutedd by the init of the person making the loadup (if that's  a person at all).
* There are four VM-saving operations: SAVEVM, LOGOUT, SYSOUT, MAKESYS.
* There are three times when an event-function could be called:   BEFORE, AFTER RESTART and continuing the same system (AFTERDO).  But there is no AFTERDOLOGOUT and AFTERDOMAKESYS is unexpected. 
* There may be bugs in the current  implementation but let's not add more patches if we can help it.


There was a time when MAKESYS didn't make a new version.

rmkaplan:
There may be, and may have been, several different issues with the MAKESYS/SYSOUT/SAVEVM interface. But I don't think we need to figure out how to fix them all if we can figure out how to fix at least some of them.

For me, version numbers in particular are crucially important.  It may be that Medley was deliberately changed at some point in the past to remove version numbers from makesys and sysout.  But that may have been because of a failure of imagination about how to do it with new version-less Maiko interfaces rather than a deep insight that there is no purpose to having version numbers on sysout files (as has been the case since the days of Tenex).

I make sysout files (and even full loadups) several times a day, sometimes several times an hour, in order to keep consistent states as I plow through a lot of related changes.  And to protect against catastrophic failures as I'm pushing on low-level updates in Tedit and elsewhere.

The version issue is separate from the -loadup and the MEDLEYDIR issues (defaulting DSK, making sure to include MEDLEY-INIT-VARS in a greet-less loadup).  Once Frank pointed out that -loadup wasn't responsible for that and that I needed to do it explicitly, the problem really became one just of documentation.

For me at least, both those issues have been resolved.  Although it may be desirable to push the versioning capability down somewhere below the \FLUSHVM, I am now seeing the desired versioning and loadup behavior, and I don't know that anything else needs to be done.

There may still be other issues with whether or how individual modules insert things on the various event form-lists, and whether there should be more automatic ways of recognizing what to do when returning in different environments.   That's a separate discussion.

masinter:
We jointly have some use cases that we actually want to ALL work -- not some theoretical uses that we imatgine. And each use case has some requirements. We need to work through the use cases so we don't windup introducing new bugs while fixing others.

masinter:
As it happens, loadups get wrtten in 'tmp' and then copied into 'loadups' usng scrpts/copyall.sh which uses 'cpv' to copy. cpv makes versions. The versons in tmp, though, are pretty beuseless.

rmkaplan:
I removed the unnecessary CLOSEF? from the restart \COPYSYS, leaving the RENAMEFILE just after the creation.

Seems to work.

nbriggs:
There's pre-existing code there that isn't doing what the comments (included below) in `\UFS.RECOGNIZE.FILE` suggest:
```
;; "Perform recognition on FILENAME, returning the %"true%" name for
the file, or NIL.  The result file name is following the Xerox Lisp file naming
convention but does not include HOST field.  It will be supplied by \UFS.FULLNAME.
```
So the code that packs on the host probably should be doing a call to \UFS.FULLNAME instead of messing about on its own.

We can probably look into this afterwards.

rmkaplan:
The ADIR change is now included in #1333

PR #1269, opened 2023-07-09
Add ipfonts to release tars

fghalasz:
As per @rmkaplan's request, this PR adds fonts/ipfonts to the release tars (and installers).

Note, however, this PR does not affect the loadups - so the ipfonts are not loaded into full.sysout as are the postscriptfonts.  See LOADUPFULLFONTS in sources/LOADUP-FULL. 

-- Frank

### DIFF
--- a/.github/workflows/buildLoadup.yml
+++ b/.github/workflows/buildLoadup.yml
@@ -56,8 +56,8 @@ on:
 defaults:
   run:
     shell: bash
-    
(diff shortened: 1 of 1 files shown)


rmkaplan (approved):
I think that’s right, just ensure that the IP files are available if any quirky code still wants them. But they shouldn't pollute the sysout.

PR #1272, opened 2023-07-13
SEDIT-WINDOW:  Shift-title selection conforms to standard mouse-keyboard conventions

rmkaplan:
In previous commit/PR #1256 the bksysbuf wasn't waiting for the copy/shift key to come up.  With this minor update, nothing happens until the copy key comes up, and then the bksysbuf happens only if the mouse is still in the title bar.

(Is there someplace a standard subfunction that everybody uses to implement this functionality?)


masinter (approved):
reviewed 7/17/23 lgtm

PR #1274, opened 2023-07-18
Move exports.all to the loadups directory; add a LOADUPSDIRECTORIES to MEDLEY-INIT-VARS

fghalasz:
Just as the title says.

-- Frank


masinter (changes requested):
I think we might need to either 
(a) add LOADUPDIRECTORIES to DIRECTORIES, or
(b) change every occurrence of (FILES EXPORTS.ALL) to (FILES (FROM LOADUPS) EXPORTS.ALL). 
(c) make a file library/EXPORT with EXPORTCOMS set to ((FILES (FROM LOADUPS) EXPORTS.ALL)), and then change (FILES EXPORTS.ALL) everywhere to (FILES EXPORT) -- with variations calling it something other than 'EXPORT'.
(d) don't make EXPORTS.ALL but gather exports when EXPORT is loaded 

Which path to choose depends a bit on whether we want to have EXPORTS for subsets (EXPORTS for TEDIT?) or things not in the medley repo (LOOPS? ...)
I prefer (b) in the short run or (d)

rmkaplan:
I think that if the variable is named LOADUPDIRECTORIES, that you have to say (FROM LOADUP), not (FROM LOADUPS).  DOFILESLOAD truncates the DIRECTORIES part, although that could be changed.

fghalasz:
Also @rmkaplan   The variable is called LOADUPSDIRECTORIES, so (FROM LOADUPS) is correct.   I named it LOADUPS... because the key directory is called /loadups and not /loadup.

Can easily change if this is not esthetically pleasing to many of you.

-- Frank

lispusers/PSEUDOHOSTS
```diff
@@ -27,7 +33,8 @@
         (DECLARE%: EVAL@COMPILE DONTCOPY (RECORDS PHDEVICE PHGENFILESTATE TARGETDEVICE)
                (MACROS PSEUDOHOST.NAME PSEUDOHOST.TARGETVAL)
                (P (CL:UNLESS (GETP 'EXPORTS.ALL 'FILE)
```
masinter, line 35: the GETP is no longer necessary but harmless

rmkaplan:
I think it is necessary until loading EXPORTS.ALL adds a FILEDATES property.

masinter (approved):
looks good

PR #1277, opened 2023-07-19
FILEBROWSER:  FB.COPYFN uses COPYINSERT instead of BKSYSBUF

rmkaplan:
So shift-selecting produces strings for file names with funky characters


masinter (approved):
reviewed 7/19/23 lgtm

PR #1278, opened 2023-07-19
GATHEREXPORTS uses standard MAKEFILE machinery

rmkaplan:
So that it gets all the normal file properties (FILE, FILEDATES) when it is LOADed, interacts with DOFILESLOADs checking.  Also nothing special about setting up its reader environment.

This does it with an E command to PRETTYDEF, which prints out the MAPC expression, a little ugly on loading.  One alternative would be to put the E command in (PACK* TOFILE 'COMS), and then it would just show EXPORT.ALLCOMS (or TEDIT-EXPORT.ALLCOMS).  Or to embed the MAPC's in a little function that would print out nicer.


masinter (approved):
reviewed 7/19/23 LGTM

PR #1280, opened 2023-07-20
release dribbles and msfuller

masinter:
- ignore loadups/exports.all (moved)
- add fuller.database and dribble files to release


fghalasz:
LGTM.  I will approve.

But it doesn't really make sense to have two different release tar builders - the one in buildLoadup.yml for github actions versus release-medley.sh for local builds.  And they are a bit out of sync.  I will open a separate Issue to change buildLoadup.yml to just use release-medley.sh AND make whatever changes to run-medley.sh to sync it with what buildLoadup.yml was doing differently.

fghalasz (approved):
LGTM.

PR #1281, opened 2023-07-20
Simplify EXPORTS.ALL loading commands

rmkaplan:
Standard FILES commands now do the right thing, no need for extra junk


library/UNICODE
```diff
@@ -1194,8 +1191,8 @@
 )
 (DECLARE%: EVAL@COMPILE DONTCOPY 
 
-(OR (GETP 'EXPORTS.ALL 'FILE)
-    (PRINT "NOTE: UNICODE requires EXPORTS.ALL for compilation" T))
+(FILESLOAD (FROM LOADUPS)
+       EXPORTS.ALL)
```
masinter, line 1195: shouldn't this and others be
(FILES (FROM LOADUPS) EXPORTS.ALL)
Instead of (P (FILESLOAD ...) ...)

masinter (approved):
it's better to avoid P at all, because WHEREIS and FILECOMSLST of FILES will work.

PR #1284, opened 2023-07-21
tcp makefile new

masinter:
- pull (non-working) TCP files to obsolete
- LOAD and MAKEFILE NEW TCP files for legibility


masinter:
unnecessary, will prettyfileindex allsources

PR #1290, opened 2023-07-24
Change github workflows to provide MacOs full release zip file

fghalasz:
This PR does two things:

1.  Changes github workflow buildLoadup.yml to create a zip file release asset for MacOS containing the Meldey full release (Medley plus maiko).

2.  Changes the naming for all release assets to use the template medley-full-<PLATFORM>-<ARCH>-<RELEASE_TAG> (e.g., medley-full-linux-x86_64-230710-e4b6f69e_230607-d1f4653b.deb) instead of the current  medley-full-<RELEASE_TAG>-<PLATFORM>-<ARCH>.

The zip file created for MacOS can be downloaded and then unzipped into any directory - say ~/il.  Then you can run medley by cd to ~/il/medley and then run either ./medley or ./run-medley.

!!! NOTE: this zip file is not yet notarized by Apple, therefore you will need to clear the quarantine attributes before maiko (lde, et al) will run.  To do so, you will need to run `xattr -d com.apple.quarantine ~/il/maiko/darwin-universal/*` (assuming you installed into ~/il. !!!

I will work on notarization next.


masinter (approved):
Trust this release only

PR #1293, opened 2023-07-28
SPY.POINTERS is always T (no 2-byte atoms)

masinter:
spy-no-background redone:
* SPY.POINTERS is always T (no 2-byte atoms)
* SPY uses special \BACKGROUND to skip sammpling both \BACKGROUND-YIELD (calls to YIELD subr), or any sample under the background functdion if \BACKGROUND.PROCESS is non-NIL (the default)


masinter:
I pushed more changes after Approval -- please re-review

rmkaplan:
Is BACKGROUNDE a typo?

masinter:
\BACKGROUNDE was a typo. I wound up changing the name of the variable to \IGNORE.BACKGROUND to be different from the function name \PROCESS.BACKGROUND. Testing this, \IGNORE.BACKGROUND NIL shows warped results

rmkaplan:
So, still more work to do?

Issue #1294, opened 2023-07-29
MODERNIZE: Odd behavior in window-moving by titlebar clicking


rmkaplan:
When you click in the moving-region of the title bar (a bit away from the corners), it puts up a ghost region and the cursor jumps to the nearest upper corner of that region (upper-left  or upper-right depending where you click).  You drag from there.

Basically, the mouse jumps to the corner.  But if the window has already been moved so that that corner is not on the screen, then something else happens:  the corner jumps to the mouse.  The ghost region doesn't start as an outline of the window's current region, it's a region of the same size but positioned somewhere else.

For example, slide an SEDIT window off so that it is off screen to the left, click to the left of what would have been the midpoint, and you see the jump.  

It is also set up to show as the ghost region the smallest region that encloses all of the attached windows.  Thus, suppose you have a Tedit window with its paragraph or character menu attached above, and you click in the main-windows title bar.  The cursor jumps to the upper corner of the region that includes the menu, i.e. the upper corner of the menu.  If that part of the menu is off-screen, again, that corner makes a big jump down to the mouse.

One (easy) option would be to have the mouse jump to the nearest on-screen corner, e.g. to go to a bottom corner when the top corner is invisible.  Jumping the mouse doesn't seem as confusing as jumping the region.

It might be more intuitive, but require a deeper understanding, to bring up the region over the window without moving the mouse to a corner.  Then have the region track the mouse's relative movement even though the cursor is never aligned with a corner.  That's probably more like the way modern interfaces behave (the window image itself tracks the mouse, we inherited this ghost thing presumably because of performance issues on slow machines).

PR #1295, opened 2023-07-29
NEARESTCORNER must be onscreen  (addresses #1294

rmkaplan:
Mouse jumps to the nearest onscreen corner of the ghost region


masinter (commented):
I think there's a problem with these merge commits, because they show no explanation when you go to view the PR on github.
So it's hard to see this from GitHub, maybe PRC shows it?
What I was wondering was: is it possible for a window to be completely off screen? What about if you resize the screen (e.g., log into online.interlisp.org from a phone)?  

Perhaps we need a separate issue for dealing with resizing without logout.

lispusers/MODERNIZE
```diff
@@ -325,21 +325,46 @@
...
+    (LET ((LEFT (FETCH (REGION LEFT) OF REGION))
+          (RIGHT (FETCH (REGION RIGHT) OF REGION))
+          (TOP (FETCH (REGION TOP) OF REGION))
+          (BOTTOM (FETCH (REGION BOTTOM) OF REGION))
+          X Y)
+
+         (* ;; "If the nearest corner is offscreen, pick the other one.")
+
+         (SETQ X (if (OR (ILESSP LEFT 0)
+                         (IGREATERP LEFT SCREENWIDTH))
```
masinter, line 342: is there an off-by-one here?  shouldn't LEFT > SCREENWIDTH be LEFT >= SCREENWIDTH (or SCREENWIDTH > LEFT)?

rmkaplan:
I fixed the off-by-one bug, thanks.

masinter (approved):
i was running with this before and after the off-by-1 and the edits look reasonable and it worked fine. So...  LGTM

Issue #1299, opened 2023-07-30
\SUBREAD doesn't check the EOL convention


rmkaplan:
This may have been something I missed when I did the external format conversion.

The function \SUBREAD (used by READ, RATOM etc.) gets its characters by calling \INCCODE.  That decodes the various external formats into character codes (interprets 255's for XCCS, high-order bit for UTF-8 etc.) but it doesn't coerce LF CR-LF to the canonical (CHARCODE EOL)=13.

The entry \INCCODE.EOLC decodes the format and canonicalizes the end-of-line.

This difference has gone unnoticed for quite awhile, mostly because LF and CR are separator characters in most readtables that would be used by these Lisp code-reading functions.

But it turns out that the Tedit Bravo converter, ancient code (LPD initials in some places) and funky in many respects, uses a readtable in which 13 is set as a break-character that it expects RATOM to return as a single-characer atom.

Is there a reason not to have RATOM and READ do the EOL interpretation, or was this just an oversight on my part?

masinter:
fixed by #1300

PR #1300, opened 2023-07-31
LLREAD uses \INCCODE.EOLC instead of \INCCODE to coerce LF and CRLF to EOL(=CR)

rmkaplan:


rmkaplan:
(addresses #1299

rmkaplan:
Test by setting the EOL convention of a file to, say, LF, setting the LF character to something other than SEPR, and reading something that includes the LF.  The LF should show up as (CHARCODE EOL) = 13 instead of 10.

Shouldn't have a material affect on performance, it should add 2 more EQ tests on most characters, a little bit more on LF and CR.  But it could be tested.

Issue #1301, opened 2023-07-31
EQUALALL should test for the equivalence of bitmaps


rmkaplan:
using basic equality predicate for bitmaps and bigbitmaps.

That equality may also be useful for testing font-equivalence?

PR #1302, opened 2023-07-31
EQUALALL tests equivalence of bitmaps and big bitmaps

rmkaplan:
Address #1301


library/BIGBITMAPS
```diff
@@ -1,15 +1,13 @@
...
-                       \\DSPCREATE.BIGBM)
-                  (MACROS |\\SFInvert|)
+                  (FNS BIGBITMAPEQUAL)
 
-      :PREVIOUS-DATE "26-Oct-2021 14:51:38" 
-|{DSK}<users>kaplan>local>medley3.5>working-medley>library>BIGBITMAPS.;6|)
+      :PREVIOUS-DATE " 9-Jul-2022 09:41:26" |{WMEDLEY}<library>BIGBITMAPS.;12|)
 
 
 ; Copyright (c) 1991, 1993-1994 by Venue.
```
masinter, line 13: please load SYSEDIT when editing system files. It among other things resets the COPYRIGHTFLG.

masinter (approved):
I wish there were a test case -- just a simple "make three  big bitmaps, the second a copy of the first and the third something different, and test that the first two are equalall and the third is different.

rmkaplan:
In SYSEDIT:

Apart from the copyright, what do the following do:

GLOBALVARFLG
CROSSCOMPILING
MSRECORDTRANFLG

IRM and man don’t appear to know about these.

PR #1304, opened 2023-08-03
Revamp loadup scripts and release-medley script; sync with github workflow buildLoadup.yml

fghalasz:
This PR does the following:

1. Rewrite the loadup-* scripts to make them more consistent across the various scripts.
2. Change the work directory scheme so that loadup does it work in the value of the env variable LOADUP_WORKDIR.  The default for LOADUP_WORKDIR is /tmp/loadup-$$, but can be changed by setting LOADUP_WORKDIR to, e.g., ./tmp, which was the previous scheme.
3. Fixed up all lisp file functions that hardcoded loadup work directories to ./tmp.  Replaced with work dir path based on (UNIX-GETENV 'LOADUP-WORKDIR).
4.  Deleted all of the .CM (and .LISP) files from medley/sources.  These are not really source files, they are each used only by a single script file, and they are short.  So I moved the .CM files into here-docs within the script that used to refer to them.  Much easier to maintain this way.
5. Separated loadup-all and loadup-db so they are independent - except to the extent that loadp-db depends on the full.sysout that loadup-all produces.
6. RDSYS.* is now copied (hardlinked) loadups.
7. Separated the creation of the medley release tars into a separate script.  This script is then used by both the release-medley.sh script and the buildLoadup.yml github action.  This ensures there is only one place that holds the definitive manifest about whats included in release tars. 
8.  Updated the cpv script so that it used hardlinks or copies depending on whether the source and destination are on the same filesystem or not.  Takes care of issues if /tmp is not on same filesystem as medley.
9.  Updated cpv script to delete $dest file before hardlinking to $dest.~$new~.  This eliminates issues where some file were hardlinked to 4 or more other "files".

All scripts continue to be run by /bin/sh - which on Ubuntu is dash.  I tested all scripts on both Ubuntu and MacOs.  I did not try on BSD-ish OSes - but I tried to pay attention to BSD requirements (e.g., doesn't use GNU tar - nor does MacOs). 

Tested on Unbuntu 22.04 and MacOs 12.6 (Monterey).


nbriggs:
So far so good - I've built a release on FreeBSD with no trouble related to these changes (there are issues if you use the scroll-wheel or trackpad scroll on the Lisp window at the "wrong" time during the loadup)

I will look at what the release tar files contain next.

nbriggs (changes requested):
If the loadups tar file doesn't include the starter.sysout then you can't make a new loadup with the scripts/loadup-full.sh (or other loadup scripts).

If you put a copy of loadups/lisp.sysout as loadups/starter.sysout you can get started, but when I was doing this from /tmp/medleyfromtar/medley, using ./scripts/loadup-full.sh, it started loading source files from ~/medley/sources instead of from the just installed release.  init.dribble attached (as init-dribble.txt)
[init-dribble.txt](https://github.com/Interlisp/medley/files/12256209/init-dribble.txt)

nbriggs:
After putting the correct starter.sysout into loadups, doing a "./scripts/loadup-full.sh", it fails in loadup-lisp-from-mid - dribble attached.
```
{DSK}<tmp>medleyfromtar>medley>library>UNICODE.LCOM;1
compiled on 19-Jul-2023 09:26:13
File created 19-Jul-2023 09:26:13
UNICODECOMS
In PATHNAME:
The value of THING, NIL, is not a PATHNAME or a STREAM or a STRING or a (AND SYMBOL (NOT NULL)).
```

[lisp.dribble.txt](https://github.com/Interlisp/medley/files/12256530/lisp.dribble.txt)

rmkaplan:
I made a change a few weeks ago in UNICODE to put in a (FILES (FROM LOADUPS) EXPORTS.ALL, the new convention.

But that’s under an ***@***.*** DONTCOPY, so it shouldn’t affect loading the LCOM.  And I haven’t had trouble with loadups.

nbriggs:
@rmkaplan - it doesn't happen when I do a loadup in the checked out git repo, but it fails when I use the tar files to make a medley installation directory and try to do a loadup there.  I suspect there's a missing file in the tar.

nbriggs:
```
(debug)btv!
Basic frame at   27060
  27056:       0      0 *local* NIL
  27060:  100000  27056 

Frame xtn at   27062, frame name= PATHNAME
  27062:  140004  26744 [USE= 4,  alink]
  27064:     172   3250 [fn header]
  27066:   27112    152 [next, pc]
  27070:  140000  26744 [nametable]
  27072:      62  72274 [blink, clink]
  27074:       0      0 "local" NIL
  27076:       0      0 "local" NIL
  27100:       0      0 "local" NIL
  27102:  177777 177777 [padding]
  27104:       0   4751 [padding]
  27106:       0      0 [padding]
  27110:  177774      4 
PATHNAME


Basic frame at   26730
  26716:       0      0 *local* NIL
  26720:       0  24563 *local* :DIRECTION
  26722:       0  24556 *local* :INPUT
  26724:       0   2332 *local* :EXTERNAL-FORMAT
  26726:     164  71574 *local* :UTF-8-RAW
  26730:  100000  26716 

Frame xtn at   26732, frame name= OPEN
  26732:  140001  26672 [USE= 1,  alink]
  26734:      66  63024 [fn header]
  26736:   27056    376 [next, pc]
  26740:  177777 177777 [nametable]
  26742:  177777 177777 [blink, clink]
  26744:      16      5 "local" 5
  26746:       0      0 "local" NIL
  26750:      16      5 "local" 5
  26752:      16      2 "local" 2
  26754:       0  24556 "local" :INPUT
  26756:      16      5 "local" 5
  26760:      16      6 "local" 6
  26762:       0   4751 "local" CL:STRING-CHAR
  26764:       0      0 "local" NIL
  26766:      16      5 "local" 5
  26770:      16      6 "local" 6
  26772:       0      0 "local" NIL
  26774:       0      0 "local" NIL
  26776:      16      5 "local" 5
  27000:      16      6 "local" 6
  27002:       0      0 "local" NIL
  27004:      16      5 "local" 5
  27006:      16      4 "local" 4
  27010:     164  71574 "local" :UTF-8-RAW
  27012:       0      0 "local" NIL
  27014:       0      0 "local" NIL
  27016:       0      0 "local" NIL
  27020:       0      0 "local" NIL
  27022:       0      0 "local" NIL
  27024:       0      0 "local" NIL
  27026:  177777 177777 local [unbound]
  27030:  177777 177777 local [unbound]
  27032:  177777 177777 local [unbound]
  27034:  177777 177777 local [unbound]
  27036:  177777 177777 local [unbound]
  27040:  177777 177777 local [unbound]
  27042:  177777 177777 [fvar *QUERY-IO*  not looked up]
  27044:     170 126276 [padding]
  27046:     164  71574 [padding]
  27050:  177776      0 
  27052:  177757     40 
  27054:  177767     60 
OPEN


Basic frame at   26656
  26650:     164 165500 FILESPEC (LATIN SYMBOLS1 SYMBOLS2 EXTENDED-LATIN FORMS SYMBOLS3 SYMBOLS4 ACCENTED-LATIN GREEK)
  26652:       0    114 NOPRINT T
  26654:       0      0 NOERROR NIL
  26656:  100000  26650 

Frame xtn at   26660, frame name= READ-UNICODE-MAPPING
  26660:  140001  26624 [USE= 1,  alink]
  26662:     164  65620 [fn header]
  26664:   26716    207 [next, pc]
  26666:     164 167320 [nametable]
  26670:     164 167320 [blink, clink]
  26672:     165  36632 "local" (NIL NIL NIL NIL NIL NIL NIL NIL NIL)
  26674:     166  57202 SEPBITTABLE {CHARTABLE}#166,57202
  26676:       0      0 "local" NIL
  26700:       0      0 FILE NIL
  26702:  177777 177777 STREAM [unbound]
  26704:  177777 177777 A0123 [unbound]
  26706:  177777 177777 [padding]
  26710:       0    114 [padding]
  26712:  177773      6 
  26714:       0      0 NIL
READ-UNICODE-MAPPING


Basic frame at   26610
  26604:     164 167324 *FORM* (READ-UNICODE-MAPPING DEFAULT-XCCS-CHARSETS T)
  26606:     164  71310 *local* READ-UNICODE-MAPPING
  26610:  100000  26604 

```

rmkaplan:
Looks like it’s trying to build the mapping tables for the various XCCS character sets, so presumably trying to find the files in the unicode/ subdirectory?

nbriggs:
Yes, the unicode directory is missing from the release tar files - I presume it should be there?

rmkaplan:
Yes, otherwise UTF8 doesn’t work, given that our fonts are XCCS.

fghalasz:
@nbriggs  It has never (at least not in the last year or so) been possible to do loadups based on the release tars since neither starter.sysout nor the /unicode directory have been included in the release tars.

Loadups have always been done in .git cloned directories or copies of git cloned directories.

Looks like /unicode should be added to the release tars no matter what.

starter.sysout can be added if we want to do loadups from release tars - at the cost of making the release tars and all installers a bit heavier.

Opinions?

nbriggs:
One should be able to do a new loadup starting from almost any old sysout - starter.sysout is just one from Venue that should work (it was a lisp.sysout that Larry copied, I think).  The fact that starting from the distributed (in the tar files) lisp.sysout loads files not from the new installation probably means that the initialization is baking in too much information - I don't know if that's because something wasn't undoable, or some other reason.  I don't see how it would have settled on ~/medley/sources when the lisp.sysout has never seen my machine before and that's not where the scripts are being run from.  It's the first step, scripts/loadup-init.sh, where you can tell whether it's going to work or not.

masinter:
the point of having a starter sysout was to add a bit of stability. It wasn't from venue -- I think we wanted the UTF-8 and EOL convention things fixed. 

Don't change it without reason. starter.sysout should be checked in, so doesn't belont in loadups.
I think the *.venuesysout belong in obsolete:

nbriggs:
```
% git log loadups/starter.sysout 
commit 789747112666823af48f91e9d810c34318486909
Author: Larry Masinter <lmm@acm.org>
Date:   Thu Aug 26 15:47:59 2021 -0700

    no longer commit loadups; start with starter.sysout (#430)
    
    * Start loadup using lisp.venuesysout as the base on which to build new   
    * Change 'starter.sysout' (taken from recent lisp.sysout) 
    * Fix problem with GREET ordering
```
You *should* be able to do loadups starting from the distributed lisp.sysout - the fact that it doesn't work is concerning.

fghalasz:
I agree with @masinter that always starting a loadup from a known base (in this case starter.sysout) is the right way to go.  It makes the loadup process replicable.  If we start with any lisp.sysout, we are opening ourselves up to a much harder problem debugging issues with loadups.

On the issue of where starter.sysout should go.  This is just one pat of a larger issue that the loadup-specific "input" files are spread out in the various medley subdirs in git.  LOADUP-XXX(.LCOM) and MAKEINIT are in sources/.  MEDLEY-UTILS is in internals/.  starter.sysout and the README.md (out-of-date) are in loadups/.   (I got rid of the .CM files, but they were in sources/).

I suggest we pull all of the loadup-specific input files into one common subdir - separate from loadups/, which would be reserved for the "outputs" from the loadup processes (and hence not included in git).

My suggestion is to move the inputs into `internal/loadups/`.

So the plan would be as follows:
1. Move lisp.venuesysout to obsolete/.
2. Move starter.sysout and README.md to internal/loadups/
3. Move the LOADUP-XXX(.LCOM) files to internal/loadups/
4. Move MAKEINIT(.LCOM) to internal/loadups/
5.  Leave MEDLEY-UTILS in internal/ since its major functions (e.g. MAKE-WHEREIS-HASH) could be called outside of loadups - although none of them currently are.
6.  Move internal/MAKE-EXPORTS-ALL to obsolete/ - not exactly part of this but ....

Since all of internal/ is included in the release tars, these changes would make it possible to do a loadup using a medley directory created using release tars.

nbriggs:
Yes, we should start our released lisp.sysout and full.sysout from a known base -- as the logs pointed out, our starter.sysout is now a lisp.sysout that we made, not the Venue one.  The fact that NEW lisp.sysouts that we make for the release tars are NOT capable of being used in place of the one that's now starter.sysout is a problem that we need to address.  That problem has nothing to do with the organization of the release files.

But - I don't think moving the Venue-created sysouts to obsolete is really the right thing to do.  Internal, perhaps, but not obsolete.

fghalasz:
@nbriggs.  Why don't you open  separate issue to cover starting loadups from an arbitrary lisp,sysout or full.sysout.  Implementing this will involve changes to the lisp side of the loadup process.  This is not something that was a goal of this PR.

The goal of this PR is to clean-up the loadup and release scripts so as to ensure a clean separation between loadup-all and loadup-db and to ensure that the loadup and release tar processes were encoded only once for use by both local builds and the github actions (e.g., so as to prevent asynchronies between release tars made locally versus by github actions).   I did not modify the loadup process and did not modify any Lisp code except as necessary to fix up the file/directory specification between the scripts and the Lisp code. I see the file reorganization as part of cleaning up the scripts.

Changing the loadup process from the Lisp side will a bit gnarly if you ask me and I don't want to take that on at this point since its not necessary for the automated builds and related topics.

nbriggs:
It's issue #1306

fghalasz:
Just added changes to move loadup-related source files to internal/loadups and updated all scripts accordingly.

Also moved venusesysteouts from loadups to internal/venuesysouts.

Also fixed issues with starting loadups from any lisp.sysout or full.sysout (instead of just starter.sysout) (see PR#1306)

fghalasz:
Ok.  Just added the following:
1.  Fixed the issue @nbriggs reported this morning re: missing starter.sysout (and other files)
2.  Removed venuesysouts from release tars.
3.  Added in changes from @masinter's PR#1307 (by hand, not by merge)

If you already pulled by latest commits from this weekend, you will need to do a `git pull origin fgh_loadups2 --rebase`.  If you didnt't pull the changes this weekend, you can do the standard pull.

nbriggs:
Or:
```
   git remote update origin # make sure that your git origin is up to date
   git checkout master # make sure you're on master (or anywhere NOT fgh_loadsup2)
   git branch -D fgh_loadups2  # delete the old local copy of fgh_loadups2 branch
   git checkout fgh_loadups2   # get the updated fgh_loadups2
```

nbriggs:
Well... it's not happy on my machine:
```
medley % ./scripts/loadup-all.sh
>>>>> START loadup-init
MEDLEYDIR has no loadups: /Users/briggs/Projects/medley
----- FAILURE -----
..... files created .....
ls: /tmp/loadups-36299/init.*: No such file or directory
ls: /tmp/loadups-36299/RDSYS*: No such file or directory
ls: /tmp/loadups-36299/I-NEW*: No such file or directory
<<<<< END loadup-init

----- loadup-all.sh: FAILURE -----
```

nbriggs:
... Because the "medley/loadups" directory is not created by the setup.

To test this from a clean checkout, assuming you have no unsaved files in the medley directory, do a
```
git clean -dxn   # check what it's going to delete!
git clean -dxf.   # delete everything in the directory that is not checked into git
```

nbriggs:
I've pushed the commit with the change to mkdir the loadups output directory.

nbriggs:
What I'm seeing occasionally is a "FAILURE" report - but I think it's really because the step to create the init-mid.sysout is so fast that the timestamp check with loadup.timestamp is failing: here's an example (I fixed the script to do a stat on the files after it has done the `-nt` check between the files)
```
----- FAILURE -----SSNSSSsions:
  File: "/tmp/loadups-41465/init-mid.sysout"
  Size: 4975616      FileType: Regular File
  Mode: (0644/-rw-r--r--)         Uid: (  501/  briggs)  Gid: (    0/   wheel)
Device: 1,4   Inode: 45083793    Links: 1
Access: Mon Aug  7 18:37:02 2023
Modify: Mon Aug  7 18:37:02 2023
Change: Mon Aug  7 18:37:02 2023
  File: "/tmp/loadups-41465/loadup.timestamp"
  Size: 0            FileType: Regular File
  Mode: (0644/-rw-r--r--)         Uid: (  501/  briggs)  Gid: (    0/   wheel)
Device: 1,4   Inode: 45083777    Links: 1
Access: Mon Aug  7 18:37:02 2023
Modify: Mon Aug  7 18:37:02 2023
Change: Mon Aug  7 18:37:02 2023
..... files created .....
<<<<< END loadup-mid-from-init
```
usually the sysout is a second newer...

fghalasz:
@nbriggs  It is possible that the loadup-mid-from-init failures are a timestamp race condition - but seems unlikely since file timestamps are kept in milliseconds.  Nevertheless I added a 2-second delay after creating the timestamp before calling run-medley.   This should avoid any possible race conditions.

Note that load-up-mid-from-init will fail if there is any keyboard input during the brief time that Medley is running.  Its a bit hard to catch.  But if you continually tap the keyboard during the script, you will catch the failure 8 times out of 10..

Mouse movements do not seem to cause the failure.

masinter:
i thought I fixed the problem with < 1 second steps by using product timestamp greter-than or EQUAL TO start timestamp. Did something change?

nbriggs:
I don't know what it was doing, but it is currently checking using the `-nt` (newer than) operator between the sysout and the timestamp file.  In the failure cases I saw, the times were the same to the second, though there are higher precision timestamps kept in the file system, and they should be available through the `stat()` call, so I'm not sure what exactly is going wrong here.

masinter:
it looks like loadup-mid-from-init.sh is the only one that has it right (as the only step short enough to finish in under a second).  -nt tests "newer than" so 
    touch timestamp
    ... make product ...
    if  timestamp  newer than  product  then  fail

nbriggs:
@masinter I don't understand that - given that the `touch` and `<product> -nt loadup.timestamp` are in the loadup-setup.sh it should be common to all the scripts - the setup, loadup_start, and loadup_finish are called from every script that directly produces output.

fghalasz:
Based on my googling, it looks like the -nt resolution (i..e., subsecond or not) varies across scripts and platforms.  So we have to assume second resolution.

The is no "newer or equal" operator - so you would have to do something like [[  $f1 -nt $f2 ||  ! $f1 -ot $f2 ]].  Nothing like that existed in the current scripts - just the standard -nt comparison.

At any rate the 2 second delay I put in between the timestamp and the run-medley will make sure that this equal creation time issue moot.

masinter:
```

larry@TV-PC:~/il/medley$ grep nt scripts/loadup*.sh
scripts/loadup-apps-from-full.sh:if [ ${SYSOUTDIR}/apps.sysout -nt ${SYSOUTDIR}/loadup.timestamp ]; then
scripts/loadup-aux.sh:if [ tmp/whereis.hash -nt tmp/loadup.timestamp ]; then
scripts/loadup-db.sh:if [ tmp/fuller.database -nt tmp/db.timestamp ]; then
scripts/loadup-full-from-lisp.sh:if [ tmp/full.sysout -nt tmp/loadup.timestamp ]; then
scripts/loadup-init.sh:if [ tmp/init.dlinit -nt tmp/loadup.timestamp ]; then
scripts/loadup-lisp-from-mid.sh:if [ tmp/lisp.sysout -nt tmp/loadup.timestamp ]; then
scripts/loadup-mid-from-init.sh:if [ tmp/loadup.timestamp -nt tmp/init-mid.sysout ]; then
larry@TV-PC:~/il/medley$
```

they are all wrong except mid-from-init; note the order of  timestamp -nt sysout => fail

but frank's patch wil work too

nbriggs:
@masinter what branch were you on?

masinter:
I was on master

fghalasz:
Sorry @masinter, I missed the subtlety of the difference in the -nt clause in mid-to-lisp versus the other scripts.  I was trying to regularize the scripts so that the common portions were just implemented once (in loadup-setup.sh).  I missed that mid-to-lisp was different on purpose.

But I think the delay for 2 seconds solution will work ok and keep to the goal that all scripts have the same common components.

masinter:
mid-to-lisp was right and all the others were wrong -- but mid-to-lisp was the only step that only took less than a second. I don't know why I didn't fix the others -- laziness?

fghalasz:
@masinter After some reflection, I agree that your original fix for loadup-mid-to-lisp was the right way to go.  So I fixed up loadup-setup.sh to use "loadup-timestamp -nt <key outp[ut file>" for all of the loadup scripts.  (and I removed the 2 second delay from loadup-mid-to-lisp).

masinter (approved):
all the rework looks good to me. Let's get this show on the road.

PR #1307, opened 2023-08-07
makeinit ok to modify

masinter:
- Remake loadup scripts to not use MEDLEYDIR/tmp as a working dir AND cleanly separate loadup-all from loadup-db; adjust buildLoadup.yml accordingly
- Add new files left out from last commit; add RDSYS to library as part of loadups-all
- WIP updating loadups scripts
- First full pass at loadups revamp.
- Handle case where files already exist in LOADUP_WORKDIR - show only the newly created files - not the old ones
- Remove now obsolete sources/*.CM files; update README a bit
- WIP fixing up scripts/release_medley.sh and syncing with buildLoadup.yml
- First pass a complete new make-release.sh
- Fix scripts/README.md
- update buildLOadup.yml to use new scripts/release-make-tars.sh instead of its almost but not quite equivalent make tars code
- Remake loadup scripts to not use MEDLEYDIR/tmp as a working dir AND cleanly separate loadup-all from loadup-db; adjust buildLoadup.yml accordingly
- Add new files left out from last commit; add RDSYS to library as part of loadups-all
- WIP updating loadups scripts
- First full pass at loadups revamp.
- Handle case where files already exist in LOADUP_WORKDIR - show only the newly created files - not the old ones
- Remove now obsolete sources/*.CM files; update README a bit
- WIP fixing up scripts/release_medley.sh and syncing with buildLoadup.yml
- First pass a complete new make-release.sh
- Fix scripts/README.md
- update buildLOadup.yml to use new scripts/release-make-tars.sh instead of its almost but not quite equivalent make tars code
- Remove references to LOADUP_WORKDIR from all lisp code and instead pass the file references down from the callinf .CM files (which are now embedded in the loadup- scripts;  also add /unicode subdir to the release tars.
- make sure unsafe-to-modify doesn't interfere


fghalasz:
I added these changes by hand to fgh_loadups2 branch (and PR#1304).  

Because the file MAKEINIT moved from sources/ to internal/loadups/, it was to hard to just merge the changes into the latest HEAD of fgh_loadups2.

PR #1308, opened 2023-08-15
Fix issue with tmp/logindir remaining after loadup-all and loadup-db scripts

fghalasz:
This PR deletes the ${MEDLEYDIR}/tmp/logindir subdir at the end of all loadup- scripts - providing that the subdir did not exist before the scripts was run.

The tmp/logindir is actually created by run-medley when it is called with the -loadup and -nogreet options.  But rather than mess with run-medley, this PR just adds the code to the loadup- scripots via loadup-setup.sh.

### DIFF
--- a/scripts/loadup-setup.sh
+++ b/scripts/loadup-setup.sh
@@ -49,6 +49,19 @@ cmfile="${LOADUP_WORKDIR}/${script_name}.cm"
 
 loadup_start () {
   echo ">>>>> START ${script_name}"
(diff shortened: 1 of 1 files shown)


fghalasz:
I agree that the right thing to do is too fix this in run-medley.  But I didn't want to mess with something as central as run-medley right now (especially since I will be out for two weeks).

I will open an issue to fix this in run-medley and assign it to myself to fix soon-ish.  At that point I will remove the code added to loadup-setup in this PR.

PR #1324, opened 2023-09-16
Fix Issue#1323: buildDocker.yml Github Action is failing due to Ubuntu 22.10 being deprecated

fghalasz:
This PR moves the docker build back to Ubuntu 22.04, which is an LTS release.

Also "fixes" an issue with how TZDATA package was loaded into Ubuntu.  Not sure that it was absolutely necessary to do this.  But it works, so ...

Finally, fixed a slew of trailing whitespace issues for buildDocker.yml.  A completely cosmetic change that makes it easier to edit in editors that see trailing whitespace as an error.


masinter (approved):
LGTM

PR #1326, opened 2023-09-19
Fix latest issues with docker build (and Release+Docker builds) for Medley

fghalasz:
Fixed issues causing buildReleaseInclDocker github action to fail.

Specifically:

1. Fixed issue with passing of GITHUB_TOKEN secret from buildReleaseInclDocker workflow to sub-workflow buildRelease.

2. Fixed issues in buildRelease workflow with Medley deb file naming convention, which was changed recently with a merge into master (without updating these two file - oops).


masinter (approved):
trusting FGH

PR #1329, opened 2023-09-24
COMPAREDIRECTORIES, COMPARESOURCES, COMPARETEXT, EXAMINEDEFS

rmkaplan:
Relatively minor cleanups, little or no functionality improvements.

These have been hanging around in my working directory for quite awhile, not worth pushing, but I want to get resynchronized.


masinter (commented):
1) more words about what changed and why would be helpful.
2) How did / should we (regression) test this?
2) Should these move to "library"? Worth a discussion?
I want to at least test before approving.

masinter (commented):
Does this require REGIONMANAGER changes?
When I run this in a newly made sysoout, it no longer automatically chooses regions for showing the two files it's comparing, but instead asks me to drag them out.

masinter (approved):
I noted that CDBROWSER no longer seems to lay out the two display files.

But that's likely a different problem.

rmkaplan:
I made a new loadup in master after the merge, and I did see two display files in the CDBROWSER under prc (the See command).  When did it misbehave?

rmkaplan:
I take it back, I do see the problem.  I'll take a look.  Maybe does depend on REGIONMANAGER  changes (I don't remember now), but I think those were already merged.

rmkaplan:
The problem is an interaction of newer REGIONMANAGER with older TEDIT.  But it is a little deeper, in that REGIONMANAGER tests for REGIONP in various ways. But there is another kind of region, a SCREENREGION, that is also being passed around through the various manipulation functions.  A SCREENREGION has a SCREEN in addition to a region, so fails the REGIONP test and the REGION-TYPE gets confused. 

I can put out an easy fix, that simply takes screen regions out of any of this management stuff (but it might help to add a SCREENREGIONP predicate in WINDOW.)  A previously used screenregion for an application/menu would never be offered as a default, the user would always have to sweep out a new one.

But I'm not sure how to think about this--screen regions came after my time, not sure what they are supposed to do.  For REGIONMANAGER in particular, does it make sense for screen regions to be managed in the same way as other typed regions (recycled by this generic mechanism when a window built in that region is closed and the region can be recycled for that application)?  If so, is it just the region part that should be recyclable, or should it be the region paired with the screen?

Or should they just be excluded  from any managed behavior.

rmkaplan:
I put out another PR (rmk-89) with a fix for the side-by-side-windows glitch you noticed.

PR #1331, opened 2023-09-24
CMLARAITH:  Makefile new to get fns/functions in filemap

rmkaplan:
Just a MAKEFILE new--one of the several (still) carryover files that didn't include commonlisp functions in the filemap, so PF didn't work.


rmkaplan:
Actually, I have no idea how CLEANUP and/or the compiler is supposed to work on files that are marked CL:COMPILE-FILE and for which there is an LCOM.  And, given that somehow a DFASL might have been produced and is newer than the LCOM, it seems that the loader still goes for the LCOM.

So, I'm not sure that the compiled files are good in this PR, they should be tested.

rmkaplan:
I created an issue #1335 about the DFASL/LCOM problem.

MattHeffron (approved):
This looks reasonable. 
HOWEVER, I tried to test this by (LOAD _xx_ 'SYSLOAD) of CMLARITH.LCOM and CMLCHARACTER.DFASL, but that didn't work.
I'm assuming that these files must be in place when the release is built.
I'm approving this on that assumption, or that it'll be no worse than it is now if it doesn't work.

rmkaplan:
I tried doing a loadup after the merge, and it failed, I think because of confusion about how to compile files early in the loadup.  The FILETYPE property doesn’t seem to do the “right” thing for files with FUNCTIONS that have to be compiled as LCOMs.  One of these files now has a DFASL, and I think that breaks the load.

I will try to recompile and put out another commit (maybe in another PR).  In the meantime, maybe best to back out this merge.

rmkaplan:
I confirmed that the problem is the DFASL for CMLCHARACTER.  If I nuke that and restore the old LCOM, the loadup works.

But:  I don’t know how to compile CMLCHARACTER into a new LCOM.  FAKE-COMPILE-FILE and TCOMPL both break with an error
"In CL:LENGTH:  The value of SEQUENCE, Newline, is not a LIST or a VECTOR”.

CL:COMPILE-FILE compiles it OK…but the DFASL. can’t be used in making a loadup.

So I think the immediate thing is to get rid of the DFASL and restore the preexisting LCOM, even though that is out of step with the symbolic file that was changed only to provide a useful filemap for pf/tf.

nbriggs:
The problem seems to stem from some Newline processing that you can see in a regular full.sysout - try doing `(CHARCODE #\Newline)` and then try to `fix` that line.  You'll get the CL:LENGTH error.

Issue #1322, opened 2023-09-15
\UFSCloseFile will attempt to close an already closed file


nbriggs:
There are two problems, fixing either of them will solve this -

1) `\UFSCloseFile` tests for already-closed by whether or not the `(UFSSTREAM UNIXNAME)` field is `NIL`, but fails to set the `UNIXNAME` to `NIL` once the stream has been closed by the SUBR that does the OS related work.
2) A more correct way to test for a closed stream is whether the `(STREAM ACCESS)` field is `NIL`, as done in `\GETSTREAM`

I propose changing the code to test for either `(STREAM ACCESS)` OR the `UNIXNAME` being `NIL` (implies already closed), and also setting the UNIXNAME to NIL after the file has been closed by the SUBR.

There's another minor design flaw - the underlying subr code doesn't distinguish between being passed `NIL` and `0` for the file descriptor - had it done that it would have been much more obvious that there was a problem, compared to the current behavior of just closing `stdin` (fd 0).

PR #1334, opened 2023-09-26
Nhb 1322 closing ufs files

rmkaplan:
This is Nicks fix to prevent \UFSCloseFile from attempting to close an already closed file #1322 .  I'm submitting this at is recommendation--he's essentially offline now.


masinter (approved):
Works fime per RMK

PR #1335, opened 2023-09-27
FILEIO:  OPENSTREAM parameters default to STREAMPROP

rmkaplan:
An extension to OPENSTREAM for convenience and completeness:  a property that doesn't have a special interpretation in \DO.PARAMS.AT.OPEN falls through to the generic \PUTSTREAMPROP.  

Also, again for convenience, adds LINELENGTH as a property that can be specified at open time.


masinter:
I feel some reluctance to add features "for convenience" if they aren't required to fix a bug, or compatibility with Common Lisp modules we might like to run.

If we're going to go down this road -- are there any requirements or situations where software that expects the added behavior will not function properly? How much more does it add "convenience" for whom?  New Interlisp programmers?

Let's just have a clear agreed-upon policy.

PR #1341, opened 2023-10-08
Add ShellOpen (and ShellOpener) to UNIXUTILS; fix small bug in UNIX-FILE-NAME

fghalasz:
@rmkaplan  NOTE:  This PR is meant to be merged (cleanly I hope) into the PDFSTREAM branch instead of directly into Master.  It is based on the latest PDFSTREAM branch on github.  This way we hopefully won't run into conflicts when trying to merge PDFSTREAM that includes this branch into Master.

This PR adds the functions ShellOpen and ShellOpener - which given a filename will attempt to open the file via a shell command to whatever opener is available on the current machine.  For MacOS this will be "open"; for WSL its "wslview"; for Cygwin-Windows its cygstart; for straight Linux its xdg-open.

ShellOpen takes a single argument - filename or URL (starts with http://, https://, file:///).

ShellBrowse and ShellBrowser have been modified to use the ShellOpen(er) function unless there is no known opener on this machine, in which case ShellBrowser will look for web-specific openers like git web--browse and/or lynx.

Also fixed a bug that caused a break in UNIX-FILE-NAME if the given filename did not exist and ACCESS was 'INPUT and COPY was non-nil.


masinter:
Looks to me like PDFSTREAM has already been merged. Now I get conflicts with the file sources/HARDCOPY when I try to merge Remove-calls-to-OPENFILE and in the file library/PDFSTREAM (and .LCOM) 
and with the file library/PDFSTREAM.

I'm not sure what's going on.

rmkaplan:
It’s not surprising that that happens when there are long delays on interdependent PR’s.

We’ll just have to unscramble it.

masinter (approved):
PDFSTREAM has been merged, so this can be merged onto master. I think.

PR #1344, opened 2023-10-10
GITFNS: preserves casing of pseudohost prefix-paths of medley projects

rmkaplan:
Also, minor edit to GITFNS.TEDIT, but now also includes a Tedit timestamp.

The prefix change makes this more compatible with the UNIX-FILE-NAME in UNIXUTILS and PDFSTREAM


lispusers/GITFNS
```diff
@@ -16,7 +16,7 @@
         (* ;; "Set up")
 
         (FILES (SYSLOAD FROM LISPUSERS)
```
masinter, line 18: I've taken to changing these to just (FILES file1 file2 ..) rather than (FILES (SYSLOAD FROM wherever) fiile1 file2 ..) with the idea that LDFLG is inherited and DIRECTORIES should be set up. Let's discuss.s

masinter (approved):
Where is the TEdit timestamp feature documented?

The first line of the document still says it was last edited in February.

rmkaplan:
This is a feature that GITFNS inherits from the new TEDIT and COMPAREDIRECTORIES (faked until various approvals and merges).  So GITFNS documentation does not particularly need to be changed.  Really, only the new TEDIT.FILEDATE function should be documented in Chapter 8 of the Tedit documentation—we haven’t yet worked out how to update and propagate basic system documentation.

I don’t think it is wise to let LDFLG be inherited when the intention really is SYSLOAD.

PR #1346, opened 2023-10-11
Manager (Lispusers) grow anchor, icon, and fix typo.

MattHeffron:
Manual cleanup of multiple "Edited" comments in 4 FNS.
Reverted: Changed to CL compiler by default (not need to go to submenu). This is my preference, so I should remove it before setting pull request. (From commit f60c6362)
Fix typo in MasterScope functions (multiple occurrences: LOADBFLG should be LOADDBFLG). (From commit f60c6362)


scripts/medley/medley.cmd
```diff
@@ -1,3 +1,3 @@
-@echo off
-powershell medley.ps1 %*
-
```
masinter, line 3: It's hard to see what changed here -- the GitHub compare shows no difference (The red 'before' and the green 'after' look identical), and the Lisp prc gives an error. And what changed doesn't seem to be obvious from the commit comments. ss

lispusers/MANAGER
```diff
@@ -825,50 +870,50 @@ Will load DataBaseFNS if necessary" (SUBITEMS ("All" 'DB
                                              (FILESLOAD 'DATABASEFNS)
                                              (CL:FORMAT T 
                                        "~&Global DataBaseFNS Flags:~%%SAVEDBFLG = ~a, LOADDBFLG = ~a"
-                                                    SAVEDBFLG LOADBFLG))
+                                                    SAVEDBFLG LOADDBFLG))
                                           (DBFILE 
                                              (FILESLOAD 'DATABASEFNS)
```
masinter, line 875: not sure if this is worth fixing, but FILESLOAD is a NLAMBDA Interlisp function so that (FILESLOAD 'DATABASEFNS) isn't the preferred usage -- you can use
(FILESLOAD DATABASEFNS)

PF* FILESLOAD

masinter (approved):
The commit comments don't reflect the reason some of the changes were made (in in some cases what actually changed isn't clear).

How would you test this (before and after differences in behavior)?

scripts/medley/medley.cmd
MattHeffron: This is just the result of the interim merges from master.

lispusers/MANAGER
MattHeffron: This was not changed. It's not worth changing at this time, as you said.

masinter:
@MattHeffron I approved the merge with the idea that you were free to choose whether to change it or not. So since you submitted and I approved, you do the squash & merge. At some point we should run regression tests automatically but until we have some and a test process -- you check after the merge that loadups work

MattHeffron:
Updated the .TEDIT file and general cleanup (see commit message)

masinter (approved):
if you're working in Windows and also in WSL it seems necessary for everyone to 
```
git config --global core.autocrlf false
```
or some such

PR #1356, opened 2023-10-18
Modify launcher to enable --title STRING option to work when STRING has embedded spaces (multiple words).

MattHeffron:
(Obviously, this is extremely low priority, but it was a little annoying that it didn't work.)


masinter (approved):
LGTM

PR #1359, opened 2023-10-19
Add copy-full.sh to loadup-full.sh script to copy loadup products into MEDLEYDIR/loadups

fghalasz:
As per request from @rmkaplan: Make loadup-full.sh akin to loadup-all.sh in that it copies its products from the tmp directory into MEDLEYDIR/loadups.


rmkaplan (approved):
I tried it both in my clone directory and also in my separate working directory.  Worked well in both places.  Thanks!

PR #1361, opened 2023-10-21
Patch TEDIT.DEACTIVATE.WINDOW in old Tedit

rmkaplan:
Frank noted a glitch introduced a few years ago in the old version of Tedit.  This is a patch, unrelated to new Tedit or anything else


masinter (approved):
LGTM

PR #1371, opened 2023-10-28
Add dribble file to apps-from-full loadup

fghalasz:
As per title.


masinter (approved):
So far I haven't discovered a use for the dribble files that we couldn't find elsewhere. But.. why not?

PR #1374, opened 2023-10-29
COLOROBJ never worked, move to obsolete

rmkaplan:
The image object doesn't have the crucial BUTTONEVENTFN, which was the whole point.

Also, depends on COLOR, which has already been moved to obsolete/.  (I think the Tedit hardcopy failed because COLOR wasn't there.


masinter (approved):
i'm not sure COLOROBJ *never* worked -- but it's pretty clear that getting it to work (again) isn't something we want to tackle. (The fact there's a TEDIT file is evidence that whoever wrote it was really unusual -- writing documentation before getting it to work -- or that it DID work once.

But moving to "obsolete" instead of just deleting it seems prudent enough.

nbriggs:
COLOR, and LLCOLOR and various other color related bits used to work on the Dorado color display, as far as I know.  They may also have worked on the 8-bit Sun displays.

MattHeffron:
Color _did_ work (to some extent) on the HP/UX systems I was using in the late 1990's.
I was surprised when I got color the first time that I ran Interlisp on that system!

Michele31415:
On 10/31/2023 12:34 AM, Nick Briggs wrote:
>
> COLOR, and LLCOLOR and various other color related bits used to work
> on the Dorado color display, as far as I know. They may also have
> worked on the 8-bit Sun displays.
>
They definitely worked on the Sun.  I used that setup at Kodak.  I still
have some sysouts with "COLOR" in their name.

             - Michele

Michele31415:
On 10/31/2023 2:23 AM, Matt Heffron wrote:
>
> Color /did/ work (to some extent) on the HP/UX systems I was using in
> the late 1990's.
> I was surprised when I got color the first time that I ran Interlisp
> on that system!
>
Color also worked on the D0 and Dorado.  My Lispusers Pacman program has
an option for running in color.  I still have a color board on my D0. 
If I can ever get it to boot I'll make a demo.

             - Michele

PR #1381, opened 2023-11-01
LLSYMBOL's FILEMAP was also incomplete

rmkaplan:
This update hopefully won't reveal any other problems.

The symptom that a map is incomplete (doesn't include DEFUNs or DEFMACROS) is that WHEREIS says that the item is on the file, but PF can't find it.  A general solution would be to fix PF so that it doesn't depend on filemaps (speed maybe not so important nowadays), but until then, file by file as they are encountered.


masinter (approved):
isn't it more work to do this one-by-one?

Issue #1303, opened 2023-08-01
PSEUDOHOST documentation missing some information


nbriggs:
The PSEUDOHOSTS.TEDIT documentation doesn't describe whether the prefix path should include or omit a trailing directory separator, and doesn't give any examples where the prefix path is given explicitly.

PR #1385, opened 2023-11-02
PSEUDHOSTS and .TEDIT: Apply TRUEFILENAME on PREFIX, update documenta…

rmkaplan:
…tion

Documentation addresses #1303 .  Using TRUEFILENAME makes sure it always goes to a ground instance in e.g.  (PSEUDOHOST 'MEDLEY '{MEDLEY}).


masinter (approved):
LGTM

PR #1398, opened 2023-11-06
Pull in newer missing library .tedit files, redo PR 1384

masinter:


rmkaplan:
Looking at a sample of these, most of these seem to correspond to the PDF's I have in the directory of old library documents (that I think I originally scanned).  It's great to have editable versions, finally.

The only discrepancy I see is with Color.  This is the only one with a previous Tedit, a new Tedit, and a scanned image.  They are all different, in substantial ways.  It could be that the information is the same, just scrambled around.  The one in this PR may be the most current, since it has the indexing image objects.  In terms of provenance, is the PR version clearly the one to keep?

rmkaplan (approved):
After our discussion in the meeting today, I think the new Color.pdf is more likely to be the best one for current implementation.  It is tagged as "Medley", the previous one was tagged as "Kaleidoscope", in 1989/

So I'm approving this.

PR #1410, opened 2023-11-11
Update ShellOpen to handle versioned files; add ShellOpen into SEE-PDF in place of MacOS-specific open.

fghalasz:
This PR fixes the issue where ShellOpen did not work with versioned files because the .~N~ at the end of versioned files (on the unix filesystem) interfered with the openers' attempt to determine filetype based on the extension.

ShellOpen handles versioned files as follows:

1.  If a version is specified in the given filename, then the file is copied to /tmp/<random#> using the filename  `filename~<version>~.extension`.  Example `foo.pdf;3` becomes `foo~3~.pdf`.   The file is then opened from this tmp file.
2.  If no version is specified but there is no non-versioned file in the unix filesystem, then the highest numbered version will be used and copied as above.
3. If no version is specified and there is a no-versioned file in the unix filesystem, then that no-versioned file is used with no copying to /tmp.

I was not able to use Ron's suggestion of `foo\;3.pdf` since it cause too many issues on several platforms - including in using COPYFILE inside Medley.

Also, could not use symbolic links instead of copying the file to /tmp since symbolic links don't work well with WSL (wslview) and Cygwin (cygstart).

This PR also fixes PDF-SEE to use ShellOpen instead of the MacOS-specific open.

Tested on WSL, Ubuntu Linux, MacOS, and Cygwin.


masinter (approved):
Assumed tested, but works for me. Will merge this and fix up the HELPSYS PR if needed.

PR #1412, opened 2023-11-11
Update CLHS reference to use Interlisp CHLS instead; use UNIXUTILS ShellBrowser

masinter:


rmkaplan:
I reloaded these files into a current sysout and tried “man CONS”.  It didn’t bring up the web page.  Does it a new loadup?

masinter:
(1) the man command only looks in the Common Lisp HyperSpec if it is called on a symbol in the Common Lisp package. For a symbol like "CONS" that is imported into IL package it might not.
(2) The variable CLHS.ROOT.URL is set with an INITVARS, and reloading HELPSYS isn't enough.
if you SET(CLHS.ROOT.URL NOBIND) and then LOAD(HELPSYS.LCOM)

fghalasz (approved):
Tested on WSL, straight Ubuntu 22.04 and on MacOS.  Works as expected on all.  Not tested on cygwin cause can't do loadups on cygwin yet.

Couple of comments:
1.  The following code in CLHS.LOOKUP
Screenshot 2023-11-12 234258

    could be replaced with `(ShellBrowse URL)`

2.  You use MEDLEY-INIT-VARS to reset ShellBrowser and ShellOpener.  This works, but it makes UNIXUTILS unnecessarily reliant on MEDLEYDIR and involves a lot of complexity that goes into MEDLEY-INIT-VARS.  
  
     Why not just add a simple AROUNDEXITFN in UNIXUTILS to reset just these two variables after return to VM?

masinter:
I changed CLHS.LOOKUP to use ShellBrowse.
I didn't make the AROUNDEXITFNS change @fghalasz suggested because it might wind up a net increase in complexity to add a new function while MEDLEYDIR and MEDLEY-INIT-VARS are still there.

PR #1413, opened 2023-11-11
INTERPRESS fixes some of the Asciifont-to-NS character mappings, MATHTONS is obsolete

rmkaplan:
The separate lispusers/MATHTONS is now obsolete.


masinter (approved):
It's a little confusing... what does the list with NILs do that it doesn't do without them?
It would be good to separate out the parts of this that aren't really about INTERPRESS into a library that did character code translations, from PARC-ASCII (with "_" = leftarrow .. for another day?

rmkaplan:
I think the NIL’s are just a shorthand way of specifying \ASCIITONS.  A silly optimization, I left it just in case there is something else going on.

The other confusing thing is to have Classic to Modern in the list.  I think this is because Modern is probably the most populated, so taking Classic over to Modern fills in more Classics. It’s odd then that that would use \ASCIITONS also, but I presumed that that was going to be applied only for non-zero character sets, where the mappings are 1-to-1.  So again, I left it.

PR #1414, opened 2023-11-11
Add a hook to extend FILEBROWSER SEE, PDFSTREAM uses it

rmkaplan:
So SEE of a PDF file should open up a desktop window in a pdf viewer.  (This uses PDFILEP, which looks inside the file, not at the extension.  So it might try to open a file with a wrong extension, and the viewer might fail.)


fghalasz:
When PDFFILEP is T and the filename extension is not .pdf, why not copy it to a file in /tmp with a .pdf extension (as we do in ShellOpen) and then call SEE-PDF on the /tmp file?

rmkaplan:
That’s a good idea.  I think I’ll try to move the filename calculation in ShellOpen into a separate function, so It can be called separately and then get the PDF packed on.

I think it would also be useful to have an inverse function that goes from the Unix target file back to what it came from.  I can then make Tedit report the actual destination of the PDF file instead of the funky tmp file when it does the postcript conversion for hardcopy to pdf.

Since there is also a POSTSCRIPTFILEP (maybe not the best; apparently I wrote it in 93), the filebrowser SEE can be set up to convert the postscript file to PDF and then show it.

masinter (approved):
lgtm

PR #1417, opened 2023-11-13
Recompile files that seemed to have \IS.NO.RANDACCESSP in compiled code

### LINKED ISSUE #1224: (SETFILEPTR T 0) results in UDF \IS.NO.RANDACCESSP
**Describe the bug**
(SETFILEPTR T 0) results in an error (sometimes break window) for \IS.NO.RANDACCESSP being an undefined function.
**To Reproduce**
1. Files loaded: generic full.sysout
2. Form to eval and/or menus to click: (SETFILEPTR T 0)
3. What happened? 

**Expected behavior**
Nothing.

**Screenshots**
image

 - OS: macOS
 - OS Version: Big Sur
 - Host arch: x86_64
 - Maiko version:  (main branch as of Wed 31 May 2023) - 2e893a75abebb57e2d7c790634cf45d3138898d4
 - IL:MAKESYSDATE: [ 14 May 2023 ]

masinter:
grep --text -lri \\is\.no\.randaccessp medley
showed there were a number of .LCOM files that seemed to reference this missspelled function.
I can't quite see how that happens. But this PR, which just has the files recompiled.

For each file where it found a hit, I did

LOAD(file PROP)
BCOMPL(FILE)
F

I didn't carefully isolate these from unfortunate declarations (as we saw with
INTERPRESS) because for the most part it's isolated. I also didn't give in to the temptation to obsolete MOD44IO DSKDISPLAY and other obsolete files, although the temptation was strong.

Will fix #1224


rmkaplan (approved):
Surprised that GITFNS.LCOM is in the list (I don't see that GITFNS references that).  I wonder whether it comes from a macro some where?

But recompiling is presumaby harmless.

PR #1419, opened 2023-11-13
Set LDEKBDTYPE to X in run-medley - if not already set.  Works around issue with loading VIRTUALKEYBOARDS

fghalasz:
As per discussion in Mon meeting a few weeks ago, setting LDEKBD in run-medley top avoid issues with loading VIRTUALKEYBOARDS.  See discussion under PR#1262.


masinter (approved):
LGTM

PR #1420, opened 2023-11-13
Fix minor issue in ShellBrowser - when using the git path was missing the web--browse subcommand.

fghalasz:
Noticed this issue while working on UNIXUTILS over the past week....


rmkaplan:
In ShellOpen, did you intend UNIXFILE to be used freely in the LET* ?

fghalasz:
> In ShellOpen, did you intend UNIXFILE to be used freely in the LET* ?

Not really.  But the compiler kept printing out a message that it was declaring UNIXFILE as SPECIAL.  No idea why since none of the other LET* variables where so treated.  So I just added the DECLARE to shut down the message, figuring if the compiler was going to declare it special then I might as well beat it to the punch.  I meant to get back to figure out what was going on - but never did get around to it.  Please feel free to change as you see fit.

masinter (approved):
LGTM

Issue #1416, opened 2023-11-12
trouble with filename with ! in the name


masinter:
UNPACKFILENAME.STRING chokes on filename with ! in the middle... so 

``./ivy/grapevine/Build.bfs!1_/MesaFont.strike``

causes an error trying to CONCATLIST 
```
("{" DSK "}" ".>ivy>grapevine>Build.bfs!1_" ">" "MesaFont" %. "")
``` 
_(what happened to the 'strike'?)_
which gives an error(!)

masinter:
```
(PACKFILENAME.STRING '(HOST DSK BODY "/A/B/C!1_/D.E"))
```
triggers the error... need to get into the filename mess. The paths through the code to deal with the ! are likely to be the problem -- unlikely to be tested thoroughly because -- who makes filenames with ! anyway?

rmkaplan:
I don’t get an error with UNPACKFILENAME.STRING, the result looks good except for the STRIKE. The exclamation mark (and also semi-colon) appear to be screwing the extension recognition (the finite-state machine has probably already gone into version mode).

I rewrote UNPACKFILENAME a couple of years ago, this is probably mine.

Note also that if I apply UNPACKFILENAME (not .STRING) to the same input, I get an uninterruptable crash into URAID, trying to build an atom that’s too large.

nbriggs:
> Note also that if I apply UNPACKFILENAME (not .STRING) to the same input, I get an uninterruptable crash into URAID, trying to build an atom that’s too large.

It's trying to deal with the "strike", but passing a block of memory with length -7

```
Fname is IL:MKATOM
## STACK FX at 0x13680 ##
[cnt = 0 ]
 #alink           0x3612 
 fnhead   0x2f9088 
 nextblock        0x368e 
 pc               0x81 
 nametbl  0x2f98a8 
 #blink           0x36a8 
 #clink           0x86 
  1368a : 0x  2f  0x38b0  
  1368c : 0x  60  0x2ea8  

< f4 
IVAR -------
  136da : 0x  60  0x5e2a  *local* IL:BASE  {IL:\UNBOXEDHUNK7}0x605e2a
  136dc : 0x   e  0x  16  *local* IL:OFFSET  22
  136de : 0x   f  0xfff9  *local* IL:LEN  -7
  136e0 : 0x   0  0x   0  *local* IL:FATP  CL:NIL
  136e2 : 0x   0  0x   0  *local* IL:FATCHARSEENP  CL:NIL
```
[...]

PR #1421, opened 2023-11-14
ADIR fix UNPACKFILENAME.STRING  addresses #1416

rmkaplan:
The early . was seen as starting an extension, the ! then was seen as starting a version.  When the other marks caused those hypotheses to be revised, the extension's end wasn't being revised


masinter (approved):
LGTM

PR #1428, opened 2023-11-20
Fix loadup-apps-from-full.sh so that it fails gracefully when a notecards directory cannot be found

fghalasz:
Previously  loadup-apps-from-full.sh failed in a mysterious way when the Notecards directory cannot be found.  Now fails with a clear error message and an exit 1.


masinter (approved):
LGTM

PR #1430, opened 2023-11-23
Restore NSPROTECTION files to lispusers

nbriggs:
The NSPROTECTION files (and likely others which I have not yet identified) provide functionality that is useful with the Dodo XNS networking implementation.  Move them from obsolete/lispusers back to lispusers/


masinter (approved):
LGTM -- what was I thinking?

PR #1435, opened 2023-11-29
Rmk94  fix checkeolc macro

rmkaplan:
FIxes confusion of ANY and CR EOL conventions in the macro, recompiled callers.

(I thought I had pushed this before, apparently not.)


masinter (approved):
there is an awful lot of complexity in end-of-line handling that I can't believe is necessary. It's being DWIMISH at the wrong level.
Some simplifying assumptions by composing pipelines when actually necessary. I haven't seen a CRLF file in years. Even if there remain some CR applications, it would remove all the extra checking and byte-length calcualations. But I understand, there's a lot of "sunk costs" in having gone down this road in the first place. 
I'd be surprised if there weren't edge cases where the code doesn't work -- it's hard to think about.
But... in for a penny, in for a pound.

Approved.

Issue #748, opened 2022-03-18
If GETFN of an imageobj is not defined, try WHEREIS


rmkaplan:
A nice DWIM idea that came up in a meeting

rmkaplan:
This capability was added, and it eliminates some of the "unknown image object" cases.  If the GETFN is not defined, the code finds its file in the WHEREIS database, and asks whether to load that file.  But the residual problem is that the file with the GETFN for a particular object may not contain all the information need to actually interpret and display the object.  The GETFN SKIO.GETFN is on SKETCHOBJ, but the root file SKETCH is needed before the imageobject and be retrieved and displayed.  There is currently no mapping between the getfn and the necessary root file of a multi-file module.

One approach would be to define a convention for the putfn that it not only writes the getfn and other image object data, it also adds a particular property that names the root file.  If that property is present, then that file (and all of its descendents) would be loaded instead of the file in the WHEREIS database.  Essentially, when a SKETCH image object is created, a property ROOT-FILE with value SKETCH would be implicitly added.

Another idea would be to piggy back on our two-level naming scheme. All the files for TMAX and now TEDIT are in subdirectories TMAX> and TEDIT>.  When you LOAD(TEDIT), it has TEDIT>TEDIT on the search path, when you load TEDIT-FILE, it has TEDIT>TEDIT-FILE, etc.  We could use a similar naming convention for image object GETFNS.  E.g. rename the function SKIO.GETFN as SKETCH-SKIO.GETFN.  If that function is not defined, peel of SKETCH and try for that file.

This is a silly nuisance, it would be nice to define a protocol that we can slowly migrate to.

masinter:
wouldn't it be simpler to just enforce that any file containing a GITFN should, when loaded, load everything else you need?

So just change SKETCHOBJ to (FILES SKETCH)? When would you NOT want that?

rmkaplan:
I presume the FILES would have to be at the top of the file, since any P’s before it might not execute if SKETCH hasn’t first been loaded.  Could there be a problem if SKETCHOBJ loads SKETCH which loads SKETCHOBJ?  Are the FILE properties set up at the beginning so there isn’t a cycle?

masinter:
It checks for recursion so if you're already loading FOO it won't try to reload.

So I think this is fixed except we need to make sure all of the GETFNs anywhere will load
the whole package.

[...]

Issue #1167, opened 2023-04-20
HPRINT problem


masinter:
FORGET, then some commands, then (HPRINT LISPXHISTORY) fails on GREATERP(0 NOBIND)

@ says the error is at CL:^^^^Z%%NOT-NONCOMPLEX-NUMNR-ERROR

masinter:
it's no longer clear to me what's going on here... closing this for now

PR #1436, opened 2023-11-29
WINDOWOBJ: Better handling of unknown image objects

rmkaplan:
2 changes:   
1. If WHEREIS says that an unknown getfn is on FOO-FIE and FOO>FOO exists, then offer FOO in the mouseconfirm.  This should get all the support code (e.g. TMAX is offered and loaded instead of TMAX-NUMBER).  Issue #748.   (Sketch files would be renamed to this hyphen naming convention.)

2. Separately, if the getfn is still not found when the file is opened (so the image object is encapsulated), the encapsulated imagebox fn will upgrade the image when the object is redisplayed, if the getfn has been defined in the meantime.  So you can TEDIT a file FOO with a Sketch image object, pass on gettiing the image object during the open, then load SKETCH explicitly.  The objects will upgrade when they are redisplayed.


masinter (approved):
See PR #1166 and #1167 -- perhaps we can consolidate? 

I want to set up so that HCFILES can try harder and maybe break instead of printing UNKNOWN GETFN

rmkaplan:
An easy addition would be a global variable that suppresses the MOUSECONFIRM, a file will load if it’s there.

We could also loosen up a bit:  Assume that FOO will load for FOO-FIE just if they are sisters, not necessarily require that they also be in a FOO> subdirectory.

The general idea is that the rootfile of an application can be computed systematically from the name of the file that contains the getfn, according to WHEREIS.

In the case of Sketch, renaming just the file containing the getfn to have a hyphen prefix SKETCH- would make SKETCH available.  (But I think moving all the Sketch files down a level is a separate thing to do.)

PR #1446, opened 2023-12-05
Rmk96 move sketch files to library/sketch 

rmkaplan:
This moves library/SKETCH to library/sketch/SKETCH, and all the library/SKETCHXXX files  to library/SKETCH-XXX.

This follows the hyphen-file convention of other multi-file applications (like TEDIT, TMAX).  It means that when the sketch GETFN on SKETCH-OBJ is required, the file SKETCH and all its components will be loaded instead.


masinter (approved):
some files were just renamed without remaking fileCOMS?

Anyway, running with this shows no problem.

rmkaplan:
I thought I got all the fileCOMS.  If I missed some, I'll fix them.  Should be consistent.

rmkaplan:
Actually, I think the coms are correct.  I did not move the files MATMULT and SCALEBITMAP into the SKETCH subdirectory, I think SKETCH is just one of their clients, they are not part of SKETCH.

rmkaplan:
But the file SKETCHOPS should not have been carried along.  It is deleted in PR #1464

Issue #1447, opened 2023-12-05
Cannot (LOAD 'NEW-CLOS-BROWSER 'PROP) from medley/clos due to inconsistent symbol exports


MattHeffron:
**Describe the bug**
The :INTERNAL/:EXTERNAL state of some symbols in NEW-CLOS-BROWSER is inconsistent between the source and the .DFASL files. This causes (LOAD 'NEW-CLOS-BROWSER 'PROP) to fail with MISSING-EXTERNAL-SYMBOL.

**To Reproduce**
1. CD {MEDLEY}<clos>
2. (IL:LOAD "DEFSYS.DFASL" T)
3. (CLOS::LOAD-CLOS)
4. (ED 'CLOS-BROWSER::BROWSE-CLASS)
5. Answer Y when asked to load the file PROP
6. MISSING-EXTERNAL-SYMBOL break window opens for the external symbol CLOS-ICON in package CLOS-BROWSER

Steps 1-3 are described in clos/README.MD as the way to load CLOS.

**Expected behavior**
The load PROP should have succeeded, and SEdit opened on CLOS-BROWSER::BROWSE-CLASS

**Diagnosis**
It appears that 4 symbols were exported before the source was written. However, that they were exported was not captured in the source in such a way that COMPILE-FILE would also export them. When the DFASL was loaded in a "clean" Medley, the symbols were not exported. 
The (LOAD ... 'PROP) fails because the symbols in the source file are formatted as external (one **:** not two **::** )
CLOS-BROWSER:CLOS-ICON, CLOS-BROWSER:CLOS-BROWSER, CLOS-BROWSER:ADD-BROWSER-METHOD, CLOS-BROWSE:CLASS 
The error occurred reading the IL:FILEDATES expression's "changes to". 
(It would have failed reading the NEW-CLOS-BROWSERCOMS if they weren't mentioned in the IL:FILEDATES expression.)

**Context (please complete the following information):**
 - OS: Win 11 WSL
 - OS Version: Ubuntu 22.04.3
 - Host arch: x86_64
 - Maiko version: 7fce08f92
 - IL:MAKESYSDATE: medley-231122-b2f750e5 -- (MAKESYSDATE) gave error "MAKESYSDATE is an undefined function"

MattHeffron:
See Pull-Request #1448

PR #1448, opened 2023-12-05
Change files DEFINE-FILE-INFO (i.e., the IL:MAKEFILE-ENVIRONMENT) to fix issue #1447

MattHeffron:
Use DEFPACKAGE instead of in-package, use-package. Added (:export ...) of the 4 exported symbols.


MattHeffron:
The source file NEW-CLOS-BROWSER diffs as essentially a replacement due to the original having Mac EOL (CR) and this one having Unix EOL (LF).  Beyond that, there are a lot of changes due to pretty-printing removing unnecessary font change characters.
The key changes are the DEFINE-FILE-INFO at the top of the file, and the writing of the IL:MAKEFILE-ENVIRONMENT property at lines 331 (old) - 276 (new).

masinter (approved):
good to see some work on the Common Lisp implementation.
d

PR #1470, opened 2023-12-21
Minor fixups (typos, formatting) to a few existing documentation files

rmkaplan:
Plus converting the Unicode documentation from TXT to formatted.

Note that COMPARETEXT, used for GITFNS-compare, doesn't recognize the formatting/looks differences in the binary-looks segment, only changes in the plaintext character sequence.

(It also doesn't recognize internal differences in corresponding image objects.  It would have been helpful if we had originally set up IMAGEFNS with a comparison function.  But when the file format gets upgraded to the next version (loosing backward compatibility), I think putting the results of the PREPRINTFN in the plain-text part of the file will partially address this, if the printed text is correlated with display differences.  The GETFN would then target the actual specification that moves to the binary looks segment.)


masinter (approved):
looks ok -- spot check but, as noted, difficult to review.

Issue #351, opened 2021-07-20
Want better Medley fonts -- Calling all strike fonts!


johnwcowan:
The Squeak version of Smalltalk (which uses a modified Smalltalk-80 image) is open source and still uses Xerox strike fonts, though it now supports OpenType as well.  Here's the current list from the image itself:

Accuat, Accujen, Accula, Accumon, Accuny, Accusf, Accushi, Accuve, Atlanta, Bitmap DejaVu Sans, BitstreamVeraSans, BitstreamVeraSansMono, BitstreamVeraSerif, ComicSansMS, Darkmap DejaVu Sans.

In addition, X ships with a large number of bitmap fonts represented in [BDF format](https://en.wikipedia.org/wiki/Glyph_Bitmap_Distribution_Format).  This format uses ASCII files to specify the metrics and the actual bits for each glyph (in hex).  PCF format is a compiled version of BDF; it can be decompiled by the standard command-line tool `pcf2bdf`.  The similar tool `otf2bdf` will convert OpenType fonts.  Note that it will be necessary to chop these into 256-glyph chunks for our purposes.

Here are some BDF font sources, all freely available:

 * [Standard X Windows fonts](https://www.cl.cam.ac.uk/~mgk25/ucs-fonts.html)
 * [IBM console fonts](https://github.com/farsil/ibmfonts)
 * [Search Google](https://www.google.com/search?q=%22bdf%22+%22fonts%22+%22download)

masinter:
this is good information, but what is missing for a good issue is a good statement of objectives (better rendering on high-res displays?  Eliminate black blobs when TEditing Unicode text files?) and a plan of attack (or possible things to investigate or test before proceeding).

johnwcowan:
> Better rendering on high-res displays?

For that we'll need vector font support, which is out of scope for the moment.

> Eliminate black blobs when TEditing Unicode text files?

For that we'll need (at least on my present view) a conversion from XCCS to Medley UCS-2.  I'm filing another issue on that point, which will be more detailed.

The main benefit is just having more fonts: we can write TEdit files that don't have that MADE BY XEROX visual stamp on them.

masinter:
Can we use smalltalk/squeak or other software to generate .strike fonts
* in higher resolution
* compatible with XCCS and fully fleshed out?
[...]

PR #1476, opened 2023-12-24
Suppresses size coercion for Titan 12

rmkaplan:


rmkaplan:
See issue #1477 for explanation

masinter:
so if we find or make a TERMINAL 14, this code is now suddenly wrong?
Without doing a FONTSAVAILABLE ?

We are much closer to having better fonts and this seems to push things in the wrong direction.

rmkaplan:
If we get better fonts, we won’t need this.

masinter:
[ttfonts.txt](https://github.com/Interlisp/medley/files/13788325/ttfonts.txt)
from chm-parc2 includes titan and terminal fonts > 12

nbriggs:
Aren't those an Alto version of Titan, not the XCCS encoded one being addressed here?

masinter:
@nbriggs yes, there doesn't seem to be a bigger Titan. And the Medley files have TERMINAL  8 10 12 and PCTERMINAL in 12 and 14 (in BRR and MRR). I'm not sure why we have TERMINAL and PCTERMINAL.

rmkaplan:
Maybe PCTERMINAL also has a different coding.

Improving the fonts is another large project.  Volunteers?

In the meantime...

Anzus:
I think the PCTERMINAL fonts are what was used by the PC Emulator card on the 1186.

masinter (approved):
this is an awful hack of an awful hack.
awaiting #351

PR #1480, opened 2023-12-25
Extended EQUALALL with BLOCKEQUALP test for equivalent arrayblocks

rmkaplan:
Added BLOCKEQUALP to LLARRAYELT to test for the equivalence of 2 arrayblocks (including those that are allocated as smaller same-size hunks).

Then inserted a call to BLOCKEQUALP in EQUALALL in HPRINT.


sources/LLARRAYELT
```diff
@@ -2130,6 +2120,49 @@ Copyright (c) 1982, 1983, 1984, 1985, 1986, 1987, 1988, 1989, 1990, 1991, 1992,
...
+        (LET (HDR1 HDR2 NCELLS PTRS (TY (TYPENAME BLOCK1)))
+             (AND (EQ TY (TYPENAME BLOCK2))
+                  [if TY
+                      then                                   (* ; "Fix-sized hunks")
+                           (CL:WHEN (STRPOS "HUNK" TY)
+                               (SETQ PTRS (EQ (CHARCODE P)
+                                              (NTHCHARCODE TY 2)))
+                               [SETQ NCELLS (SUBATOM TY (ADD1 (CL:IF PTRS
+                                                                  (NCHARS "\PTRHUNK")
+                                                                  (NCHARS "\UNBOXEDHUNK"))])
```
masinter, line 2141: this seems to put off to runtime things that are constants (like NCHARS).
If someone made a datatype called FOOHUNK would the STRPOS succeed?
It's not clear what's going on here.  The comment gives a clue but what is the story?

rmkaplan:
Good catch, those NCHARS can be embedded in CONSTANT. Maybe it would be more precise to do separate STRPOS’s for the 2 different prefixes and confirm with the FIXP test on the tail, and not do the NCHARS.  But better would be a more professional way of detecting pointers to the fixed-size hunks that \ALLOCBLOCK can produce, if anybody knows of such a thing.

In terms of the call from EQUALALL, for this to get confused with a user-defined datatype, its type-name would not only have to conform to one of these patterns, it would also have to be the case either that the datatype was defined without field descriptors or the field descriptions for that datatype are not available when the test is run.

rmkaplan:
I tightened up the hunk tests in BLOCKEQUALP (LLARRAYELT).

But as mentioned elsewhere (I think), the READTABLEP test and maybe the TERMTABLEP test in EQUALALL are bogus, left over from Ascii days.  I think it should work properly if those specific tests are removed, just letting it fall through the GETDESCRIPTOR (unless there are internal structures like alists or proplists where order is not defined).  I didn't deal with that in this pass.

The READTABLEP datatype declaration itself is suspicious.  It looks like it is set up for a 24-bit pointer world, padding out the high-order byte of pointer fields.  I also leave this for some other time.

masinter:
@rmkaplan we talked about this 1/8/23 meeting

masinter:
@MattHeffron wasn't there a better way to test if something is a hunk?

rmkaplan:
Yes, there is a very clean way of testing whether something is a hunk, and getting its properties.

The problem arises when \ARRAYBLOCK returns an actual block, not a hunk whose size is determinable from its datatype.  It appears that the block’s “size” may be larger than what was requested, for convenience of the allocator.  So 2 blocks of nominally the same requested size may actually be of different physical lengths, each of which is at least as large as the request.

If we assume (or require) that the allocator always clears any excess storage, then there might be a version of BLOCKEQUALP that confirms not that they have the same size with the same contents, but that they agree on their contents up to the MIN of the 2 physical sizes.  Maybe return T if they match on both size and contents, ? if they match on contents up to the minimum size.

masinter:
return NIL if the content of the larger block isn't cleared (0) in the part after the MIN size and the max.

rmkaplan:
Makes sense.  Would it still be good to have a non-NIL return value to indicate that an exact match is not completely confirmed?

masinter:
I'm a little fuzzy on what this is all for... there are perfectly good array data types. Is this an optimization we should move away from just to simplify?

I think in lieu of a deep understanding of the uses, it's better to be conservative. EQUALALL didn't descend into block before at all. So return T if you're sure they're equal and NIL otherwise would be conservative.   It will miss some but it's better than it was before.

masinter (approved):
reviewed 3/4/24

Issue #1486, opened 2024-01-03
Medley doesn't treat Caps-Lock + Shift like Windows and Ubuntu do


MattHeffron:
**Describe the bug**
Medley capitalizes letters if `Caps-Lock` is set **OR** _either_ `Shift` key is down. Windows and Ubuntu use an **XOR** between those, so that `Shift` _un-shifts_ if `Caps-Lock` is set. (And others? mac?)

**To Reproduce**
1. Clear the `Caps-Lock` mode on the keyboard.
2. Start a medley sysout.
3. Type `abc` in an Exec window.
4. Press the `Caps-Lock` key to set that _mode_.
5. Type `def` in the Exec window.
6. Press and hold either `Shift` key.
7. Type `ghi` in the Exec window.
8. Notice that the characters typed are entered as `abcDEFGHI`

**Expected behavior**
The characters typed would be entered as: `abcDEFghi`

**Context (please complete the following information):**
 - OS: Win 11 WSL Ubuntu
 - OS Version: Ubuntu 22.04.3
 - Host arch: x86_64
 - Maiko version: -not a maiko issue-

**Additional context**
This is in LLKEY in `\DECODETRANSITION`, as well as `SHIFTORLOCK` in the `BLOCKRECORD KEYBOARDEVENT`, and in `SHIFTDOWNP`.

nbriggs:
I don't think that's a bug, I think it's the way Interlisp has always worked.  Shift does not turn into un-shift on the Mac (in any application) when caps-lock is on, either.

nbriggs:
I wondered if it was in the keyboard hardware - but I think it's an OS thing, for the OS that is decoding the keyboard - a USB keyboard plugged in to my mac behaves as my builtin laptop keyboard (no un-shift), but plug the same keyboard into a FreeBSD 14.0 box and it does do un-shift.
It might be nice to have it as an option for those that are used to typing that way, but I wouldn't make it permanently that way.

nbriggs:
BTW - when you were testing Ubuntu, were you doing it on a keyboard connected directly to the Ubuntu system, or ssh'd in from your Windows PC, or with WSL providing the Ubuntu?

MattHeffron:
Making this optional should be pretty straightforward.
(Any suggestion on the name of the Lisp global variable to control this?)

I tested the Ubuntu **both** under WSL, and as the native OS on a laptop with the laptop's own keyboard.

nbriggs:
[...]

PR #1487, opened 2024-01-03
change for CapsLock & Shift to be XOR vs OR.

MattHeffron:
This addresses issue #1486.
It adds a SHIFTXORLOCK in addition to SHIFTORLOCK.
Because this is low-level, it should be tested significantly!


nbriggs:
I can't tell if this is an unconditional change from the old to the new behavior, or if it is selectable, and if it's selectable... how?

MattHeffron:
This is an unconditional change, as is.
Making it conditional should be pretty straightforward.

MattHeffron:
Many of the changes seen in the diff are due to PRETTYPRINT changes since previous LLKEY version (Aug-2021).

rmkaplan:
Yes, that’s what the GITFNS commands are trying to deal with.

masinter (approved):
lgtm

PR #1489, opened 2024-01-06
Rmk100  unicode utf 8 update

rmkaplan:
The first set of XCCS-to-UNICODE mapping files were cobbled together in 2021 partly by hand and partly from imperfect sources, mostly based on the 1987 XCCS version (XC1-3-3-0, 1987).  A short while later, Peter created and delivered a more complete version that targeted XCCS version 2.0, 1990 (the pdf of the 1990 XNS documentation is also stored here). His version corrected some errors in the earlier files, filled out the existing character sets, and added some more.  He also incorporated the Unicode character names for virtually every character.

The revised files languished on my machine, I have now moved them over to the unicode/xerox/ directory, replacing the old versions. This involved some renaming of files, to correspond more closely to the XCCS character-set names.

This PR also includes a relatively minor revision to the UNICODE Lisp file.  I added a few run-time interface functions to support Tedit's operation on UTF-8/Unicode files (external format :UTF-8).  And I updated the UNICODE.TEDIT documentation file.


masinter (approved):
The "software in use" theory leads me to believe that this   is a net improvement.

PR #1500, opened 2024-01-12
oops missing edit cpv

masinter:


fghalasz (approved):
LGTM.

masinter:
actually, this isn't enough. If you start with a clean clone of medley (no loadups directory), it fails.
If there's already loadups and you're just doing another one, it works fine.

PR #1506, opened 2024-01-14
Further cleanup on cpv script;

fghalasz:
1.   add back a ln_or_cp function and use it to cp only when ln fails. Deploy t5his function for all instances of linking/copying in the script;  
2. better handling of case where the unversioned destination file does not exist but versioned destination file DOES exist.  In this case creates a new next higher versioned file.  Previously, in this case just created a new unversioned file and did not create a new higher version file.
3. more informative error messages


nbriggs (commented):
Doesn't make Solaris happy, because /bin/sh does not support the "local" declarations, so an attempt to use cpv results in
```
% scripts/cpv /tmp/xxx /var/tmp/yyy
scripts/cpv[7]: local: not found [No such file or directory]
scripts/cpv[8]: local: not found [No such file or directory]
cp: Insufficient arguments (1)
Usage: cp [-f] [-i] [-p] [-P] [-z] [-@] [-/] f1 f2
       cp [-f] [-i] [-p] [-P] [-z] [-@] [-/] f1 ... fn d1
       cp -r [-H|-L|-P] [-f] [-i] [-p] [-z] [-@] [-/] d1 ... dn-1 dn
       cp -R [-H|-L|-P] [-f] [-i] [-p] [-z] [-@] [-/] d1 ... dn-1 dn
Error: failed to add yyy to /var/tmp
```

masinter:
note that #1352 inconsistent

fghalasz:
OK. Removed local declarations from cpv script to make it POSIX-compliant.  Ready for re-review.

masinter (approved):
lgtm -- not sure why this wasn't reviewed before.

PR #1508, opened 2024-01-15
Replace \TEDIT.FORMATTED1 with TEDIT.FORMATTEDFILEP

rmkaplan:
Internal \TEDIT.FORMATTED1 is being decommissioned.

Separately, I would like to put (MOVD? 'NILL 'TEDIT.FORMATTEDFILEP) somewhere early in the Lisp loadup, there is still at least one place where it is testing for existence before execution.


nbriggs:
> there is still at least one place where it is testing for existence before execution

I don't see that in any of the files that have the text TEDIT.FORMATTEDFILEP:

```
% grep -lr 'TEDIT.FORMATTEDFILEP'  * | grep -v -e loadups -e LCOM -e DFASL               
internal/MAILCLIENT
internal/venuesysouts/full.venuesysout
internal/envos/NSMAIL
library/tedit/TEDIT-FILE
library/lafite/NEWNSMAIL
library/lafite/docs/release-notes/Lafite-Update.TEdit
library/lafite/docs/release-notes/Lafite-Jun-88.TEdit
library/lafite/UNIXMAIL
library/lafite/LAFITESEND
obsolete/library/new/TEDITFILE
```

they all just call it directly without checking whether it is defined.

rmkaplan:
CL:COMPILE-FILE

nbriggs:
Ah, it's checking before using `\TEDIT.FORMATTEDP1` and `OPENTEXTSTREAM` (not the `TEDIT.FORMATTEDFILEP` that I was looking for), and it's in `sources/XCLC-TOP-LEVEL`

rmkaplan:
Yes, I want to get rid of \TEDIT.FORMATTEDP1, replace it internally by a more useful function (\TEDIT.GET.TRAILER, which helped isolate Lafite), and eliminate external dependencies.

nbriggs:
What are the advantages of predefining TEDIT.FORMATTEDFILEP as NILL rather than checking whether the functions are bound?  Obviously it needs to do something since you need to be able to `CL:COMPILE-FILE` when TEDIT isn't loaded, though it's not going to do anything reasonable if TEDIT isn't loaded and you try to CL:COMPILE-FILE a TEDIT formatted file!

rmkaplan:
The predicate function is simple, only looks at a few bytes, doesn’t depend on any deep Tedit details.  So, instead of putting in a dummy NILL, it could actually have its true definition early in the loadup.  And it could then return something useful:  the byte position of the beginning of the formatting part.  So a caller, like the compiler, could know to compile from 0 to that byte position (e.g. copybytes to a dummy stream and compile that).  The value could be non-NIL only if the content part is plaintext, does not include image objects.

nbriggs (approved):
There are two changes, only one of which is mentioned - \TEDIT.FORMATTEDP1 is replaced by TEDIT.FORMATTEDFILEP, but it also changes the character returned in place of an image object from (CHARCODE NULL) to (CHARCODE *) - which should be fine since this is presumably internal to the comparetext package.

rmkaplan:
Yes, I noticed that Lafite was set up to return * in that case (maybe an arbitrarily different choice that I made along time ago), so at least there is an indication that an image object actually appeared in that situation rather than have it effectively disappear with a null.  So I did the same thing in the compare.

It is still not a good solution, since it changes the stream itself to substitute this byte for any object that BIN would otherwise return, and that would happen if this textstream had a life outside of this particular context.  Perhaps this should be done under a RESETLST?

rmkaplan:
(although in this case the textstream is being opened locally and then thrown away, so it doesn't matter)

Issue #1496, opened 2024-01-11
NIL is not a NUMBER in FFILEPOS


masinter:
![image](https://github.com/Interlisp/medley/assets/1116587/4d060fd1-32c3-49e8-b3bf-d10318b4d213)
![image](https://github.com/Interlisp/medley/assets/1116587/6f713fe7-972f-4ec0-9271-a3ad618a56de)

looks like *BYTECOUNTER* is NIL

rmkaplan:
Can you give me the actual file name?  It isn't visible in the screen shot

rmkaplan:
The problem isn't that *BYTECOUNTER* is initialized to NIL in \SLOWFILEPOS.  The problem is in \XCCSINCCODE, the function that reads character codes in files in external format :XCCS.  If that function is called with COUNTP T, which it is, then it is assumed that it will set *BYTECOUNTER* to the number of bytes it read to produce the character.

In this case, for this file, it is not setting that variable.  And that's because the file is not a valid XCCS file.  At the byte position 619579 where I see the failure, there is a string of about 40 255 bytes all in a row.

I would have expected to see an error in that case, but I don't think I checked for that bogus configuration.  I suppose the solution when it can't interpret an initial 255 as a valid shift is to return 255 as the character, and say that it took only one byte?

masinter:
(FILESLOAD GREP)
cd medley/docs/medley-irm/
(GREP 'LISPXHIST "*.TEDIT)

gives an error. The medley-irm isn't a XCCS file??

rmkaplan:
Presumably the plaintext parts of IRM Tedit files are XCCS, but the binary formatting parts are not.  (BTW, Tedit can now produce UTF-8 plaintext parts).

Apropos of the discussion about exposing the byte pointer to the beginning of the format information, GREP could stop searching at that point.

rmkaplan:
With respect to having GREP do something useful on (most) Tedit files, if TEDIT.FORMATTEDFILEP is available, either because it is dummied to NILL or 4th-round Tedit is loaded, then calling FFILEPOS with (CAR (TEDIT.FORMATTEDFILEP stream)) as the END argument should cause it to stop searching before it enters the binary formatting part of the stream. NIL is returned if stream is not a Tedit-formatted file, so in that case it will go to the EOF.

[...]

PR #1513, opened 2024-01-20
GREP: avoids tedit-file formatting, font change chars in Lisp source files, adds TGREP

rmkaplan:
The tedit-file part works only in 4th round or later (see discussion in #1496).

New entry TGREP puts the results in a scrollable TEDIT stream.  

This also suppresses characters below (CHARCODE LF) in a Lisp source file.

A cosmetic different: filenames are preceded by a blank line and printed in bold.


lispusers/GREP
```diff
@@ -1,32 +1,34 @@
...
-")
 
 (PRETTYCOMPRINT GREPCOMS)
 
-(RPAQQ GREPCOMS ((FNS DOGREP GREP PHONE)
-                 (INITVARS (PHONELISTFILES))))
+(RPAQQ GREPCOMS
+       [(FNS DOGREP GREP LISPSOURCEOUTCHARFN TGREP)
+        (P (MOVD? 'NILL 'TEDIT.FORMATTEDFILEP))
+        (DECLARE%: EVAL@COMPILE DONTCOPY (FILES (FROM VALUEOF (MEDLEYDIR "loadups"))
```
masinter, line 17: could we load exports.all more consistently, or move the functionality into tedit or ...

nbriggs: Yeah, not a big fan of `(FROM VALUEOF (MEDLEYDIR "loadups"))` embedded in the COMS.

rmkaplan:
On that front, I tend to say (FILES (FROM loadups)…), because MEDLEYDIR is used basically to set up DIRECTORIES, LISPUSERSDIRECTORIES…  Is that better?

I think the only need for this is to get the stream declaration, so it can change the OUTCHARFN to suppress font change characters.  Wouldn’t be necessary if STREAMPROP saw the external format fields as properties, so that the format behavior can be changed without creating a whole new format.

masinter:
the font control-codes are just treated as a SEPRCHAR ... is this going to be used for EDITCALLERS too?
(This was originally the 'phone' LispUsers package, wasn't it?)

rmkaplan:
I got rid of the dependence on the stream record, by just suppressing lower-control codes in all cases. It could be made sensitive to LISPSOURCEFILEP, but that's probably not too valuable.  (Note that the prior set up had a terrible bug, anyway, this supercedes).

nbriggs:
Is it expected that if you do, when connected to `medley/library`
```
(GREP "FOO" "BROWSER.TEDIT")
```
you get a `TYPE-MISMATCH` error for `NIL is not a NUMBER` ?
[grep-error.txt](https://github.com/Interlisp/medley/files/14624663/grep-error.txt)

rmkaplan:
That’s interesting.

It’s trying to look just at the “logical” characters of the Tedit file, and it is passing in the byte position of the end of the plaintext part of the file to prevent the search from going uselessly into the formatting section (unless you were looking, I suppose, for the name of a font or the internals of an image object—then use the real grep).

But the branch for Tedit files should be outside of the search loop.  After the with-open-file opens it as a simple stream, if it satisfies the Tedit test, the stream should be coerced to a text stream and the search should run on that.  FILEPOS will probably break when it tries to map an image object through the upper-case array, so the stream should also set the property that returns a special character code for objects.

rmkaplan:
Try it now

nbriggs:
Works on BROWSER.TEDIT, but try it on lispusers/BLACKBOX.TEDIT -- I got a `HELP! FILE BUFFER FOR NON-FILE PIECE`

nbriggs:
When it finds the text in a TEdit file, it reports it as being found in (for example)
`(from #<IO Text Stream/137,5600)`
rather than the original file name.

nbriggs:
Test case:
```
CONN medley/lispusers
(GREP "FOO" (DIRECTORY "*.TEDIT"))
```

rmkaplan:
This is a problem in Tedit, Grep is OK.

(It seems to get in the wrong state when binning from a file is followed by peeking at an object.)

rmkaplan:
This is fixed in last commit.  (the HELP is an independent problem)

nbriggs:
OK, looks good then.  I discovered by accident that you have to be careful not to give it PATHNAMEs instead of atoms or strings for the files - grepping in `(CL:DIRECTORY "*")` fails, but that's `DIRECTORYNAMEP` failing because `PACKFILENAME.STRING` is unhappy.
image

nbriggs (approved):
Looks better, see comments in conversation.

rmkaplan:
Apart from fixing the lowe-level peekbin bug in Tedit, Grep could probably do more with image objects.

The line detection just ignores them for the match (you presumably can’t match on an object, and FILEPOS doesn’t like them), but it would be relatively easy to produce some representation of a particular image object on a line that elsewhere has a matching substring—a GETFILEPTR and a TEDIT.NTHCHARCODE.  We could at least show its getfn, maybe in italics, or even run the  display function of the image object so the whole thing appears in the output stream (if output is going to an image stream).  Or just run its PREPRINTFN.

PR #1518, opened 2024-01-26
A few fixes to UNICODE that I stumbled across.

MattHeffron:
READ-UNICODE-MAPPING-FILENAMES returned a bare string if FILESPEC matched 1 file (first clause of the (OR...) in join), this caused READ-UNICODE-MAPPING to fail. 
SHOWCHARS referenced variable CODE that should have been C 
Added FILETYPE property to UNICODE to specify TCOMPL compiler. 
(Other changes are formatting by pretty printer, not mine.)


rmkaplan (approved):
LGTM, thanks for finding these

PR #1519, opened 2024-01-27
In MODERNIZE, fixed \MODERNIZED.TEDIT.BUTTONEVENTFN so it can work with Notecards

fghalasz:
\MODERNIZED.TEDIT.BUTTONEVENTFN passes down a hardwired value for TITLEPROPORTION of NIL.  This prevents the Notecards left button title bar menu from ever being shown.  Replaced this hardwired NIL value with (WINDOWPROP W 'MODERNIZE.TITLEPROPORTION) so that Notecards can set this Windowprop and hence get its left title bar menu.


rmkaplan (approved):
I thought we had dealt with this quite some time ago.  Maybe I then screwed it up in trying to behave reasonably with split windows.

Anyway, LGTM

PR #1523, opened 2024-01-30
fix FILEPKG so that MAKEFILE(NEW) never references the 'previous version' and recompiles the whole file

masinter:


masinter:
I updated this PR to fix MAKEFILE rather than make a new MAKEFILE-NEW function.
I also changed the default for compiling -- which was set to only recompile those functions that were EXPRP; this was done for speed but it's not a good idea for many situations.

masinter:
closed by mistake

MattHeffron:
I thought that this actually had already been done...

masinter:
I made this PR a draft in order to combine the logic for the MAKEFILE(flie NEW) and the "normal" MAKEFILE rather than hacking another pass.

The problem is that there are a number of loose ends in the current design/implementation.
The current design is that it will copy the bytes of previous definitions without re-pretty printing. But that misses some cases where things have changed but the communication from MARKASCHANGED and teh file package is faulty.

PR #1527, opened 2024-02-03
Extended the COPYRIGHT property on FILENAME to allow for multiple successive copyright owners.

MattHeffron:
If the COPYRIGHT property on the FILENAME is a LIST of multiple ("owner" year1 year2 ...), then print all of the dates & owners on separate lines. 
The Owner entries MUST be chronologically ascending. 
The year1 of T still gives the extra notification, and reports the EARLIEST year as the creation.


masinter:
i'm not sure about this. The 'copyright' feature was never legally reviewed, it was done as an ad-hoc feature.  We had a brief (pro bono) consultation with a lawyer who suggested several possibilities, ranging from copyright (c) Interlisp.org or removing copyright and license information from the source code leaving a reference to the repository LICENSE statement.
I previously had introduced a new value for COPYRIGHTFLG of IL:NEVER: if IL:COPYRIGHTFLG is IL:NEVER then ignore prior copyright assertions.

Anyway, I'd like to hold off merging this in, at least for a while, with the hopes we could get more clarity on copyright notices. 

I suppose we should more strongly encourage setting COPYRIGHTFLG to NEVER.

MattHeffron:
I'm not concerned with Copyright notices for the Medley (and Lispusers) files, _per se_.
This code is for the cases of _resurrection_ of development of old Medley applications where the Copyright owner has changed name, or ownership has been transferred (which I'm _hoping_ to accomplish with my old Beckman Instruments Expert System applications, so I'm kind of jumping the gun here).
I could, _instead_, make a Lispusers file which just replaces these two functions in PRETTY so that the core Medley code isn't changed.
Do you think that would be a better path?

sources/PRETTY
```diff
@@ -1,16 +1,16 @@
...
-      :CHANGES-TO (FNS PRINTDATE1)
+      :CHANGES-TO (FNS PRINTCOPYRIGHT1 PRINTCOPYRIGHT)
 
-      :PREVIOUS-DATE "19-Jan-2022 20:35:18" {DSK}<home>larry>il>medley>sources>PRETTY.;1)
+      :PREVIOUS-DATE " 8-Feb-2023 16:21:26" {DSK}<mnt>e>Interlisp>medley>sources>PRETTY.;1)
 
 
 (* ; "
-Copyright (c) 1984-1990, 1999, 2018, 2023 by Venue & Xerox Corporation.
+Copyright (c) 1984-1990, 1999, 2018, 2023-2024 by Venue & Xerox Corporation.
```
masinter, line 13: But the copyright owners were something like 1984-1990 Xerox 1999 Venue 2018 Ron Kaplan 2023-24 Interlisp.org.  Rather than making auto-generation of (legally dubious) copyright and license statements to make it simpler. Oh, like a file property you can set to point people to the LICENSE statement.

masinter:
If there's a need for this, I'd prefer it to not disturb the copyright property: by allowing the COPYRIGHT property to include a string.

masinter:
closed by mistake

MattHeffron:
A different tactic appears more appropriate.

Issue #1522, opened 2024-01-30
UNICODE:  Should add function for merging mapping tables


rmkaplan:
UNICODE has functions for reading and writing the XCCS-to-UNICODE correspondence tables in the Unicode/Xerox/ directory.

The main reading function takes a list of character-set names, extracts the correspondences, and then gives the mapping pairs to a function that puts them in various arrays and alists where the external-format functions can rapidly access them.

But there isn't a clean interface for merging the mapping arrays corresponding to one collection of charsets with the mapping arrays corresponding to another one.  

The tables are initialized with latin, greek, various symbols, etc. but the large set of Japanese correspondences are not included by default.  So, if you want to work with Japanese UTF-8, you have to do some fiddling to extend the basic default list of character sets with the japanese, and then reread everything and built an entirely new set of arrays.

It would be better if you could just say that you wanted to merge the Japanese stuff into what you already have.  Better still would be a low-level interface that loads the mapping tables that you need on demand, i.e. if you encounter an XCCS Japanese character code that you want to put out into a UTF-8 file, go read the mapping table for that character set and smash it in.  (Harder to do in the other direction, figuring out from a Unicode code which mapping table you need to load--we would have to precompute tables that index in the opposite direction.)

[When you do have the right tables, then things now work as you would like:  When I put out a UTF-8 plaintext file of the Japanese Tedit file in obsolete/sunloadup, I get a file that is readable by external text editors.]

rmkaplan:
I will push a version of UNICODE that deals with this issue and also has some other cleanups.

I added a function MERGE.UNICODE.TRANSLATION.TABLES, that will merge an in-memory mapping table for a new collection of XCCS character sets into an existing in-memory mapping table for another collection of character sets.

I provided a function to produce a simple inversion of the XCCS-to-Unicode correspondences and write it on a file. When loaded in, the inversion makes it possible to find the XCCS charset that has a character that corresponds to a given Unicode code.

I then modified the slow case of the primitive code-translation macro to do on-the-fly reading and merging of the XCCS character set that contains the mapping for a not-yet-seen XCCS code or the character set that contains the XCCS code corresponding to a not-yet-seen Unicode code.

[...]

PR #1528, opened 2024-02-04
UNICODE provides on-demand loading of mapping tables 

rmkaplan:
see #1522


rmkaplan:
READBOM returns the right value (:UTF-8) for the md files that Matt pointed to,(EQ x (READBOM (WRITEBOM  x)))  holds for each of the formats.

rmkaplan:
This PR is replaced with #1620, I am closing it and it can be deleted.  #1620 does not have a conflict.

PR #1529, opened 2024-02-06
WIP: improve hcfiles using directory apply

masinter:


rmkaplan:
Here are a few comments after taking an initial look at this:

The NTH and ASSOC in MAKE-WHEREIS-LOOPS should be replaced by (GIT-GET-PROJECT 'LOOPS 'CLONEPATH).

I don't think that MAPFILES should be a special function here that defaults its base to MEDLEYDIR. We should provide a more general facility in DIRECTORY, perhaps just by adding a functional argument to the existing FILDIR that coerces the COLLECT into an apply.  Then MEDLEYDIR would be passed here as the default argument to the generic.

But then, I would prefer to see the default argument passed as (PSEUDOFILENAME (MEDLEYDIR)).  This would effectively simulate commonlisp's notion of a "logical pathname" (albeit currently without accepting path structures--a separate issue), would tend to make the file references portable to other environments.

rmkaplan:
Also, did you intend to switch from LCOM to DFASL ?

masinter:
I'm restarting with this PR -- will close others.

masinter:
subsumed

PR #1532, opened 2024-02-07
DUMPDB annoyingly always asked COPYRIGHT owner when dumping DB

MattHeffron:
... (unless `COPYRIGHTFLG` suppressed it).

I added `DEFAULTDATABASECOPYRIGHTOWNER` (`INITVARS` to `NIL`; to preserve current behavior). 
If it is `EQ` to `NEVER`, then the `COPYRIGHT` property on the _file_.DATABASE is set to `(NONE)` to forever suppress asking about copyright.
If any other non-`NIL` value, then `COPYRIGHTFLG` is bound to `DEFAULT`, and `DEFAULTCOPYRIGHTOWNER` is bound to the value of `DEFAULTDATABASECOPYRIGHTOWNER`.

(_I_ keep the copyright enabled for the tracking of _when_ I made changes in a file (roughly). This just provides a way to make the DATABASEFNS library be quiet when dumping a DB. All of this goes away once an alternative to Copyright notices in files is developed.)


masinter (approved):
discussed at 2/19/24 meeting.

PR #1533, opened 2024-02-08
COMPARESOURCES ignores arbitrary forms, not just comments

rmkaplan:
The IGNORECOMMENTS argument can be one or a list of atoms, and any forms head by those atoms will be replaced by **IGNORE** before comparison.  If IGNORECOMMENTS is or contains the atom T, then comments will be stripped before the comparison, not just replaced by **IGNORE**.  E.g. the list (NLAMA T) will suppress comments and ignore NLAMA forms.

It may also be useful for COMPARESOURCES to suppress all output and simply return NIL if there are no differences. But that involves a little more massaging, since currently the headlines are printed before the comparison is done.


masinter (approved):
4/1/24 mtg

PR #1544, opened 2024-02-17
record RING rename

masinter:
- WIP: improve hcfiles using directory apply
- LISPSOURCEFILEP returns NIL for files starting with; -- i.e., only for managed files
- DUMPDB annoyingly always asked COPYRIGHT owner when dumping DB (unless COPYRIGHTFLG suppressed). I added DEFAULTDATABASECOPYRIGHTOWNER (INITVARS to NIL; to preserve current behavior). If it is EQ to NEVER, then the COPYRIGHT property on the file.DATABASE is set to (NONE) to forever suppress asking about copyright. If any other non-NIL value, then COPYRIGHTFLG is bound to 'DEFAULT, and DEFAULTCOPYRIGHTOWNER is bound to the value of DEFAULTDATABASECOPYRIGHTOWNER.
- Added NONE in addition to NEVER as DEFAULTDATABASECOPYRIGHTOWNER to preset to (NONE).
- COMPARESOURCES ignores arbitrary forms, not just comments
- Update COMPARESOURCES.TEDIT
- Made this "smarter": If COPYRIGHTFLG is NEVER    or DEFAULTDATABASECOPYRIGHTOWNER is NIL    or the .DATABASE file already has a COPYRIGHT property Then    No need to do anything special (it already shouldn't ask) Else If DEFAULTDATABASECOPYRIGHTOWNER    is NONE or NEVER Then Set the COPYRIGHT to NONE (I.e., never mention it again.)    is SAME Then Same as the source file. If it doesn't have one, then just normal handling    is DEFAULT Then Use the general default for copyright: DEFAULTCOPYRIGHTOWNER    Otherwise: Enable the general copyright defaulting.               Hopefully, DEFAULTDATABASECOPYRIGHTOWNER is one of the COPYRIGHTOWNERS keys.
- Update license to mention Interlisp.org but portions copyright Xerox & Venue & others
- rename RING record to HRING in HANOI


MattHeffron:
There are a lot of commits in this PR that have nothing to do with renaming RING to HRING.

masinter:
This is nonsense... I must have used the wrong branch somewhere. Sorry, wil close.

PR #1545, opened 2024-02-18
more utils and remake recompile

masinter:
- WIP: improve hcfiles using directory apply
- Add in utilities to make PDFs from all in repo TEDITs and SKETCH files also trying to recompile everything in clean


masinter:
subsumed by other PRs

PR #1550, opened 2024-02-21
Fix DUMPDB bug referencing the wrong ROOTFILENAME for the COPYRIGHT property.

MattHeffron:
The setting to NEVER wasn't suppressing asking.
The issue was incorrectly getting the ROOTFILENAME as where to attach the COPYRIGHT property.


masinter (approved):
I think this might be an interim solution; seems like this is too much mechanism for what it is -- the file.database files should never have a copyright, without any flag.

 But it's fine  for now, too.

PR #1553, opened 2024-02-24
Add (FILES LAFITE) to UNIXMAIL -- must load lafite before loading UNIXMAIL

masinter:


nbriggs (approved):
LGTM.

rmkaplan:
In order to get UNIX mail to run, it appears that you also have to execute (LAFITEMODE ‘UNIX).  If UNIXMAIL hasn’t been loaded by then, shouldn’t that cause the file to load?

nbriggs:
I don't think so.  You load the delivery modes you want and it updates the menu, and if there are multiple you can choose one, otherwise it uses the single one loaded.  I'd rather not have it go off trying to load a file that may or may not exist, and may or may not be a file that contains a Lafite mode definition -- why should `(LAFITEMODE 'MAILSHARE)` go and load `MAILSHARE.LCOM`?

nbriggs:
(and the modes are not predefined, so there's no reason to know to map UNIX to UNIXMAIL)

PR #1556, opened 2024-02-27
Some initial work on adding support for Japanese glyphs.

MattHeffron:
(At this point I'm not sure this is the right approach, but I don't want to lose this work. Just in case it's useful.)


MattHeffron:
Essentially obsolete

PR #1557, opened 2024-02-27
Initial changes to make GITFNS work with a personal GitHub repo.

MattHeffron:
It still expects the repo to be on GitHub, but this seems to help a bit.


rmkaplan:
@MattHeffron, we talked about this in the technical meeting today.  It isn't clear where this now stands, given the more recent changes in GITFNS.  In particular, is this related to the problem that showed up in Maiko, where there are PRs based on repositories not owned by Interlisp?  The current GITFNS recognizes that situation and prints a warning.  It doesn't yet run the additional git commands to make them accessible, but that's on the list.  If that is done, would it solve the problem that you were working on here?

MattHeffron:
The situation these changes were made as a start to address was that I had a local git repository that was NOT on GitHub. The GITFNS seemed not to be able to deal with this kind of repository. As I recall, these edits handled some of the issues, but the dependency on using GH meant that some things couldn’t be addressed. (Pull requests are irrelevant for a local repo.) I don’t know what other issues I ran into, so I brute-force solved this by pushing this repo up to a private repo on GitHub.

I don’t know if any of my changes would be useful, in general, for GITFNS.

 

Matt

MattHeffron:
This seems to be _trying_ to add to GITFNS functionality outside of its design.
I'm comfortable closing this instead of leaving it in Draft indefinitely. 
(Just don't delete the branch, in case there's future interest in looking at it.)

Issue #1509, opened 2024-01-16
Hyphenate file names in library/lafite/


rmkaplan:
Except for LAFITE-INDENT, none of the filenames in the lafite subdirectory conform to the hyphenated filename convention uses by TEDIT, TMAX, and now SKETCH.

It appears that, as a consequence, the lafite files are not analyzed when the whole-system masterscope database is constructed.  And functions on them are not found by WHEREIS unless connected to that subdirectory.

The hyphen convention has proven to be useful and robust.  I propose renaming the Lafite files according to this convention.

masinter:
i'm uneasy about this special rule ... but can we write it up in the README for the medley repo?

rmkaplan:
Unless we come up with a better solution for organizing all the subfiles of a larger application, I think this is effective and good enough.  An additional important advantage of having all such files in a separate subdirectory is that you have a natural way of focusing the targets of FB, FINDCALLERS, EDITCALLERS.

rmkaplan:
The file LAFITE has a variable LAFITEFILES that includes OLDNSMAIL as well as NSMAIL and NEWNSMAIL.

The files that Lafite actually loads are specified elsewhere, so this list may just be away of setting up a Lafite development environment.

But OLDNSMAIL doesn't appear to exist.  Is that just a hangover from an early phase?

MattHeffron:
This looks like it has been done. (hyphenation of file names)
If so, please close as complete.

rmkaplan:
Completed and closed

nbriggs:
lispusers/UNDIGESTIFY was not updated to refer to LAFITE-DECLS instead of LAFITEDECLS.

rmkaplan:
Oops, new pr makes the change.

PR #1558, opened 2024-02-27
Rmk107  Rename lafite files according to hyphen conventions #1509

rmkaplan:
All files in library/lafite/ now begin with LAFITE-.


nbriggs:
It says there are conflicts on library/lafite/UNIXMAIL (and UNIXMAIL.DFASL).  It's claiming that the LCOMs weren't renamed, but the old ones were deleted and the new ones added, and that LAFITETEDIT wasn't renamed LAFITE-TEDIT, and a few other things.  Have a look at the file changes list it sent out, where anything that is "A" and "D" should really be "R".

rmkaplan:
I don’t understand what happened.  There were 2 commits, the first one was after doing git mv’s to rename all the files to hyphenated versions (including LCOMS and TEDIT).  After that there were no non-hyphen files left in the directory (according to ls).

I then copied updated versions (fileCOMS etc) of those files from my working directory, which did not contain any of the unhyphenated files.  And did the second commit.

So maybe git mv was not the right command? How did the original files come back?

(Separately, the goal of this is to preserve history.  But I don’t think there is much git history with these files.  Since uploading to git, has anything been done here besides the few changes that I made to update for Tedit and external-format stuff?)

nbriggs:
The conflict with UNIXMAIL is because Larry updated UNIXMAIL to load LAFITE after you branched from master (or you hadn't updated your master branch before you created the new branch)

Your branch also seems to include Larry's "Make macros that expand to fetch or replace note the record fields used." commit to MSANALYZE.

There is one missed change to a file name, LAFITE-TIMEDDELETE still has a `(FILESLOAD LAFITEFIND)`

I looked at the commits, and I don't know why github is reporting the changes in the way it is - they seem correct.

rmkaplan:
I’m pretty sure I started with a master that already had the UNIXMAIL update, although I had done the renaming in my working directory before that.  I was careful to put Larry’s change separately into the renamed version that I copied over.

The “Make macros” had already been merged before I created the renaming branch, that shouldn’t be an issue.

nbriggs (commented):
No, adding it back to delete it is not the right answer.  It needs to be resolved with command line git, since they're binary files as far as git is concerned.

The point your branch started from on master was
```
commit 2647d98f8f21e59b05318733a039b18d07c577ec (tag: medley-240205-2647d98f)
Merge: 39ee2ecb b52015e7
Author: Frank Halasz <frank@halasz.org>
Date:   Mon Feb 5 15:12:13 2024 -0800

    Merge pull request #1519 from Interlisp/fgh_modernize-for-nc
    
    In MODERNIZE, fixed \MODERNIZED.TEDIT.BUTTONEVENTFN so it can work with Notecards
```
Let me see if I can work this one by hand.

nbriggs:
LAFITE-HAX and LAFITE-PRIVATEDL didn't get compiled and included?
LAFITE-PRIVATEDL will compile but won't load because ADDRESSPARSERRDTBL is unbound.  Not sure where that is supposed to come from, it's not in any current LAFITE file (or anywhere else I can find) - it might be SMTP related?

rmkaplan:
That’s the way they were, I just renamed them to carry them along.

rmkaplan:
BTW, LAFITEPRIVATEDL was written by N.H. Briggs in 1987.

PR #1561, opened 2024-02-28
Fix problems due to FX record incompatibly on SPY, miscompiled

masinter:
The record definition of FX was wrong .. not sure if the bad version on SPY polluted the LLSTK version.
In any case, this brings in new SPY, new LLSTK with improved records (only changed two fields NAMETABHI and LO)
and recompiled PROC (only use of badly defined fields outside of LLSTK.)


nbriggs:
Looks good, and a hard one to find!  I compared the (correct) definition to what's in the Maiko code and it appears to match.

nbriggs (approved):
LGTM.  I've tested the case which previously produced an error, and checked that the record definition matches what Maiko has been using.

masinter:
To test: in Interlisp exec: 
* `(LET ((X 0) (Y 1)) (HELP))`
* at the HELP! prompt, type  `PB X` and return.
Before, you would get an error. Now it should Print the Bindings
image

Issue #1560, opened 2024-02-27
TTYINPROMPTFORWORD  does BKSYSBUF if terminated with mid-string CR


rmkaplan:
TTYINPROMPTFORWORD is advertised as a nice way of allowing mouse-based editing for simple user inputs, a la TTYIN.  TEdit uses it (via TEDIT.GETINPUT) when asking for file names, search targets, etc. in the prompt window.

If the call to TTYINPROMPTFORWORD can include a list of terminating characters.  When any of those characters is entered, the editing session ends and the string in the buffer is returned.

However, if CR is one of the terminating characters and the session is terminated with CR, the result may not be as intended.  It depends on where the caret is when the CR is entered.

If the CR is at the end of the string, the whole string visible in the buffer is returned, as expected.  However, if the caret is somewhere earlier in the string, then the return value is only the prefix in front of the caret.  But the characters after the buffer are not thrown away, instead they are stuffed into the TTY buffer and show up in the current TTY stream.

For example:  if the string is foofum with the caret between after foo, the value is "foo" and "fum" appears as input somewhere else.

For TTYIN in the exec, the CR goes to a new line, and the suffix characters then appear at the beginning of that line, which makes sense--the editing session continues in place.

But in the Tedit (and maybe other cases), the effect for the Find operation is that the search string is just the prefix...and the suffix is silently inserted in the text at the position of the caret, presumably corrupting the document.  The effect when trying to get a filename is that you only get the prefix as a name, and again the suffix may be inserted somewhere.

A solution for Tedit is to disable CR as a way out, require perhaps only something like ^X as  a terminator following the TTYIN convention.

Or to somehow fix TTYIN so that CR as a specified-terminator takes precedence over CR as a new-line insertion. (Although, again in the TEDIT case, you might want to be able to search with a string containing CR.)

rmkaplan:
Just to reinforce the point, the rename command in the file browser also demonstrates undesired behavior.  If you aren't careful to move the caret to the end before you hit CR, you end up with a truncated filename--a pain to recover from.

masinter:
I think we might consider changing TTYIN to treat a user press Enter when the caret is in the middle of the line as a "move to end" first.

fghalasz:
Re: CR as "submit" versus CR as "new line" is a pretty common issue in  various web apps.  One pretty common solution is to have CR be "submit" and CTRL-CR (or maybe Meta-CR) be "new line".

rmkaplan:
[...]

PR #1565, opened 2024-02-29
Fix for TTYINPROMPTFORWORD's bad behavior on CR termination (cf #1560 )

rmkaplan:
Brute-force addition of parameter to TTYIN that BKSYS/DELETE function (TTYINUNREADBUF) can access freely.

Without understanding or fiddling anywhere else in TTYIN, this seems to work.  It seems to fix the really bad behavior described in #1560.  I still don't see anyway of getting a CR anywhere in the output string.


masinter (approved):
TTYIN hacking based on understanding what it's trying to do is on the critical path to getting other problems fixed.
But as a short term solution, sure

PR #1571, opened 2024-03-07
IMTEDIT and IMINDEX

rmkaplan:
Backquote commas were escaped in IMTEDIT.  Also put the AFTERHARDCOPYFN on the text proplist, removed the advice in IMINDEX.


nbriggs (approved):
This is good.  It's necessary after the latest TEdit changes or you can't do anything with Library TEdit files.

PR #1573, opened 2024-03-09
Rmk110  fix unpackfilename.string

rmkaplan:
@nick encountered a segmentation fault on Intel (but not Apple silicon) macs when UNPACKFILENAME.STRING was called under PACKFILENAME as
      (UNPACKFILENAME.STRING "" NIL NIL 'RETURN)
This now returns NIL also on Intel, which this also now matches the behavior of this function in the Venue sysout.

The arguments besides the first one are not documented, and there is no indication in the documentation that this function will ever return NIL, it explicitly says that empty fields are denoted by "".  But internal functions that call this have always returned NIL if the field isn't present.  Thus,(FILENAMEFIELD.STRING "" 'DIRECTORY) returns NIL, but (FILENAMEFIELD.STRING "<>" 'DIRECTORY) returns the empty string. FILENAMEFIELD returns NIL in both cases.

I'm leaving this behavior alone.

(This PR also moves the coercions of GENERATION and STRUCTURE into UNPACKFILENAME.STRING into FILENAMEFIELD and FILENAMEFIELD.STRING, so the behavior is uniform for all callers.)


masinter (approved):
LGTM -- curious how it wound up with a -1.

rmkaplan:
The -1 was because it was an empty string, and in that case the end ptr is less than the offset

PR #1576, opened 2024-03-10
DINFO was expecting to use a closed textstream without reopening

rmkaplan:
Not sure why it got closed.  Also fixed some escaped backquote commas (not sure that was necessary, but...)


masinter (approved):
LGTM

Issue #609, opened 2021-12-05
Evaluate compatibility of Medley Common Lisp (formerly XCL) with popular books about Common Lisp


masinter:
As a Lisp learning environment, divergence from the spec in obscure edge cases may not matter as much.

Evaluate and document divergence from exercises and examples for PCL and PAIP.

This is a step toward ANSII Common Lisp compatibility which would be useful.

reflektoin:
As a beginner using the environment it took me a while to load the CLOS in the online version. Hopefully these steps are helpful for someone in the future. So to load the CLOS I did the following steps:

1. Open Interlisp Exec. Steps for opening the Exec are explained for example [here](https://github.com/Interlisp/medley/issues/690#issuecomment-1047269639).
2. Navigate to clos folder with command "cd <home>medley>medley>clos"
3. type "(LOAD 'DEFSYS.DFASL)" into Interlisp Exec
4. Type "(CLOS::LOAD-CLOS)" into Interlisp Exec.

Now you should be able to do class definitions using "(CLOS:DEFCLASS ...)"

masinter:
@reflektoin I've been meaning to take this and put it in the wiki as a How To in the "Experiments" section.
Should we load CLOS by default when you ask for "Common Lisp"?
@fghalasz #742

masinter:
@fghalasz do you think it would be useful for now to link to this issue?

reflektoin:
@masinter: Was the question "Should we load CLOS by default when you ask for "Common Lisp"?" directed to fghalasz?

By the way, in your first post the link http://cs.cmu.edu/~dst/lispbook/book.pdf gives me "Document not found". I could access the file from link https://www.cs.cmu.edu/~dst/LispBook/book.pdf. It seems that character casing plays a role here.

masinter:
@reflektoin you, thanks. Anyone responding to this issue.

reflektoin:
No problem. 

About loading CLOS by default. Yes I would prefer that it would be loaded by default. And that might lower the barrier for people to experiment with Common Lisp when they don't have to do extra steps on order to run CLOS code.

pamoroso:
I'm going through Peter Seibel's book [Practical Common Lisp](https://gigamonkeys.com/book) and testing the code of each chapter on Medley to see how far I can go.

[...]

PR #1579, opened 2024-03-14
import an LOOP macro

masinter:


pamoroso:
On Medley Online, when I load CML-LOOP.DFASL I get this break window with the error `Can't find package for DEFINE-FILE-INFO reader environment XLP`:

cml-loop-error

Not sure I'm loading it correctly.

masinter:
Tbh I don’t know how to set up the package declaration so it works with makefile.--LarryMasinter.netOn Mar 14, 2024, at 9:19 AM, Paolo Amoroso ***@***.***> wrote:﻿
On Medley Online, when I load CML-LOOP.DFASL I get this break window with the error Can't find package for DEFINE-FILE-INFO reader environment XLP:
cml-loop-error.png (view on web)
Not sure I'm loading it correctly.

masinter:
I changed the file name to XCL-LOOP and added it to the loadup.
I fixed the package problems I hope.

pamoroso:
Now it seems to be working. Pretty cool.

xcl-loop

nbriggs:
Why XCL- instead of CML?

Should loop be in its own package or should it be part of cl:/lisp: package if it's part of the language spec (or xcl: if it's "our own" implementation)

MattHeffron:
I concur with @nbriggs . 
Why is this in the "XLP" package? Doesn't that mean that all of the symbols that the `loop` macro uses for its _extended_ form would have to be imported into whatever package the code using it is read in? Or all used as **xlp::_symbol_**?
In fact, the `loop` macro **_itself_** is in the "XLP" package and it isn't even exported from that package. So, even to use the simplest loop form requires either: 
**(xlp::loop _forms_)** 
or to **(import (find-symbol "loop" "xlp") "_name of package of code using loop_")** before the code using the `loop` macro is read, 
or to put a similar import as part of the DEFPACKAGE of the file environment of a file using the macro.

If it's part of the standard, then put it into the correct package.

Is the point not to redefine the existing `loop` macro as defined in CMLSPECIALFORMS? It looks like _that_ implementation and syntax would be handled correctly by the new _fuller_ `loop` macro.

masinter:
Interestingly the other words in the extended f  op rm are pname only and can be in any or no (I interned at all. Only two symbols are exported: loop and loop-finish. That's why it works without the contortions @MattHeffron predicted. Try it!

pamoroso (approved):
The latest version seems to be working fine and no special handling of packages is required. Here's a quick test of evaluating at a XCL Exec a few `loop` expressions from Peter Seibel's book:

xcl-loop

Running an infinite loop with `(loop)` works too.

pamoroso:
A call to this implementation of `loop` to iterate over a hash table produces the error `Undefined car of form XLP::WITH-HASH-TABLE-ITERATOR` as [reported](https://github.com/Interlisp/medley/issues/609#issuecomment-2002053775) in issue #609.

nbriggs:
Both `WITH-HASH-TABLE-ITERATOR` and `WITH-PACKAGE-ITERATOR` definitions seem to be missing. (Latest version of the branch moves the package for the internal mechanism of the `loop` macro to LOOP: and sorts out the conflict with CMLSPECIALFORMS which already defined LOOP)

nbriggs:
The author of the `loop.lisp` has the hash table bits as well.

masinter:
I changed the package to "LOOP" and made sure "LOOP" and "LOOP-FINISH" were exported from the LISP package.

masinter:
i spent the better part of a day on Both WITH-HASH-TABLE-ITERATOR and WITH-PACKAGE-ITERATOR and have decided to put that on ice because I don't want to change Medley's implementation of hash-tables or packages.
Those two features aren't part of LOOP.
I did find a couple of differences/bug fixes with a newer loop.liap.

masinter:
I missed the parts about -TABLE-ITERATOR and WITH-PACKAGE-ITERATOR being used in some LOOP iterators.
But I don't want to hold this for them. I made a separate issue.

PR #1580, opened 2024-03-15
Rmk112  fix tedit hardcopy smash

rmkaplan:
Nick and I have been trying to figure out why we are seeing system smashes during hardcopy stress testing.  This PR includes fixes to issues that we were able to identify, and also a new file TEDIT-STRESS that includes various stressing functions.

There is still an unresolved mystery:  Nick sometimes still sees some crashes when he runs the stress tests, but I now have not experienced them, even after 10,000 or so hardcopy attempts.  The changes here are an improvement to the current master, but may not be the last word.

Here is what I have come to understand about the original (and current) Tedit architecture.

The underlying problem is that there are some connections in Tedit's datastructures that are implemented in terms of Xpointers to avoid circularities that would prevent collection by the reference counting GC.  In particular, the main external datastructure is the text stream, produced by OPENTEXTSTREAM, and this is what supports the normal stream operations (BIN, BOUT, READ, PRINT...).  It points to the main internal structure, the TEXTOBJ, which holds all of the Tedit specific stuff.  The TEXTOBJ points back to its stream by an Xpointer (the STREAMHINT field). 

As a consequence, after a strean us created, if its textobj is passed around to clients who don't also have a handle on the stream, the stream can be collected even though it still might be in use.  And in fact, the stream is key to the line-formatting and display routines, both for hardcopy and the screen, since the formatter uses BIN to walk through all of the underlying text pieces to produce the vector of characters and image objects that make up each line, including the line breaks, widths and heights.  That vector then drives the display/hardcopy.

The hardcopy routines are fairly complicated (and I have so far tried to stay away from them) because each page is an assembly of boxes containing different typographic elements: a box for the main text and subboxes for headings, page numbers, footnotes).  Each of the subboxes is formatted by creating its own textstream with its own textobj, running the line formatter on that stream, and then returning the formatted lines for insertion in the main page.

There were 2 bugs.  The first was very local in the formatting routines for the subboxes.  The code was written to create the textstream and then immediately extract its textobj to pass to the formatter--without also first binding the textstream on the stack.  So the stream could have disappeared before the formatter extracted the stream from the textobj it was provided. 

The second bug had to do with how the lines computed in the subboxes are passed back for insertion in the main text.  The main text has to know not only the break and layout information, but also the source of that information, the stream/textobj.  It needs the source because it has to again extract the specific character information for display.  So, each line descriptor had a pointer field LTEXTOBJ containing its source textobj.  That tied down the TEXTOBJ but only temporarily, because the hardcopy routines smash out that field and eliminate the circularity (modulo an unfortunate interrupt) as soon as it is used.  (Happily, the screen routines don't use that field, just the page formatter).

But the TEXTOBJ was the wrong thing to tie down, since the Xpointer STREAMHINT doesn't guarantee the validity of the stream.

The fix is to change LTEXTOBJ to LTEXTSTREAM, save the stream in the line descriptor, and extract the textobj from the stream when it is needed, as opposed to the other way around.

(A more general solution, in the long run, is to systematically pass the stream instead of the textobj through all the internal function calls, not just hardcopy, where the stream might eventually be needed--the cost for this safety would be one additional fetch).


nbriggs (approved):
This has been working well for me.

PR #1582, opened 2024-03-16
Fix bug in setting package with wholine

masinter:
Fixes a bug introduced when sorting the menu items


MattHeffron (approved):
Looks good. Tests fine.

PR #1585, opened 2024-03-17
Off-by-one in the interpretation of the event length

rmkaplan:
Very minor.  Undo undo of move is still outstanding, also the problem with peeking after binning that Nick identified (maybe has to do with EOL convention)


masinter:
ok

masinter (approved):
looks ok -- 4th round problem?

PR #1586, opened 2024-03-18
Two minor Tedit tweaks to help Notecards get rid of TEXTOBJs

fghalasz:
Two minor tweaks to Tedit to help Notecards get rid of its use of TEXTOBJs.

1.  Allow TEDIT.PARALOOKS to take a TEXTSTREAM as it's "TEXTOBJ" argument.

2.  New function TEDIT.PROMPTWINDOW that returns the Tedit prompt window from a TEDIT.THING - a textobj, text stream or TEdit window.


rmkaplan (approved):
Looks good.  TEDIT.PARALOOKS is already fixed in my renamed version, I'll include TEDIT.PROMPTWINDOW there.  But let's upgrade the master now.

PR #1589, opened 2024-03-18
Fix issue with release workflow whereby not pushing docker image to docker hub / update versions on all actions 

fghalasz:
Scheduled kickoff of the buildReleaseInclDocker workflow was not properly storing the Medley docker image in Docker Hub.  This (hopefully) fixes this issue.

Also updated all action versions in all workflows to account for the fact that Node 16 has been deprecated.  New versions use Node 20.

### DIFF
--- a/.github/workflows/buildDocker.yml
+++ b/.github/workflows/buildDocker.yml
@@ -63,31 +63,6 @@ defaults:
 
 jobs:
 
-######################################################################################
-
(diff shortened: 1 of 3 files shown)


masinter (approved):
discussed 3/18/24 meeting. Can't easily be tested, problems don't happen with manual call to action, just the timing-based automatic push to DockerHub.

PR #1590, opened 2024-03-19
BUTTONS lispusers package commandeers tty process

MattHeffron:
The BUTTONS lispusers package will edit the button content in the TTY process. So, for example, a copy-paste to the EXEC that had the TTY is not possible. Changed to invoke EDITE as a new process, instead of in the TTY process.
(Lots of changes in file because change from FAST-formatted to PRETTY-formatted file. Actually only 1 line of code changed.)


masinter:
looks like Frank's changes got merged with this PR by mistake. Is this a job for git cherr7pick?

MattHeffron:
Closing this and will make a cleaner commit for the pull request.

PR #1594, opened 2024-03-19
The BUTTONS lispusers package commandeers the TTY process. (v2)

MattHeffron:
The BUTTONS lispusers package will edit the button content in the TTY process. 
So, for example, a copy-paste to the EXEC that had the TTY is not possible. 
Changed to invoke EDITE as a new process, instead of in the TTY process.

(Lots of changes in file because change from FAST-formatted to PRETTY-formatted file. Actually only 1 line of code changed.) Second, cleaned up, attempt.


masinter (approved):
minor change, lispusers package.
prc shows minimal difference
image

PR #1600, opened 2024-03-21
add more info on setting up directories

masinter:
I thought I had pushed through this afternoon but I apparently missed these little bits -- that you need notecards cloned in a particular place.


pamoroso (approved):
Looks good to me.

PR #1604, opened 2024-03-21
Add TEDIT.XYTOCH for Notecards

rmkaplan:
Provide a simple public interface function to isolate Notecards from Tedit internals


fghalasz (approved):
LGTM.

Issue #1601, opened 2024-03-21
TEdit:  TEDIT.NORMALIZECARET scrolls caret to second line when selected character is leftmost on its line.


fghalasz:
TEDIT.NORMALIZECARET is designed to scroll the textstream in the Tedit window such that the selection point (caret) is in the top most line in the window.  Most of the time it does this.  But if you select the leftmost character in a line (either LEFT or RIGHT selection), then TEDIT.NORMALIZECARET will scroll such that the selection is located in the second line - not the first.

Test case:

(SETQ W (WHICHW))
(SETQ CH# (TEDIT.SELECT 'LEFT 'TOP W))
scroll window to arbitrary point
(TEDIT.SETSEL W CH# 0 'LEFT)
(TEDIT.NORMALIZECARET W NIL T)

Result window will be scrolled such that what was in the topmost line during TEDIT.SELECT is now in the second line.

BUT, if pass (ADD1 CH#) to TEDIT.SETSEL, it will scroll properly such that the former topmost line is once again the topmost line.

Goal:  Trying to save the scrolling position of text in a card in the Notecards database and reuse it next time the card is brought up.  The size of the card's Tedit window is controlled so that its the same at all relevant times.

fghalasz:
Fixed in PR#1606

PR #1606, opened 2024-03-22
TEDIT.NORMALIZECARET moves line with left-most selection to the top

rmkaplan:
Removed what looked like an erroneous SUB1.  Address #1601 for Notecards.


fghalasz (approved):
LGTM.  Seems to work as expected.

PR #1611, opened 2024-03-24
#P"pathname" reads in as pathname

masinter:
phase 1: #P"string" reads in pathnames


masinter:
the one commit here is included in #1612 and there's no need to separate them...
I can find no instances of #.(PATHNAME ...) as a literal and this wouldn't affect that anyway.

PR #1612, opened 2024-03-24
hash p write pathname

masinter:
- **#P"pathname" reads in as pathname**
- **#P"pathname" used for printing pathnames**
do reading first


masinter:
I closed #1611 since it was included in this PR, and you will need both. This isn't changing the implementation, just adding the #P reader macro.

nbriggs:
Presumably this doesn't affect reading the old `#.(PATHNAME "foo")`, since `#.` is just an eval reader macro and `(PATHNAME "foo")` is a normal way to create a pathname.  There are a bunch of references to `#.(PATHNAME ...)` in the envos repository, though there are none that I can see in Medley or Notecards.

masinter:
If there were any #.(PATHNAME "string") around, they would read in no problem.

nbriggs (changes requested):
I believe loading this file 'ALLPROP and remaking it would fix the duplicate COMS issue.

masinter:
The duplication of CMLPATHNAMESCOMS in this case is a leftover from Interlisp-10, where the compiled code for calling a lambda-nospread function (LAMBDA N ... (ARG N 1) ...) was different than calling a spread. It is still needed for NLAMBDA's.

nbriggs:
OK, but I don't understand how having the exact same RPAQQ to set the CMLPATHNAMECOMS twice in different places in the file affects how lambda no-spread and nlambdas are compiled.  Can you enlighten me?

nbriggs:
And this was a newly introduced change - it was apparently OK before the addition of `#P pathname`, but something about that caused it to be added?

nbriggs (approved):
LGTM.

PR #1615, opened 2024-03-25
Remove extra scheduled Medley builds that were added for testing purposes last week

fghalasz:
Removing  from BuildReleaseInclDocker workflow the extra scheduled Medley builds that were added for testing purposes last week (since there is no way to test scheduled workflows on a non-main branch).


masinter (approved):
go for it

PR #1617, opened 2024-03-26
Minor updates to Tedit documentation files

rmkaplan:
Note that the prc compare of Tedit files doesn't recognize formatting changes.  So when no actual text is shown as changed, it means that the only change was (presumably) an improvement to the format of some character or paragraph.


masinter (approved):
documentation files -- trust the editor.

PR #1618, opened 2024-03-26
Fixes to \FORMATBYTESTRING and \FORMATBYTESTREAM

rmkaplan:
Motivated by the fact that FILEPOS was not being properly informed that the characters of the stream have stable, precomputable byte representation (unlike XCCS) so it can run at the byte level without on-the-fly character decoding.

The main change is that the format-string function returns NIL instead of "NIL" for a stable encoding.

The interface to the format-stream function (used to simulate the byte sequences of another stream) is slightly adjusted, and the XCCS function tries to set up the appropriate charset.


masinter (approved):
reviewed 4/1/24 meeting

PR #1620, opened 2024-03-28
On-demand UNICODE with READBOM, new .TEDIT, and small 2-way run-time mapping tables

rmkaplan:
Replaces PR rmk105--UNICODE-on-demand.

Only the smaller table-files UNICODE-MAPPINGS.TXT and INVERTED-UNICODE-MAPPINGS.TXT need to be distributed, they include all the information needed for on-demand mapping for all character sets.


rmkaplan:
This replaces #1528.  It installs the mappings more comprehensively in both directions, includes all of the UTF-16* external format definitions, revised documentation, and the two run-time-table files that are computed from the verbose textual mapping files.  The verbose mapping files in /UNICODE/xerox/ are not needed at run-time.

masinter (approved):
4/1/24 review in meeting... 
seems to have survived at least a little use.

Issue #1623, opened 2024-03-28
Add SEdit command to look up symbol documentation


pamoroso:
It would be handy to add to SEdit a command key, possibly bound to `Meta-D`, to look up in DInfo the documentation of the selected symbol.

@rmkaplan [suggested this feature](https://groups.google.com/g/lispcore/c/pfR0xPYY1qQ/m/AWCXxjTZAgAJ) motivated by [user feedback](https://fosstodon.org/@dziban@functional.cafe/112172917718445115).

pamoroso:
If the selected symbol has no DInfo or HyperSpec entry the command might show the output of `CL:DESCRIBE`.

masinter:
I think there are several sources of interest:
 the ARGLIST (sometimes an 'actual' and a presumed pattern)
 the DOCUMENTATION string (for DEFnnn) definers
 the output of CL:DESCRIBE
 The DInfo (for Interlisp) and CLHS (Common Lisp HyperSpec)
 The in-repo .TEDIT file for the LispUsers or Library package

In HELPSYS I tried to make ?<return> and the MAN command lookup all that apply among some of these sources.

I was thinking that having meta-H repeated cycle through the ones that apply.

rmkaplan:
A separate issue wrt meta-H would be to make it work not just on function-atoms but also on lists headed by function-atoms.  In that case it could show the correspondence between argument names and arguments, the way ?= does in TTYIN, and not just say that args are not available.

nbriggs:
@rmkaplan - it already does that - if the point is in a list it looks up the arglist of the head of the list.

rmkaplan:
It shows the argument names, as if you had clicked on the atom, but it doesn’t show the alignment:

Select the whole of (CONS A B), type meta-H, get
(CONS
X=A
Y=B
)

nbriggs:
@rmkaplan right - that would be better.  It would need to do it somewhere other than the single line SEdit prompt window.  It already is suboptimal on long arglists, even though it *says* it should show it in the main promptwindow if it's too big.

rmkaplan:
For me, when there are a lot of arguments, they show up in the system promptwindow (which in my set up is far off to the side and I don’t always immediately notice).  We could also pop up a temporary Tedit window next to the Tedit.

pamoroso:
SEdit now provides the `Meta-D` command. Can this issue be closed?

rmkaplan:
[...]

PR #1624, opened 2024-03-29
SEDIT-MAN  arms Meta-D in SEDIT for man page as per #1623

rmkaplan:
Marked as draft because  SEDIT:ADD-COMMAND doesn't quite work as documented.  In particular, it doesn't automatically put in both upper and lower case versions, even though all other commands are specified as case-insensitive.

So the work-around here adds 2 menu items instead of one.  

This issue would go away, if/when this is built into SEDIT-COMMANDS.  But that would create a dependency on DINFO being present.  So, another MOVD? ?

(Meta,I ) (for Info) would have been a better mnemonic, but that's taken for the Inspect command.)

(Separate issue:  SEDIT:GET-SELECTION causes an error if there is no selection, rather than returning SELTYPE NIL as documented.  NLSETQ suppresses that--CLIPBOARD probably also needs an NLSETQ)


masinter:
you didn't like the idea of multiple meta-H bringing in more documentation?
Or meta-?

nbriggs:
I don't like the idea of a modal interface and having to count how many meta-H have been done without changing the selection and reseting the count if you change the selection or issue any other command.

rmkaplan:
I didn’t think about key alternatives, just worked through the Sedit issues.  I think recognizing multiple keypresses would be hard.  I like meta-?.

nbriggs:
Entering meta,? will mean you have to type meta-shift-/ and that's not a particularly convenient key combination.

nbriggs:
and by the way, meta,/ is taken for extract so if you make a mistake typing meta,? you'll make a mess.

rmkaplan:
So, stick with meta-D ?

pamoroso:
To get `?` on my [Italian keyboard](https://journal.paoloamoroso.com/in-praise-of-cheap-keyboards) I press `Shift-'`. While pressing `Meta-?` (aka `Meta-Shift-'`) at an Exec flashes the screen in Medley Online, SEdit seems to recognize it as it warns `Unknown command: Meta-?`.

However, `Meta-D` would be more straightforward.

nbriggs:
I would (stick with meta,[Dd]) - though I'd think about whether it should be case-insensitive or whether meta,d and meta,D should do two related functions, say, meta-d does man and meta-D does cl:describe

pamoroso:
Good point, I actually meant the `D` key without regard for a specific case and either or both would be fine with me.

masinter:
My concern is that there are several kinds of documentation available unevenly. People say Interlisp is a "Lisp2" but it is really a "lisp N" where every filepkgtype is a (different or overlapping) namespace.

If the symbol is defined with a "definer" (FUNCTIONS with DEFUN or DEFMACRO, VARIABLES with DEFPARAMETER DEFCONSTANT .. ) then there might be a documentation string. Not all functions have one and FNS and VARS ..?
If the symbol is in CAR of form or in a `#'` FUNCTION context there are TWO argument lists, an 'actual' and one with extended syntax. And the possibility of showing the correlation to the argument list.
For things that are defined in the repo (WHEREIS ..T) and there's a TEDIT file that mentions it (FFILEPOS) .
And, finally, there is the possibility of showing the CLHS page.  

we could show all that are available at the same time... thats what HELPSYS does currently. In some cases its too much.

rmkaplan:
I don’t think there should be a lot of different keystrokes to distinguish a lot of different kinds of information that may or may not be available.  This is an interface to get information, it shouldn’t require the user to already have a lot of information in order to make use of it.

Right now I made it call the GENERIC.MAN.LOOKUP that the man command calls.  This often pops up a menu to select different types (although it isn’t exactly clear why the menu often has what appear to be duplicate items, sometimes with asterisks.  Mentions of the same thing in different places?).

We could have perhaps a different implementation function pop up a different kind of menu to let the user choose what source or what kind of display they want to see.  Or, as it does now with this generic function, let it rip.

masinter (approved):
let it out for others to experiment with, even though the design isn't settled.

rmkaplan:
Note:  this isn’t (yet) part of the loadup, to play with it you have to LOAD(SEDIT-MAN.LCOM).

pamoroso:
When I load SEDIT-MAN.LCOM on Medley Online and open SEdit (e.g. `(ED 'MYFUN '(FNS :DONTWAIT))`) the program hangs, i.e. pressing any key or clicking any mouse button on the SEdit window has no effect. The rest of the Medley session is responsive as usual though.

pamoroso:
I experimented a bit more with SEDIT-MAN.LCOM and it seems only middle-clicking has no effect.

pamoroso:
Never mind, middle-clicking seems to be working again with SEdit when SEDIT-MAN.LCOM is loaded.

PR #1626, opened 2024-03-29
unclear why this is necessary, but at least for now, insure CCLIPBOARD is enabled?

masinter:


rmkaplan:
This solves the problem in the initial loadup.  I tested a full without out, got a proper paste in my Interlisp exec, opened an XCL exec, got a flash instead of a paste.

I then made a new loadup including this patch, did the same tests, and paste worked in both execs.  Good.

However:  I then did a sysout with the patch installed, started up that sysout, and did the test again.  I got the now-familiar flash in the XCL window.

So however the interrupts are getting out of step, I don't think this hides the problem in general.  I think that (INSTALLCLIPBOARD) needs to be added to AFTERMAKESYS and AFTERSYSOUT (and maybe AFTERLOGOUT) forms.  It may be that whatever the problem is, it was encountered before when WHEELSCROLL was introduced.  It also sets up interrupts, and it also has entries on the AFTERxx forms.

(Which also leads me to wonder whether these two add-on faclities are somehow stepping on each other)

rmkaplan:
A little more:  I added INSTALL-CLIPBOARD to AFTERSYSOUTFORMS, did a sysout, and restarted.

Meta-V still flashed in an XCL window that was on the screen before the sysout.  But paste worked properly in a new XCL window created after restarting

nbriggs:
It would be good to know why this happens as well as what a fix is.  If it confused us it is likely to confuse others.

masinter:
There's an enormous amount of mechanism around WHEELSCROLL and CLIPBOARD; I'm hoping I can simplify.
Both of these add additional interrupts and enable some keyactions: WHEELSCROLL uses 4 codes (defined by the constants \WSUP, \WSDOWN, \WSRIGHT as CONSTANTS), while CLIPBOARD needs 1 (for PASTE). It's asymmetrical COPY is handled by a synchronous action by a call to an "add content" function for SEDIT, TEDIT, filebrowser etc.
But both WHEELSCROLL and CLIPBOARD use the same kind of "MOVD-ADVICE" mechanism that, while it seems to work fine individually but have some problems, where you MOVD?(fn fn.original) and then MOVD(newFN FN).  If done carefully, this has the desired property of being idempotent (you can repeat the operation without harm), but it's a problem in this case.

The function LISPINTERRUPTS is defined as 
image
but both WHEELSCROLL and CLIPBOARD use the MOVD-advice method to change what it calls.

LISPINTERRUPTS is actually defined to just return a constant list of interrupt character definitions. It is called by ADD-EXEC (to set up the termtable for each exec -- each process has its own interrupt list, so that, for example, the LISP interrupts don't get invoked when typing into a CHAT process window).

The simplest fix I think is to change it so that LISPINTERRUPTS just returns the value of the global variable LISPINTERRUPTS and have the ENABLE functions to add values to it.

I got down the road a bit on a bigger cleanup but want to try this simpler fix first.

masinter:
I wound up using LISPINTERRUPTS as an a-list in PR #1634

PR #1629, opened 2024-03-30
Redo simplify install of clipboard and wheelscroll

masinter:


nbriggs (commented):
That looks better.  While you're in AINTERRUPT you could fix the comments that misspell "more" in "No mroe room in GC tables" in the BLOCKRECORD INTERRUPTSTATE.

nbriggs:
This requires that you be running in a loadup made with the new AINTERRUPT -- otherwise LISPINTERRUPTS, the global variable, is NOBIND.  It will need a warning that the new WHEELSCROLL and CLIPBOARD are no longer compatible with older sysouts.

nbriggs:
Separate issue - and I'm not sure if this should be for Maiko or for WHEELSCROLL - the trackpad scroll direction is reversed - it doesn't honor the macOS preference for the direction to scroll.  I suspect this is because XQuartz doesn't honor the preference, so there should probably be a WHEELSCROLL option in the install to decide which way UP and DOWN should go.

rmkaplan:
I think I set it up so that the user can control the direction by setting the deltas positive or negative. I set WHEELSCROLLDELTA to -20 (default is 20), so that it conforms to the “natural” Mac scrolling. I suppose the Tedit file should say what the defaults are.

I ran the little Tedit test in a fresh loadup that included the new AINTERRUPT.

nbriggs:
If you load the new AINTERRUPT.LCOM, then WHEELSCROLL.LCOM then CLIPBOARD.LCOM:
copy from outside Medley, paste to IL exec is fine, paste to XCL exec flashes.  Scrolling generates the meta-foo characters into the XCL exec.

If you make a new loadup, and start from the new lisp.sysout, it appears to already include WHEELSCROLL - since it's a LISPUSERS package it probably should not be in the Lisp loadup but could be in the Full loadup.

Also, in the new Lisp.sysout, none of the wheelscroll works - just black boxes.  Paste doesn't work in either exec (flashing).  Rerunning (INSTALL-CLIPBOARD) makes the IL exec work, but the XCL exec still flashes.

nbriggs:
re-running (INSTALL-WHEELSCROLL) fixes the black-blobs in the IL exec, but TEDIT and XCL exec still get black-blobs.

Also, the second call to (INSTALL-WHEELSCROLL) returns results I didn't expect - I don't think the 159 NIL NIL etc should still be there...?

image

nbriggs:
Typing anything into a TEDIT window resets the wheelscroll state to not-working.  Redo (INSTALL-WHEELSCROLL) and it will work for a short period, or until you type something.  Possibly if you click outside and then back in a TEDIT window it'll stop working.

masinter:
Redid with a better PR
#1634

rmkaplan:
Clipboard, wheelscroll seem to work on Mac, including Tedit without changes.  Also works after restarting a sysout.

PR #1634, opened 2024-03-31
lmm57 interrupts clipboard wheelscroll

masinter:
- **Changes to interrupt initialization for CLIPBOARD and WHEELSCROLL**
- **fixed initialization**


nbriggs:
Better.  Interrupt settings don't come and go now.  Loading CLIPBOARD installs the paste operation in a pre-existing (XCL) exec window.  If you load TEDIT after loading the CLIPBOARD and WHEELSCROLL code you need to re-run (INSTALL-CLIPBOARD) in order for TEDIT to get the pastes (scrolling just works, nothing needs to be done).  LOGOUT and restart seems fine.

rmkaplan:
Should loading Tedit call INSTALL-CLIPBOARD?

nbriggs:
Try this, and see if you get the same result I do:
Do a new loadup then do, without any personal greet file, just the basic stuff:

```
./run-medley -g 1200x700 -sc 1200x700 -noscroll loadups/lisp.sysout
```
Then
    Leave the XCL exec alone, get a new Interlisp exec, and in the Interlisp exec do
   `(FILESLOAD WHEELSCROLL CLIPBOARD)`
then copy something outside lisp and try pasting it into the Interlisp exec (OK), and into the XCL exec (flashes).
then try using the scroll wheel in the Interlisp exec (OK, does nothing), and in the XCL exec (black boxes).

Is this what we expect?

nbriggs:
> Should loading Tedit call INSTALL-CLIPBOARD?

Not sure - it would check if INSTALL-CLIPBOARD is present, but perhaps it should be looking at things in the interrupt environment to see if copy/paste interrupt characters have been set up, and then do what it knows to be the right thing for TEDIT to do.

rmkaplan (approved):
4/1/24:  reviewed in meeting...
Better than current set up, although Tedit interface is not yet optimal

PR #1637, opened 2024-04-01
COMPAREDIRECTORIES makes directory filedates consistent with internal date of Tedit files

rmkaplan:
Like it does for Lisp source files.


masinter (approved):
lgtm

PR #1638, opened 2024-04-02
Default prc menu shows all open PRs, draft and nondraft

rmkaplan:


MattHeffron:
It seemed an interesting challenge to get the information in JSON, and then to convert it into lisp.
This probably will not behave well if the PR description contains a "
I think I know how to fix that if necessary.

MattHeffron:
> I’ve written a first-pass JSON parser (I think).

I found the [YASON](https://phmarek.github.io/yason/) implementation of a JSON parser/encoder in Common Lisp. I got the parser to load and compile. (The encoder uses CLOS heavily; I didn't try. I was able to _code-around_ the CLOS in the parser.)
When I looked again at `gh` using `jq`, I realized that the "GO templates" were what I used a lot in the `bibliography` processing, so it seemed easier to switch to what I was familiar with.

rmkaplan:
As we agreed on Monday, I am closing and deleting this PR in favor of the JSON approach in #1689

PR #1605, opened 2024-03-21
Add support for cl: loop for hash tables

masinter:
Adds to XCL-LOOP an addition for supporting iterating through a hash file.
Not a direct port, had some ANSI . CLtL2 updates (declare (ignorable --))
includes TEST-HASH-LOOP()

PR #1642, opened 2024-04-02
Fix bug in macroexpansion of loop with 'repeat N' clauses

masinter:
Will need to merge with other loop macro PR but can be evaluated indpendently


masinter:
i went ahead and merged the change to loop::repeat-clause in the file XCL-LOOP with the other changes in #1605 so will close this one.

PR #1647, opened 2024-04-05
Move PLOTANDNC-PATCH from obsolete back to lispusers since its used by Notecards

fghalasz:
Move PLOTANDNC-PATCH from obsolete back to lispusers since its used by Notecards - specifically NCPLOTCARD


masinter (approved):
LGTM

PR #1649, opened 2024-04-05
IMINDEX:  The displayfn of index image objects sets AFTERHARDCOPYFN to close the index file

rmkaplan:
Removes the need for advising the Tedit hardcopy function.  This won't have an effect until a separate PR (after rmk7 is merged) that updates the hardcopy function, but it is still compatible with currently released code.


masinter:
doesn't seem to work when using HCFILES which uses

image

rmkaplan:
By not working, you mean that the index file still stays open?

Yes, it won’t do the close until the renaming PR is merged, and then another one that I am holding off in order to not get dependencies screwed up.

The problem is that the release TEDIT.FORMAT.HARDCOPY extracts the AFTERHARDCOPYFN before it starts to display the image objects, so the fact that the different image objects set the property has no effect.

This PR is independent of other changes (also includes the readtable fix), but by itself it doesn’t solve the closing problem.

But if “not working” means something else is broken, that’s something I should look at.

masinter (approved):
THis works for me.

PR #1654, opened 2024-04-08
Working over HCFILES

masinter:


rmkaplan:
In HCFILES, why do you embed TEDIT.FORMATTEDFILEP in an NLSETQ.  Have you seen that cause an error, rather than returning NIL?

PR #1655, opened 2024-04-08
NSDISPLAYSIZES:  A better attempt at TERMINAL and TITAN

rmkaplan:
This now builds in knowledge of which Titan and Terminal fonts actually exist, so it doesn't try to increase the size to a black hole.  Still a hack, but there are so many MOVD's involved that it is really hard to see how to do it with a generic FIND-NEXT-LARGEST-FONT.


masinter (approved):
sounds good to me.

PR #1657, opened 2024-04-10
Remove derived PDFs from repo; now in 'src' repo, comstructed by HCFILES and MAKE-INDEX-HTMLS

masinter:
If we put the PDFs in the gh-pages branch, we don't have to carry them along whenever we clone 

we can set up some automation for auto-updating the gh-pages when the master copy of the source changes, of course.


rmkaplan (dismissed):
LGTM

rmkaplan (approved):
LGTM

PR #1658, opened 2024-04-10
Fixed a missed case in NSDISPLAYSIZES

rmkaplan:
FONTSAVAILABLE calls with SIZE = *


masinter (approved):
need better fonts, but until then

PR #1659, opened 2024-04-11
DINFO:  recreate textstream if it has disappeared from the window

rmkaplan:
Not sure how this can happen--a Tedit window without a textstream--but now testing for it.  DINFO must have removed it for some unknown reason.


nbriggs (approved):
LGTM

Issue #1667, opened 2024-04-17
TEDIT: Bitmaps in-line with text have width ignored in line formatting for PS printing


MattHeffron:
**Describe the bug**
When printing a TEDIT document to PostScript (or PDF), a bitmap `IMAGEOBJ` in-line with text is being overwritten by following characters in the line and the right margin alignment is _off_ by the bitmap width.

**To Reproduce**
1. Using full.sysout
2. TEDIT("STRESSTEST.TEDIT")
3. Right click menu: Hardcopy > To a file
4. Specify the file name. (For me this defaulted to "STRESSTEST.PS")
5. Use separate application to interpret the PostScript. (Or view PDF directly)
6. Near the bottom of the first page, there is a line: "the left mouse button. The caret shape (^) will appear. Type". The ^ is a bitmap `IMAGEOBJ`. The ") will appear. Type" prints as if the ^ was not there (overwriting it) and causing the right edge to be a bit left of correctly aligned. See screenshot below.

**Expected behavior**
Print formatted consistent with TEDIT display.

**Screenshot**
![image](https://github.com/Interlisp/medley/assets/8261874/83c0cf93-186f-40ae-be3b-839a0c60bf7c)

**Context (please complete the following information):**
 - OS: Win 11 WSL
 - OS Version: Ubuntu
 - Host arch: x86_64
 - Maiko version: ce20acca
 - IL:MAKESYSDATE: [ date ]

**Additional context**
Here is the PostScript code for generated for that line:
```postscript
25800 9420 M (the) S
( left mouse button.  The caret shape \() 60 0 32 4 -1 roll widthshow
/bitbltsave save def 44290 9420 translate 1056.0 480.0 scale 11 5 true thebitimage
60C0
3180
1B00
0E00
0400

bitbltsave restore
(\) will appear.  Type the ) 60 0 32 4 -1 roll widthshow
```
Here is the full PostScript file: 
[STRESSTEST.PS.TXT](https://github.com/Interlisp/medley/files/15018035/STRESSTEST.PS.TXT)

Note that it does not set a new position (no: **X# Y# M**) after the **bitbltsave restore** before the following text.
[...]

PR #1668, opened 2024-04-19
TEDIT-HCPY set X position after image object hardcopy display

rmkaplan:
Addresses issue #1667


MattHeffron (approved):
Almost identical to my version of the fix.
I'll remove my PR 1672

PR #1669, opened 2024-04-19
TEDIT-FILE  with TEXTOBJ argument passed to lower get-piece functions

rmkaplan:
An internal extension, so that error or warning messages if any can be displayed in the local prompt window


masinter (approved):
works for me

PR #1670, opened 2024-04-19
TEDIT-MENU: Eliminate illegal arg and caret location errors

rmkaplan:
@nbriggs got an error when (as I remember) parsing the OTHER font in the charlooks menu.  This was because of a missing BOUNDP check when trying to decide if the other font is a fontclass.

Separately, the caret appeared in the wrong place if a field insert was the first interaction with a menu.


nbriggs (approved):
Works for me.

PR #1672, opened 2024-04-19
Always update both X and Y stream positions after inline IMAGEOBJ is output

MattHeffron:
This seems to fix issue #1667.
I have tested only PostScript (and PDF) output, with only the one file mentioned in the issue noted.
It **ought** to be harmless; at worst, generate redundant positioning output.


MattHeffron:
Redundant with PR 1668.
@rmkaplan and I were independently working on it simultaneously.

PR #1673, opened 2024-04-19
Mainly scrolling and line updating when Tedit windows are partially offscreen

rmkaplan:
This is hopefully getting near the end of the rework of scrolling aimed at avoiding flicker with the fast repetitive scrolling that wheelscroll can produce.  The idea is to calculate what part of the screen bitmap can just be moved around, including the selection highlighting, so that only newly revealed lines need to be redisplayed.  The easier cases are when the window is totally on screen.  Next easiest, in previous versions, is when the bottom of the window is offscreen.  This aims at the hardest case, where the top of the window is not visible.

This also includes a fix to the misnamed \TEDIT.COPYINSERTFN that @nbriggs noticed, plus perhaps some other issues that have been reported that I now don't remember.

It does not address the problem that @pamoroso recently noticed, where type-in selection is getting screwed up (and another glitch where type-in overflow at the end of the window moves up too many lines).


nbriggs:
In this version (haven't checked older versions), when I open a TEdit window, type some text then type delete the blinking cursor disappears.  If I type another character the character appears and the cursor reappears and starts blinking again.
The whole window was visible at the time.

nbriggs:
If you place the TEdit window so that the left side of it is offscreen and then scroll it (I was using trackpad), then about as much as is off the left side of the screen doesn't scroll on the right hand side of the window.

nbriggs:
Now if you attempt to shift-select a SKETCH into the end of the document it gets further, but dies with
```
NIL
MOUSE/14(debug)BT!
CL::%%NOT-NONCOMPLEX-NUMBER-ERROR
IDIFFERENCE
\TEDIT.COPYINSERTFN
COPYINSERT
SK.COPY.BUTTONEVENTFN
\PROTECTED.APPLYA0001
\PROTECTED.APPLY
WINDOW.MOUSE.HANDLER
```
because NIL is not a number for the IDIFFERENCE.  
[select.txt](https://github.com/Interlisp/medley/files/15046260/select.txt)

masinter:
now that we're running on faster hardware, and scrolling is a user-interaction-driven update, why not handle scrolling by just REDISPLAYFN the text after you scroll?

nbriggs (commented):
I think this might fix Paolo's #1662 but the scrolling offscreen windows problem remains.

masinter:
@nbriggs at the meeting today 4/29/24 we came to a different resolution after I added you to the reviewer list. Run has some other fixes he's still working on and will close this PR and replace itwith another.

PR #1675, opened 2024-04-21
TEDIT.COMMAND.LOOP doesn't lose highlighting with clipboard meta chars

rmkaplan:
Should be a more general way of setting a don't-highlight readtable property, but this solves the problem


rmkaplan:
To be replaced with one that moves Tedit-specific stuff out of CLIPBOARD.

PR #1676, opened 2024-04-21
Rmk16  remove tedit special stuff from  CLIPBOARD, move it to TEDIT-COMMAND.

rmkaplan:
This also tries to put out a useful representation of image objects rather than just causing an error because an object is not a character code that can be printed to the clipboard stream.  It applies the object's preprintfn, if there is one, otherwise prints the name of the getfn.

This isolates the issue in one place, as a first approximation.  In the long run it might be possible to shadow the system clipboard with a local tmp file, and separately run the object's PUTFN to that file.  Then copy/paste could move objects (with binary data) as well as characters from one place in Medley to another--or from one Medley image to another.  We would need some way of keeping the clipboard and its shadow in sync as other applications or other Medleys make use of the clipboard.

If other consumers of the system clipboard respect some sort of convention for ignoring byte-runs, we could just put our additional information there and not deal with synchronization.  Unlikely.


nbriggs:
Looks as though rmk15 and rmk16 have a conflict over changes to TEDIT-COMMAND ?  I can't merge rmk16 into a test branch after I've merged rmk15.

rmkaplan:
I tried to close/delete rmk15, rmk16 is a replacement.

nbriggs:
Ah, I see the problem - the PR was closed but the branch wasn't deleted, you have to do that explicitly on the https://github.com/Interlisp/medley/branches page.  I've deleted the rmk15... branch.

masinter (approved):
there's lots more to be done with CLUPBOARD but this is fine.

PR #1679, opened 2024-04-23
WINDOWOBJ:  Add ENCAPSULATEDOBJP function

rmkaplan:
If provided to TEDIT.MAP.OBJECTS, this will return a list of all unknown image objects and their locations


masinter (approved):
Could you say more about what this is good for?
How should this be documented?
Where?
Release notes?

rmkaplan:
(TEDIT.MAP.OBJECTS TSTREAM  ‘ENCAPSULATEDOBJECTP  NIL T)

returns a list of objects in TSTREAM whose GETFN’s could not be found.  Also shows their locations, for TEDIT.NORMALIZECARET.

I don’t know where/if encapsulated objects are documented.

rmkaplan:
The second commit corrects the test for an encapsulated object. Despite its name, ENCAPSULATEDIMAGEFNS is not actually the imagefns, but a different background data structure.

Also, the Tedit example is also more complicated because, for whatever accidental reason, the first argument of the functional argument passed to TEDIT.MAP.OBJECTS is the character number, not the object.  TEDIT.MAP.OBJECTS is undocumented, maybe only used by Notecards and TMAX.  I'd like to change it so that the object is first, character is second (I think this oddity was noted in an earlier discussion).

masinter (approved):
harmless

PR #1683, opened 2024-04-26
cumulative updates for HCFILES

masinter:


masinter:
combined with previous HCFILES updates in #1657

Issue #1685, opened 2024-04-30
Recently introduced bug in PACKFILENAME


fghalasz:
Here is a PACKFILENAME call from most recent Master branch:
![master_now](https://github.com/Interlisp/medley/assets/1251539/82769726-c8c8-4c27-8b9e-383eee013146)
Note the empty top level directory

Here is the same PACKFILENAME call using the Master branch as of February this year (specifically, 2024-02-28)
![master_240228](https://github.com/Interlisp/medley/assets/1251539/2ebdd3e1-d956-4c24-b5b5-8e20c943bc5d)
The top level directory is correct here - kind of.  But if you then add a filename or subdirectory, the loss of the closing angle bracket gets papered over.

Assigning to Ron since I think he recently made some changes to PACKFILENAME.

nbriggs:
The current behavior is correct - the /medley is a *file* named medley not a *directory* named medley.

nbriggs:
```
2/13_ (PACKFILENAME 'HOST 'DSK 'DIRECTORY 'MEDLEY )
{DSK}<MEDLEY>
2/14_ (PACKFILENAME 'HOST 'DSK 'DIRECTORY 'MEDLEY 'BODY 'FOO.TXT )
{DSK}<MEDLEY>FOO.TXT
```

fghalasz:
Actually /medley in the Unix world could be either a file or a directory - can't tell from the pathname, as is the case for any Unix pathname that doesn't end in /.   This leads to an ambiguity when translating into {DSK}<> path notation since /medley could be translated equally correctly into either `{DSK}<>medley` or `{DSK}<medley>`.   To complicate things even more, until recently in /medley was translated into `{DSK}<medley` (without the closing bangle bracket).  Is this is even a well-formed pathname?

Note that most of the system seems to operate using the '{DSK}<medley` (no closing angle bracket) notation.
See the following examples:
![case1](https://github.com/Interlisp/medley/assets/1251539/82464624-d426-459e-a24b-076d8621ff4f)
 
![case2](https://github.com/Interlisp/medley/assets/1251539/607891ef-ab29-4017-bf0d-7ba99f7757b1)

![case3](https://github.com/Interlisp/medley/assets/1251539/410e7b0e-40fb-4e85-b820-2755269b9e71)

Also see the following:
![case4](https://github.com/Interlisp/medley/assets/1251539/90b2ad2e-8f74-4431-b56e-29d2b39dd75b)


[...]

PR #1686, opened 2024-04-30
Temporary fix in medley.sh to work around Issue #1685.  Applies to cygwin only.

fghalasz:
Temporary fix to workaround issues discovered by Steve Kaisler in cygwin start-up.  Permanent fix is probably the fix to Issue #1685 - bug in PACKFILENAME.


masinter:
For urgent issues:
* the @Interlisp/admin group can override the review requirements for any PR. 
* the default agenda for weekly implementor meeting is to review open PRs and those that needed override.

Issue #1691, opened 2024-05-02
GITFNS prc fails if pull request branch name contains parenthesis


MattHeffron:
**Describe the bug**
The `prc` command of GITFNS fails with a "Command failed: ..." error if any pull request branch name includes parens. For example, `rmk19--Shorter-directory-names-in-COMPAREDIRECTORIES-printout-(costmetic)` . 

**To Reproduce**
1. Files loaded GITFNS.LCOM
2. Type `prc` to Exec window.
3. INTERLISP-ERROR Break window opens with 
![image](https://github.com/Interlisp/medley/assets/8261874/9fbdceab-c756-4ac8-96f2-75eaa471232a)

**Expected behavior**
Display of menu of open PRs.

**Context (please complete the following information):**
 - OS: Windows 11, WSL, Ubuntu
 - OS Version: Ubuntu 22
 - Host arch: x86_64
 - Maiko version: [e.g. commit ID from `git log | head`]
 - IL:MAKESYSDATE: 30-Apr-2024

**Additional context**
See PR #1692 for potential fix (but may conflict with PR #1689)

rmkaplan:
I saw this when I first tried to create it with the GITFNS cob command. I made a branch without the parens, but I couldn’t figure out how to find the bad one to delete it.

Is this an issue in github, that it can’t have parens in these title strings?  Or do the parens have to be escaped ior quoted n the command string that GITFNS passes down.

Is this related to the JSON return value, or a separate issue?

MattHeffron:
The problem is the shell command line parsing. It doesn't like the unquoted parens. I verified this by entering the identical command to the shell and it gave a shell error. Just quoting the branch name in the command fixes this.
I've cancelled PR #1692, and I will create a new PR in a new branch based on PR 1689.

PR #1692, opened 2024-05-02
mth7--rmk5--GITFNS-defaults-to-all-PRs-including-drafts

MattHeffron:
Change parsing of JSON returned from gh, to lisp alist vs. plist.
 Added double-quotes around all (?) places where a branch name is given to GIT-COMMAND. This seems to fix issue #1691


rmkaplan:
I have some other changes to the prc menu interface in GITFNS, stacked up behind the current switch to JSON.  So we will have to resynchronize.

MattHeffron:
I'll separate out the JSON parsing bits from the branch name quoting.

PR #1695, opened 2024-05-03
GITFNS prc menu is added to the screen instead of just popping up

rmkaplan:
So the full set of available PRs (with their draft/approved status) remains visible.  You can poke from one to the other. May be useful when we go through the PRs in Monday meetings.  (Still doesn't bother with a menu at all if there is only one PR to work on.)

On this first pass, the menu is placed (but not attached) just under the current exec window, on the hope that this keeps it out of the way of the windows that are created during the comparison.

Try it, let me know how this feels, whether this is an improvement over the previous pop-up.


rmkaplan:
This does include the JSON changes in PR #1689.  After this is merged, 1689 can be deleted.

(Note:  I don't see an override option for me to approve on my own)

MattHeffron (approved):
LGTM

PR #1696, opened 2024-05-04
UNPACKFILENAME produces lt-gt for top-level directory

rmkaplan:
For virtually any combination of leading <, > or / that is not followed somewhere by another bracket.  Addresses   #1685.

Please test the various cases, including round-trips through PACKFILENAME.


masinter (changes requested):
@rmkaplan will update

nbriggs:
@masinter - what changes/update are you asking @rmkaplan to make?

rmkaplan:
Here is what I understood from the meeting:

 The empty directory should be indicated by < instead of <>, along the lines of unix /.

All directory-internal closing brackets (> and /) should be reduced to a single >, as if they were /.

An open question whether directory-internal open brackets < should be treated as / and reduced, or cause an error.

rmkaplan:
I switched <> to <, as requested.  The duplicate-reduction (or error) for screwy combinations of directory-internal brackets still to be done.  But this addresses the original request for  representing the top-level directory, should be approved.

masinter (approved):
as per 5/13/24 meeting

PR #1698, opened 2024-05-06
Overhaul of the medley.sh script

fghalasz:
This PR is a major overhaul of the medley.sh (or ./medley) and associated scripts that can be used to start Medley.

The major changes are:

1.  The scripts are all Posix-compliant (as attested to by shellcheck) and use #!/bin/sh instead of #!/bin/bash as they previously had.  Tested so far on systems where /bin/sh is zsh, dash, and bash.

2.  There is now a config file that can be used to "prespecify" command line arguments.  The config file consists of command line arguments, one argument (or argument pair - flag plus value) per line. The config file is read before processing the command line arguments and any command line argument overrides what's in the config file.

    Only a single config file is processed either from MEDLEYDIR/.medley_config (allowing for per installation configs) or from $HOME/.medley_config (allowing for per user configs).  Also a --config command line argument has been added allowing you to specify an alternative config file location (or to ignore any config files). 

    Finally, where necessary command line arguments have been augmented so that you can "nullify" an argument set in the config file from the command line.  Example if "--noscroll" is set in the config file, then you can use "--noscroll -" on the command line to override it.

3. Explicit Docker on Windows support has been removed from the medley.sh scripts.

4. Some new arguments added:  --nethub, and --continue.  Also --nofork and --maikoprog to support loadups.

5.  medley.sh is implemented as a tree of sourced scripts.  This tree is now compiled into a single script before it is deployed.  This allows meldey.sh to run marginally faster.

6.  medley.sh no longer uses run-medley.  Instead it locates and calls maiko (lde or ldeinit) directly.

The new medley.sh has been tested on Ubuntu running dash and bash-v5.6.1, on MacOS running zsh and bash-v3.2 and on Cygwin running bash-v5.2.21.


fghalasz:
@rmkaplan   I added your feature request to the new medley.sh scripts.  Specifically, you can now create a numbered id - which is an id that ends in a '+' - including a singleton "+".  The plus will be replaced with a number unique among running medleys with the same base id.  

I also added to the title argument that a "%i" in the title will be replaced by ":: id".  And the default title is now "Medley Interlisp %i".  

So if you invoke `./medley --id + --title "This Title %i"`, then the window title would be, say, "This Title :: 5".

Or `.medley --id lfg_+ --title "LFG Medley %i" `, would result in the window title "LFG Medley :: lfg_3".

For your workflow, you should probably always use a numbered id - even if its just a +.  Without a numbered id, the medley script will prevent two medleys with the same id (including no id) from running simultaneously.

rmkaplan:
Thanks!

fghalasz:
Closing since this PR was merged as part of PR#1699 - minor loadups update.

PR #1699, opened 2024-05-06
Minor overhaul of loadup scripts

fghalasz:
NOTE:  This PR incorporates PR#1698.  Thus PR #1698 should be approved and merged before this one.

This PR is a minor overhaul of the loadup scripts including:

1.   Better handling of exit codes so that when one step in loadup-all, loadup-full, or loadup-db fails, it is recognized as a failure immediately and the remaining steps are skipped.

2.  By default the loadup scripts now use (the new) medley.sh instead of run-medley to start up Medley.  But added a command line argument (-os or --oldschool) that will use run-medley instead of medley.sh.

3.  All scripts are now Posix-compliant (as per shellcheck).  Previously, there were some non-Posix lines even though all scripts used #!/bin/sh.

Tested on Ubuntu running dash as /bin/sh and MacOS running zsh as /bin/sh.


masinter (changes requested):
loadup-init fails.

fghalasz:
@masinter   The problem in loadup-init.sh is the result of a bug in command line sysout argument processing in Maiko.  I will soon submit a PR for Maiko to fix said bug.  But meanwhile, I put a workaround into medley.sh to get around the Maiko issue.  loadup-init.sh should now be working?

masinter (approved):
Works for me. I'm not sure about the .medley_config option, but that's a new feature.

nbriggs:
@fghalasz - what's the bug in the maiko command line processing for sysouts?  If the sysout is specified on the command line for lde/ldex/ldesdl it is required to be the *first* argument, preceding all the options, not the last argument as run_medley presented it.

fghalasz:
Under the right circumstances on X Windows, the sysout as first argument is
never processed - the code in xrdopt just crashes out before the code in
main can get to processing the first argument.  See the exact circumstances
in Issue#1703.

Issue #1701, opened 2024-05-08
Peculiarities with printing(BOUT) to a Tedit stream


rmkaplan:
A Tedit stream has 2 personalities, as the basis for interactive text editing and as a Medley file stream.  I don't think all the details of their behavior and how they interact were worked out in the Venue years.

On the one hand a Tedit stream is set up as a normal text editor that supports selection, insertion, deletion, looks-changes etc.  Typically those are specified by on-screen keyboard actions, but there are also public functions (TEDIT.INSERT, TEDIT.DELETE, TEDIT.GETSEL, TEDIT.SETSEL and others) that allow programs to manipulate the stream whether or not it is on screen.  (If it is onscreen and those functions are called, the screen is updated to reflection the changes).

On the other hand, the Tedit stream acts (mostly) like a Medley file stream.  You can read from it, write from it, move the file pointer around, etc.   Importantly, Tedit using the stream interface internally for line formatting and display, as changes are made through the editing interface.  Also for hardcopy.

The behavior we inherited from Venue leaves some things to be desired.

1.  The Tedit BOUT doesn't actually act like BOUT's on other file streams.  Normally, if BOUT is called when the file is positioned at a particular byte in the middle of the stream (GETFILEPTR), the character at that position is replaced by the new character.  But Tedit's BOUT currently doesn't do an insertion, it does a replacement.  It inserts the new character in front of the character at that position, everything at that position and beyond is pushed back, and 1 is added to the length of the stream.  (There is no difference in the typical case when appending to the end of the stream.)

The current behavior seems wrong.  OK to change it from insert to replace?

2.  Edit insertions are typically undoable.  Since BOUT is currently going through the standard character inserter, the bouting is also undoable. The history list will get very long, and every sequence of BOUT's will be undoable possibly as a single event.  Is this a good idea?

3. Tedit treats image objects as pseudo-characters, with respect to most internal operations.  This leaks over into the stream interface, in that BIN will return an image object if it is the next character at the file ptr.  Should BOUT also be able to install an image object, to complete the analogy?  Currently a separate object-inserting functions is need to insert an object.

4.  If the stream is being displayed and a character is bouted, should the screen update?  What if a character is bouted (insertion or replacement) in a part of the stream that is not currently visible.  Should the display scroll to show where the change happened?  That's what happens if the caret is off screen and a character is typed.

[...]

PR #1702, opened 2024-05-08
Fixes recently identified glitches, eliminates some code duplication, removes inessential STREAMHINT back references

rmkaplan:
This replaces PR #1673, which should be closed and deleted when this is merged.

BOUT (printing) should work better (but see issues in #1701), Move should be properly undoable (and undo-undoable), offscreen scrolling on the left should not garble the display.

The STREAMHINT xpointer field is still present but now only for its essential uses in line formatting and display.  Those still need to be examined.

Remnants of an old strategy for screen updating were removed, hopefully will make it easier to fix lingering bugs.

Note that typing at the bottom of the screen still scrolls to the top rather than going up just a big to reveal the new caret line--next time.

With respect to undoing Move, undoing where the source and target are in the same document puts that document back to its previous state.
But if the source was taken from, and deleted in, a different Tedit document, undo doesn't undo the source deletion. If that document had been changed after the move, the deleted material would be inserted at some arbitrary place.
(It would be easy to detect whether the source document is in its original state so that the deletion could be sensibly undone.  But that would require back-pointers between separate documents, causing either a circularity (using pointers) or the possibility of smashing a previously collected and perhaps reallocated structure (with xpointers).)


masinter:
closed by mistake

rmkaplan:
This is superceded by a major (5th round) release.

Issue #1704, opened 2024-05-10
CYGWIN install fail


masinter:
Downloaded 
medley-full-cygwin-x86_64-240502-9f047593_240211-ce20acca.exe
from
https://interlisp.org/software/install-and-run/windows/native/

and going through the dialogs got 

![image](https://github.com/Interlisp/medley/assets/1116587/09c8bcfb-6dcb-4d89-9b65-0ebfced56756)

fghalasz:
Fixed in PR#1707

PR #1707, opened 2024-05-10
Fix to Issue #1704. Cygwin Install fail

fghalasz:
In cygwin installer, change cygwin download site to https://mirrors.kernel.org since old gatech site is deprecated.

See draft release medley-240510-80f56456 for working installer.


masinter (approved):
as per 5/24/24

nbriggs (approved):
LGTM.

PR #1709, opened 2024-05-12
Update medley man page with latest updates to medley.sh

fghalasz:
Update medley man page with latest updates to medley.sh


masinter (approved):
lgtm -- doc only

PR #1710, opened 2024-05-12
medley.sh: Fix handling of '--title -' in medley_run.sh and medley_args.sh

fghalasz:
Minor fix to medley.sh: Fix handling of '--title -' in medley_run.sh and medley_args.sh.


masinter (approved):
per 5/13/24 meeting

PR #1712, opened 2024-05-13
Fix Issue 1711: problem with "-x ~" in new medley.sh

fghalasz:
Fixed two typos in medley_utils.sh:  $%2 -> $2


MattHeffron (approved):
Just what I thought...

Issue #1715, opened 2024-05-17
MANAGER doesn't print CL docstrings


pamoroso:
**Describe the bug**
The MANAGER LispUsers module doesn't print the docstrings of Common Lisp functions. The output of the MANAGER command `Documentation`  in the `Manager Command Activity` window contains only this header but not the docstring itself:

```
Documentation for [FUNCTION-NAME]:
------
```

**To Reproduce**
1. sign into your Interlisp Online account
2. under `Initial Exec`, select `Interlisp`
3. click `Run Medley`
4. at an Interlisp Exec, evaluate `(FILESLOAD MANAGER)`
5. at a XCL Exec, load a Common Lisp file that contains functions with docstrings such as [INSPHEX](https://github.com/pamoroso/insphex/blob/main/INSPHEX)
6. from the background menu, execute the command `File Manager`
7. from the `Manager` menu, select the option `INSPHEX` > `FUNCTIONS`
8. from the `FUNCTIONS on INSPHEX` menu, click `HEXDUMP`
9. from the resulting menu, select the option `Documentation` > `Documentation`

**Expected behavior**
The `Manager Command Activity` window shows the docstring after the header.

**Screenshots**
The `Manager Command Activity` window:

![manager-documentation1](https://github.com/Interlisp/medley/assets/10342479/b2605510-87eb-40a7-94e0-71fb32b6ec8f)

The full desktop:

![manager-documentation2](https://github.com/Interlisp/medley/assets/10342479/bac1cbb5-5cac-4910-9df1-fcec1bb6fb7b)

**Context (please complete the following information):**
 - OS: chromeOS Stable
 - OS Version: 124.0.6367.154
 - Host arch: x86_64 (ASUS Chromebox 3)
 - Maiko version: N/A
 - IL:MAKESYSDATE: 15-May-2024 11:18:02

**Additional context**
Nothing obvious in the MANAGER [code of the `Documentation` command](https://github.com/Interlisp/medley/blob/1134cb1ce6aa7e215f80dda8b4814140a059147c/lispusers/MANAGER#L663) jumps at me. Maybe `CL:DOCUMENTATION` is printing the output to the wrong stream?

Evaluating `(documentation 'hx:hexdump)` at a XCL Exec prints the docstring as expected.

pamoroso:
Closed as fixed by PR #1716.

PR #1716, opened 2024-05-17
Fix 1715 - "structural typo" in output of MANAGER Documentation menu item

MattHeffron:
Fix issue #1715 
`CL:DESCRIPTION` value was being dropped, not printed.


masinter (approved):
This is fine as it is. There is an opportunity to merge the behaviors of HELPSYS and SEDIT Meta-H and Meta-D and GRAPHER (a masterscope extension) to possibly unify the various kinds of help/expansion. Arglist, docstring, IRM, CLHS, in-repo TEdit and/or derived PDF. For another discussion/issue.

PR #1718, opened 2024-05-19
Further updates to loadup scripts to make its easier to use Lisp-side copy functions to copy medley directories

fghalasz:
Update all loadup scripts so that 3 things are true:

1. all sub-scripts are called via `/bin/sh <subscript>` so that exec bit need not be set on any loadup script

2. medley is called via `scripts/medley/medley.command` rather than just `medley` so that the top-level medley symbolic link is not necessary for loadups

3. the loadup scripts do not have to be started while cwd is MEDLEYDIR;  instead MEDLEYDIR is automatically computed based on where the loadup script is executed from (as the medley script does).


hjellinek (approved):
Looks good to us.

PR #1722, opened 2024-05-20
Change Medley release builds from Mon 2:00a PT to Wed 2:00a PT ...

fghalasz:
... as per discussion in implementors mtg a few weeks back.


masinter (approved):
as agreed

PR #1723, opened 2024-05-21
gather-info analyzes with reading, HCFILES rename dest

masinter:
GATHER-INFO: run Masterscope while reading sources  rather than reading all and then analyzing all.
HCFILES: rename destination to be `<full source name>.pdf` for all file types.


masinter:
the .pdf changes to HCFILES were redone in PR #1776, and I'm having second thoughts about the GATHER-INFO changes because they don't really help. So closing this PR.

Issue #1719, opened 2024-05-19
MANAGER doesn't print function definitions


pamoroso:
The MANAGER LispUsers module doesn't print function definitions. The output of the MANAGER command `File Manager` > *`FILE-NAME`* > `FUNCTIONS` > *`FUNCTION-NAME`* > `PrettyPrint` > `Function Def`  in the `Manager Command Activity` window contains only this header and the function definition is missing:

```
Function definition of [FUNCTION-NAME]:
------
```

This is similar to issue #1715 but located in the [function definition printing code](https://github.com/Interlisp/medley/blob/6c47d75ab94d6676bf3dad203031436c37424bf5/lispusers/MANAGER#L649). The cause of the bug, a "structural typo", is likely similar as well as the fix in #1716 by @MattHeffron.

MattHeffron:
Fixed in PR #1726

PR #1726, opened 2024-05-22
Fix issue #1719.

MattHeffron:
MANAGER called PF with function NAME in a variable. BUT PF is NLAMBDA so didn't eval its argument. Changed to use CL:APPLY*. Moved the process function from a local LAMBDA in Manager.DO.COMMAND to a separate named function. Tweaked prompt strings for the "CopyDef" and "Rename All" item menu commands.


pamoroso (approved):
Looks good to me.

masinter:
I was hoping to rationalize the use of NLAMBDA nospreads  vs exec package independent "commands" for cl users

Issue #1727, opened 2024-05-22
Editing function from MASTERSCOPE show paths graph locks MOUSE process


MattHeffron:
**Describe the bug**
If you use the left mouse button on a node in a MASTERSCOPE show paths graph, the popup menu offers "Edit". Selecting that locks up the mouse process. Using the middle mouse button to directly edit the function doesn't have this problem.

**To Reproduce**
1. Files loaded 'MANAGER, MASTERSCOPE, GRAPHER, and BROWSER'
2. In INTERLISP Exec, type: `ED(PRETTYCOM)`  answer yes to loading PRETTY `PROP`
3. Exit SEDIT of PRETTYCOM
4. Left mouse click on PRETTY in MANAGER main menu. Select "MasterScope" (don't use SUBMENU)
5. In INTERLISP Exec, type: `. SHOW PATHS FROM PRETTYCOM TO EDITLOADFNS?`
6. Middle mouse click on a `PRETTYCOM` node in the graph.
7. Note that it opened SEDIT without locking up access to the mouse.
8. Left mouse click on a `PRETTYCOM` node in the graph. Select "Edit" from the popup menu.
9. Notice that the mouse cursor still tracks, but you cannot click on anything.
10. Type ctrl-B to enter a BREAK. This spawn's a new mouse process, so clicking works fine now.
11. Type OK in the BREAK. Everything proceeds well.

**Expected behavior**
Menu selection doesn't lockup the mouse process.

**Screenshots**

**Context (please complete the following information):**
 - OS: [e.g. Mac/Linux/Cygwin]
 - OS Version: [e.g. High Siera/Ubuntu 18/Raspbian]
 - Host arch: [e.g. x86_64, arm7l, arm64, sparc]
 - Maiko version: [e.g. commit ID from `git log | head`]
 - IL:MAKESYSDATE: [ date ]

**Additional context**

PR #1728, opened 2024-05-22
Edit of function from popup menu in MASTERSCOPE show paths graph locks mouse process

MattHeffron:
Fix issue #1727
Edit of function from popup menu in MASTERSCOPE show paths graph locks mouse process.
Left buttonfn didn't include :DONTWAIT option when launching the editor; middle buttonfn did.


hjellinek (approved):
LGTU

PR #1733, opened 2024-05-24
Fix Issue#1730: Break during OPENSTREAM on file on XNS Server

fghalasz:
Add a DEVICE argument to \NSRANDOM.CREATE.STREAM so that when it creates a stream, it can pass a device to the stream creation code.  As of PR814 of July 2022, streams whose DEVICE field is NIL break during creation.  Update all callers to \NSRANDOM.CREATE.STREAM to pass the new DEVICE argument.


rmkaplan (approved):
LGTM

Issue #1735, opened 2024-05-25
Question mark (?) command fails to find documentation for named argument


MattHeffron:
**Describe the bug**
The ? command alone displays documentation for all commands. If given a FUNCTION or COMMAND name, it reports "No documentation found". (I did not check with other definition types that have documentation strings.)

**To Reproduce**
1. In Exec type: `? see` [enter]
2.  `No documentation found` is displayed
3. In Exec type: `?`[enter]
4. The display will show documentation for all defined commands, including "see"

**Expected behavior**
The documentation for the named argument is correctly displayed.

**Context (please complete the following information):**
 - IL:MAKESYSDATE: 15-May-2024 11:18:00

MattHeffron:
See also Issue #1725

masinter:
There is a problem with the “name” of a command in that commands are case and package independent but the file manager doesn’t know that.

So while the symbols SEE, see, CL:SEE and the strings "see" and "SEE" can be used to invoke the command,
any of them, and any of them can be used in a DEFCOMMAND, the file manager and whereis don't take that into account.
We should pick ONE of the choices and make others coerce.

Without getting too far ahead of the design issues around commands vs.functions, I would propose that we name a command with the INTERLISP package symbol all uppercase as the 'canonical' way of defining a command i.e. the DEFCOMMAND definer would (IL:MKATOM (IL:UCASE NAME)) and similarly the COMMANDS and WHERIS database.

This is a general problem but in this case the ? command in the CL Exec seems to pick up on xcl-user::?

masinter:
The command `? see` should work in the XCL exec too.

Issue #1725, opened 2024-05-21
MAKEFILE of file with edited function includes 2nd copy of COMS, but modified!


MattHeffron:
If I edit `PRINT-ALL-DOCUMENTATION` in CMLEXEC, doing a MAKEFILE (or CLEANUP) includes in the new file, a **_modified_** copy of the `CMLEXECCOMS`. 
I `(LOAD 'CMLEXEC 'PROP)` before editing the function.
How should this be done to avoid ending up with the modified COMS? 
The modification seems to be happening **during** the MAKEFILE.

masinter:
What I remember is that this is expected behavior if the MAKEFILE encounters a misspelled filecom and spell corrects, or if NLAML NLAMA LAMA are not set because the appearance of functions on those lists determine in some Interlisp implementations how calls are compiled. This is old behavior from Interlisp-10 but LAMA and NLAMBDAs are compiled differently than calls to LAMLs. 

I hope this is all in the IRM but I haven't checked. What is the change to the fileCOMS that you notice?

MattHeffron:
@masinter The change to the COMS is adding 7 FNS to the LAMA variable.

masinter:
i'm not sure if the CL:COMPILE-FILE DFASL compiler vs. BCOMPL/TCOMPL and CL:EFUN vs DEFINEQ  of  CL:LAMBDA, and &KEY &REST but at some point the behavior is intended if unexpected.

MattHeffron:
See also Issue #1735

PR #1736, opened 2024-05-25
Fix PRINT-ALL-DOCUMENTATION locating documentation string

MattHeffron:
In PRINT-ALL-DOCUMENTATION only one of two places was checked for the documentation string.
(Also, MAKEFILE updated the LAMA variable in the CMLEXECCOMS.)
Issue #1735


sources/CMLEXEC
```diff
@@ -1737,26 +1742,28 @@ Copyright (c) 1985-1988, 1990-1991, 1993, 2021 by Venue & Xerox Corporation.
 
 (ADDTOVAR NLAML )
 
-(ADDTOVAR LAMA )
+(ADDTOVAR LAMA PROCESS-EXEC-ID PRINT-EVENT PRINT-HISTORY EXEC-PRIN1 EVENTS-INPUT EVAL-INPUT EXEC-READ)
+)
+(PUTPROPS CMLEXEC COPYRIGHT ("Venue & Xerox Corporation" 1985 1986 1987 1988 1990 1991 1993 2021 2024)
```
hjellinek, line 1747: Is there a way to override the "Venue & Xerox Corporation" attribution to include Interlisp.org?

rmkaplan:
In the technical meeting today, we focused on the LAMA specification in the files COMS.  It seems very odd (as in I have never seen it) that a file has explicit settings for the LAMA, NLAMA... variables, those were always automatically supplied and properly updated  by the MAKEFILE.

Is there something special about this particular file?  Or are these variable settings just a historical mistake that is being carried forward.  What would happen if this command is simply removed from the file?

(We noted that this file is compiled by FAKE-COMPILE-FILE, if that is somehow relevant to this question.)

sources/CMLEXEC
MattHeffron: Yes, but it'd probably be better to define a change that we want to apply to ALL of the files that are now under our "control", instead of doing them one at a time.

MattHeffron:
@rmkaplan The changes to LAMA are the subject of my Issue #1725.
In this case, instead of having the file contain two instances of the COMS; the original, and the one updated by the MAKEFILE, I did a few steps:
1. I saved the file with `MAKEFILE` writing the COMS twice. 
2. I exited the sysout and started a clean FULL sysout.
3. I loaded the file `'PROP` (At this point the COMS were updated.)
4. I marked `PRINT-ALL-DOCUMENTATION` as changed.
5. I did a `CLEANUP` of the file. (Only the single, updated, COMS are written.)
6. Created the Pull request.

sources/CMLEXEC
MattHeffron: @hjellinek See the discussion with @masinter on my draft PR #1527 from February.

hjellinek: Yes, this issue I mentioned in my comment would be superseded by a solution to trust one.

MattHeffron:
I don't know if removing the 
```newlisp
(DECLARE%: DONTEVAL@LOAD DOEVAL@COMPILE DONTCOPY COMPILERVARS 
(ADDTOVAR NLAMA DIR)
(ADDTOVAR NLAML )
(ADDTOVAR LAMA PROCESS-EXEC-ID PRINT-EVENT PRINT-HISTORY EXEC-PRIN1 EVENTS-INPUT EVAL-INPUT EXEC-READ)
)
``` 
from the `COMS` would have any effect, good or bad. It seems simplest just to leave it and include `MAKEFILE`'s update.

rmkaplan:
It does seem that MAKEFILE adds/updates these commands automatically as needed in all the managed files, so I think leaving them is the right thing to do.

fghalasz (approved):
Approved as per Implementors Mtg on 2024-06-03

Issue #1742, opened 2024-05-29
<lspusers>MANAGER has call to FB.FASTSEE.ONEFILE whidh errors ARG NOT FILEBROWSER


masinter:
<img width="404" alt="Screenshot 2024-05-28 at 7 56 28 PM" src="https://github.com/Interlisp/medley/assets/1116587/7876d55e-0e64-4c2f-b3fa-95f0e9b3f23c">

PR #1743, opened 2024-05-29
Fix FILEBROWSER error when a borrowed function needs the prompt window.

MattHeffron:
FB.PROMPTW.FORMAT now uses the global PROMPTWINDOW if called under a borrowed function (i.e., not a FILEBROWSER window). Tweaked the format string provided by FB.FASTSEE.ONEFILE to go to newline conditionally.
Issue #1742


fghalasz (approved):
Approved as per Implementors Mtg on 2024-06-03

PR #1752, opened 2024-06-07
Update buildReleaseIncDocker workflow to automatically kick off a buildAndDeployMedleyDocker workflow in the online repo

fghalasz:
With this change, whenever a new Medley release and docker image is built successfully (e.g., every Wed morning) then a new online.interlisp.org Medley image will be built and deployed to the online.interlisp.org server.


masinter (approved):
tested by fgh

Issue #1749, opened 2024-06-05
type treatment of NIL does not follow CLtL2


eblanton:
Medley XCL as of [maiko 3ef7d2d](https://github.com/Interlisp/maiko/commit/3ef7d2d68ecc8ff9ca839c5c068e7d7ce7768583) and [medley 1148cd59](https://github.com/Interlisp/medley/commit/1148cd594554ad2cfcdb23a45a683c214424c9a9) returns CONS for (type-of '(1 2 3)) (expected), but SYMBOL for (type-of '()), which surprised me (SBCL returns NULL).

This issue is being filed in response to https://github.com/Interlisp/medley/issues/609#issuecomment-2148985306 and is related to #609.

This may or may not be a problem, but it _is_ associated with a divergence from CLtL2 behavior with respect to `typep` and `subtypep`.  CLtL2 (p.66, Section 4.9, "Determining the Type of an Object") says:

> * Let _x_ be an object such that `(typep` _x type_`)` is true and _type_ is one of the following:
>
> [...] `null`
> 
> Then `(subtypep (type-of` _x_`)` _type_`))` must return the values `t` and `t`; that is, `type-of` applied to _x_ must return either _type_ itself or a subtype of _type_ that `subtypep` can recognize in that implementation.

However, on Medley, I get:

```
2/40> (type-of '())
SYMBOL
2/41> (type-of nil)
SYMBOL
2/42> (typep '() 'null)
T
2/44> (subtypep (type-of '()) 'null)
NIL
T
```

This final result should be `T T` as dictated above.  I believe `(type-of nil)` should probably be `NULL`, although as far as I can tell the standard does not dictate that this is the case (all responses to `type-of` are permitted to be "implementation-dependent" if the rules on p. 66 are followed; as far as I can tell, all rules except the above are honored, although the `class-of` rule is not relevant without CLOS, which is not loaded in Medley by default (and I have not tried it).

PR #1753, opened 2024-06-07
Fix issue #1749 - type-of NIL doesn't match CLtL2

MattHeffron:
Adds a "special case" to `CL::SYMBOL-TYPE` so `(type-of NIL)` behaves according to CLtL2.
Reported in Issue #1749


sources/CMLTYPES
```diff
@@ -394,10 +398,13 @@
...
-   (IF (KEYWORDP SYMBOL)
-       'KEYWORD
-       'SYMBOL))
+(DEFUN SYMBOL-TYPE (SYMBOL)                           (IL:* IL:\; "Edited  4-Jun-2024 23:23 by mth")
+   (COND
+      ((NULL SYMBOL)
+       'NULL)
+      ((KEYWORDP SYMBOL)
+       'KEYWORD)
+      (T 'SYMBOL)))
```
eblanton, line 407: This appears to be correct and match the expectations in CLtL2 2.15 (Overlap, Inclusion, and Disjointness of Types) on p.39, which states "The type null is a subtype of symbol; the only object of type null is nil" (the first part of this statement was already true) and p.66, which states that `(subtypep (type-of nil) 'null)` must return `T T`.

eblanton (commented):
The commented hunk is correct and complete.  All other changes in this PR are whitespace or editor/files package artifacts.  If project policy is to allow committing arbitrary formatting and metadata changes with code, I believe this change can be applied as-is and will resolve #1749.  If project policy discourages this, the size of this PR diff should be reduced appropriately.

sources/CMLTYPES
MattHeffron: Yes, the changes due to the file being rewritten by `MAKEFILE` (including as from within `CLEANUP`) are considered as _ancillary_ and are acceptable.

nbriggs: Since there’s no way on GitHub to control its diff interpretation of “whitespace”, such as the font shifts, and prettyprint layout changes, we use a lisp based set of tools in medley to review changes to lisp source.

eblanton: I'd be interested to know about that tool, it was a lot of scrolling through that diff to find the real change!  Is there a process guide here somewhere?

rmkaplan:
man GITFNS

should show you the documentation.  You should be able to get it by (LOAD 'GITFNS.LCOM).

The command you want is prc (pull-request-compare).

MattHeffron:
Decided that we need to find and verify that nothing depends on the old behavior before merging.

MattHeffron:
@masinter While checking for possible changes in behavior due to the change in `cl:type-of`, there are a couple of potential issues in the file **XCL-LOOP**. I can make the edits, but cannot save the file. `MAKEFILE` chokes because the `COPYRIGHT` property on `IL:XCL-LOOP` is not the expected structure. This is already the case in the file as it already exists in the repo. 
How did you get `MAKEFILE` to write that file?

masinter:
I run with SYSEDIT which currently sets COPYRIGHTFLG to NEVER.
I imported the code from another place which had a copyright notice and I jusst stuffed it in the COPYRIGHT property, which is OK for LOAD but if you do a MAKEFILE with a different value of COPYRIGHTFLG...

I should do more research into what other similar projects do. Or we could ask the SPN "Law and Policy" working group for a consultation?

In the meanwhile, I recomend setting COPYRIGHTFLG to NEVER and MAKEFILE away.

sources/CMLTYPES
masinter: I thought I had seen a note (possibly email to @eblanton ?) 👍🏽 
THe tool is (in lispsuers). in the file GITFNS.

(FILESLOAD GITFNS). should load it in. The Exec command "prc (either case and in any package) will bring up a pull-request menu.

masinter (approved):

PR #1756, opened 2024-06-11
For testing GITFNS before and after mth17

MattHeffron:
Just for testing upcoming changes to GITFNS.
**DO NOT** approve this pull request, until my subsequent one that should fix the GITFNS problem has been approved.

PR #1757, opened 2024-06-11
This fixes GITFNS prc not showing any changed files with certain filename/path patterns.

MattHeffron:
This fixes GITFNS `prc` not showing any changed files if they included `.git` in the full name.
It now is specific, and excludes only those in the `.git/` top-level directory.
GITFNS also ignored changed files at the top-level of the repo. This is fixed also.
This can be tested with PR #1756


rmkaplan:
When I run prc and click on this pr in the menu, I get an "unknown branch origin/mth17... error.

rmkaplan:
Actually, maybe in the wrong version.  Let me check...

rmkaplan:
OK, it works for prc on the test pr 1756, but does give the "unknown branch" error on 1757.  Error is inside GIT-BRANCH-EXISTS?

MattHeffron:
@rmkaplan It turns out that .git in the branch name caused the same issue of the branch unseen.
I believe that this version is correct, but I may have missed some cases where it should have allowed .git in a branch name, or allowed .git where it shouldn't have.

rmkaplan (approved):
Mostly just passing the ALL flag, plus specifing top-level .*.  LGTM

PR #1762, opened 2024-06-14
AFTERHARDCOPYFN not needed in IMTEDIT

rmkaplan:
It is put in the first time the indexing image object is displayed, this was left over from earlier cleanup


masinter (approved):
LGTM -- simple enough :)

Issue #1141, opened 2023-04-06
Fix errors discovered by GATHER-INFO (in Medley Utils)


masinter:
I wrote GATHER-INFO to gather some information about the state of Medley repository sources.
The GATHER-INFO operates in 5 phases: 0 is setup, 1-4 do gross things to analyze the sources.
GATHER-INFO(-1 thru -3 prints out the results. 
Yes, this is ugly.

- [ ] GATHER-INFO(-1) shows "loaded files not in SYSFILES or FILELSTS": these are files that were loaded at some point, added to LOADEDFILELST (full pathnames) but the root name wasn't remembered. THESE FILES DONT GET MASTERSCOPED in fuller.database.  Why this happens? anyway, 5-6 files that should get analyzed too.
- [ ] Sources not loaded:  Files not in the sysout but in the <sources> directory. A motley collection of old obsolete and unused things and newer, as well as some DECLs files that are only loaded when you're working on the suite.
- [ ] Files in FILESETS not loaded -- just one, POSTLOADUP.
- [ ] phase 2: GATHER-INFO(2) maps through all of the compiled code of a full.sysout (using MAPATOMS) looking for functions called but not having a definition. Some of these are the "renamed" version used by MAKEINIT ... functions that start with I. or R.  ... a long motley list. 
- [ ] Phase 3: The prep for this is time-consuming. It LOADs every file (about 388) and does (LOAD file PROP). It also uses (READFILE) to read in all the S-expressions and matches against the values on the EXPR property of FNS. Creates a value assigned to ALLCONTENT (VARS) which has all of the file contents.
- [ ] GATHER-INFO(-3) shows a pretty weak collection of conclusions drawn:  NO-SOURCE: functions defined but no EXPR property, and DUPFNS: functions on more than one file.
- [ ] Phase 4 goes through all of the files and Masterscopes them all.

There are more things that could be checked for that seem likely: variables declared GLOBAL in some places and SPECIAL in others, calls to RESETLST / RESETSAVE which sets a global variable on the way in and resets it back on the way out, vs. making the variable SPECIAL and rebinding it.
Interlisp-10 only had 1 thread. A lot of the code wasn't rewritten to take advantage of mulit-threads.

Everytime we recompile a file we risk creating a conflict.

rmkaplan:
There are also some ambiguous-field and field-not-defined errors in the transcript from making the fuller database.

[...]

PR #1763, opened 2024-06-15
FILEWATCH, QIX, SOLITAIRE--rename fields that conflict with REGION and POINT

rmkaplan:
This eliminates a bunch of errors in building the fuller database (#1141).  Basically, renaming fields like LEFT, BOTTOM or X Y that conflict with core system field names.  

LLKEY, CHAT, and EDITBITMAP are still troubled.


masinter:
This patch might be OK, except it makes the code uglier in order to work around a (unimpactrul) step in the process.

How can we fix the problem once and for all? Would it work to have file or package relative fieldnames? The whole (fetch (RECORDNAME FIELDNAME) of X) is so verbose compared to X:FIELDNAME tht was originally used for record access in INterlisp-10

rmkaplan:
Packages that are included in the fuller database (or in the full sysout) should not have unqualified record accesses in general, and they should not redefine other system fields (like X, Y, LEFT...), particularly ones that are likely to appear unqualified in lots of other places.  A more compact syntax would be nice (even some systematic macros like I have done in Tedit), but avoiding conflicts is higher priority.

masinter (approved):
Good enough to patch the problem for now.

PR #1765, opened 2024-06-17
Pass the DW? argument to restore DWIMIFY functionality

rmkaplan:
Got lost somewhere along the way, probably when I was creating the browser


masinter:
closed by mistake

masinter (approved):
as per 6/24/24 meeting

PR #1766, opened 2024-06-17
Extra argument to COMPAREDIRECTORIES to specify directory names

rmkaplan:
Allows smaller headers for GITFNS browser


masinter (approved):
as oer 6/24/24 meeting

PR #1776, opened 2024-06-30
More consistent HCFILES filenames

masinter:


pamoroso:
I don't know if it's related to these changes but I get a 404 error when visiting https://interlisp.org/src

masinter:
interlisp.org/src should be live again with the "new" URLs. I was working on trying to simplify the process of building it.

pamoroso:
I confirm https://interlisp.org/src is back online.

masinter:
with this PR, the substring "/src/" isn't in the calls to HCFILES and MAKE-INDEX-HTMLS.
setting it up to be /src/ or /hc/ and/or   https://releases.interlisp.org  a network file server will also work'

MattHeffron (approved):
Approved in 7/8/24 meeting.

PR #1778, opened 2024-07-03
Add git commit ID to beginning of loadups .dribble files.

MattHeffron:
The commit ID is put into the .dribble file in the .sh scripts. (using `echo ... > name.dribble`) 
The changes to the lisp files **MEDLEY-UTILS**, **LOADUP-FULL**, and **LOADUP-LISP** are to enable DRIBBLE in append mode; i.e., the passing of `APPENDFLG` to `(DRIBBLE ... APPENDFLG)`.


masinter (commented):
I can't tell what problem these changes are solving.
The original problem, I thought, was that it was difficult to tell, from inside OR otuside the sysout (or preferably both) what system you were running in, ffor the purposes of aiding someone reading or writing an issue.

I see it's adding some info in the dribble files, and saving the dribble files was intended to somehow aid in this task, but it turns out they don't and adding more information to them don't help to close the connection.  The filedates of each file loaded in the sysout are actually in the sysout already. We're missing the commit IDs of the files actually loaded. 

The changes look innocuous but my (earlier) PR also changes MEDLEY-UTILS although the merge of the two sets of changes isn't something you cah do easily, unfortunately, with current GITFNS.

MattHeffron:
Withdrawing these changes.
I will leave the branch to keep the implementation of acquisition of the git commit ID in the scripts.
Once it is defined **where** and **how** that is to be retained within the `sysout`, _then_ **that** can be implemented instead.

masinter:
Putting more info in the dribble file is useful and certainly harmless, i just thought we likely need more. I look around and I have a lot of sysouts and there's no way now to tell exactly what they are, even when I run them. 

This is more a modern problem that wasn't much of an issue back then because sysouts were relatively huge.

masinter (approved):
discussed 10/7/24

PR #1779, opened 2024-07-05
Add INSPECTMACRO for windows

rmkaplan:
The WINDOW datatype has a bunch of fixed fields for items that are commonly used, and then it has USERDATA property list for other fields, some of which are part of the window system and some of which belong to particular applications.

When you inspect a WINDOW, you see the common fields all nicely laid out, but you have to click again to bring up the user-data properties in another window.  And if your are inspecting several windows, you have to keep track of which user properties go with which windows.

This adds an INSPECTMACRO for WINDOW that brings up the USERDATA properties to the single inspect window, as first-class properties at the bottom of the fixed-field entries, with a --USERDATA-- separator


nbriggs (commented):
Was there a reason to use backslash separators in the new names where the other functions in that style use forward slashes?

rmkaplan:
Well, there are both slashes and backslashes in there, I didn’t look too carefully at their distribution.

MattHeffron (approved):
Approved in 7/8/24 meeting.

Issue #1781, opened 2024-07-10
TEDIT.HARDCOPY will hang forever if called from XCL EXEC but not so from INTERLISP EXEC


fghalasz:
Weird bug!

This happens on certain (all) files with IM Image Objects (I think that's the trigger).   My main test file has been internal>Peano.tedit.  Another is lispusers/spy.tedit.

This happens on current Master as well as on RMK30--Another-Tedit-release

This happens whether outputing IP or PDF (and whether PDFSTREAM is loaded or not).

This happens whether you start Medley with or without greet.

From an XCL exec: `(IL:TEDIT.HARDCOPY (IL:OPENTEXTSTREAM 'internal/peano.tedit) 'Foo.IP T)` will hang forever (repeatedly calling \TEDIT.FORMATBOX, not recursively).

From an INTERLISP exec: `(TEDIT.HARDCOPY (OPENTEXTSTREAM 'internal/peano.tedit) 'Foo.IP T)` will finish quickly as expected.

FWIW:  This bug is holding up the final implementation of HCFILES automation.

rmkaplan:
The bug seems to be that the pageframes were being read from the file with the current readtable instead of the readtable that they were written with. So the atoms were showing up in XCL-USER instead of IL.  I don’t quite understand why that would cause this particular behavior, but fixing that seems to make the problem go away.  I’ll put up a PR.

Debugging Interlisp code under an XCL exec is cumbersome.  If you open an Interlisp function with SEDIT under an Interlisp exec, then look at it in a break under XCL, all of the symbols are reprinted with IL: in front, but Sedit’s alignment doesn’t know that.

The command VALUE seems to exist in XCL, but the variable IT (set by VALUE and by INSPECT so you can get your hands on a the datum or selection) has to be qualified with IL:.

PR #1782, opened 2024-07-11
Fix hardcopy under XCL  exec #1781

rmkaplan:
Read pageframes with proper readtable


masinter (approved):
minimal change, LGTM....

PR #1786, opened 2024-07-17
In scripts/do_hcfiles.sh, save hcfiles.dribble and hcfiles-fails.txt in loadups/.   

fghalasz:
do_hcfiles.sh now stores hcfiles.dribble in loadups/.  It also stores a hcfiles-fails.txt file in loadups/ that shows just the FAILS lines from the dribble file.

These files will not be indexed by MAKE-INDEX-HTMLS, but they will be uploaded to files.interlisp.org in the doHCFILES workflow.  To get them from fio, you can use:

`wget https://files.interlisp.org/medley/loadups/hcfiles.dribble`
`wget https://files.interlisp.org/medley/loadups/hcfiles-fails.txt`

**Also, added a new scripts/clean_hcfiles.sh that cleans out all of the generated files from the medley directory after running do_hcfiles.sh.**


scripts/clean_hcfiles.sh
```diff
@@ -0,0 +1,125 @@
...
+#  Copyright 2024 Interlisp.org
+#
+
+main() {
+    MEDLEYDIR=$(cd "${SCRIPTDIR}/.." && pwd)
+    export MEDLEYDIR
+    cd "${MEDLEYDIR}"
+    git clean -f
+    find . -iname index.html.~\*~ -exec rm {} \;
+    find . -iname \*.pdf.~\*~ -exec rm {} \;
```
masinter, line 22: There are a couple of .PDF files that are checked into master or loops. MAKEHCFILES adds ".pdf" lower case. Simplest is to change "-iname" to "name".

masinter (dismissed):
looks good. Cleanup needs a patch and testing but that can wait for next patch or whatever.

fghalasz:
Changed clean_hcfiles.sh so that index.html.~*~ files and *.pdf.~*~ files are deleted if and only if their non-versioned file is not being tracked by git.  This prevents the deletion of any files files related to pdf files checked into the repo.

PR needs to be re-reviewed.

masinter (approved):
As per 7/29/24 meeting

PR #1788, opened 2024-07-18
WINDOWOBJ:  Install classname in IMAGEFNS

rmkaplan:
IMAGEFNSCREATE had an argument CLASSNAME and the IMAGEFNS datatype had a field IMAGECLASSNAME, but the CLASSNAME wasn't being installed in the field


fghalasz (approved):
LGTM.  And looks good to Monday meeting. So go.

PR #1789, opened 2024-07-19
TAB-WINDOWS: A lispusers package that lets you step through open windows

hjellinek:
Start it running with `(START-TAB-WINDOWS)`.

Bonus debug tool: `(KEY-WINDOW)` starts a process that monitors keyboard and mouse button events and displays them in a little window.


nbriggs (commented):
If one has the meta key as the default (at least on a mac) COMMAND key, this loses out to the OS's COMMAND-TAB to cycle through the windows.

It should probably use `(SHIFTDOWNP 'META)` rather than checking for the `EXPAND` physical key - which may not be the key the user has assigned to the META function.

pamoroso:
Not sure how it's supposed to work. After evaluating `(START-TAB-WINDOWS)` in Medley Online on my Linux system, pressing `Meta-Tab` (mapped to `Alt-Tab`) carries out the corresponding Linux action of cycling though Linux windows, not Medley ones.

`(KEY-WINDOW)` works fine.

pamoroso:
What's *really* interesting is that, although the `(KEY-WINDOW)` window doesn't detect presses of the cursor keys, it does detect cursor keys in the numeric keypad as here in which I pressed the keypad up arrow, i.e. key `8`:

keypad8-up

nbriggs:
@pamoroso - that's because they're physical keypad keys, (many of) which are mapped, not the 4 cursor keys, which are not mapped.

fghalasz:
Some feedback:

1.  As other have pointed out:  ALT-TAB (META-TAB) is captured by the system and never gets to Medley.  This is true for me on Linux, WSL and Cygwin.  On Windows I disabled ALT-TAB window switching and TAB-WINDOWS still did not work because the ALT and TAB keys were still not making it thru to Medley (both WSL and Cygwin).

2.  On my machine for both WSL and Cygwin, the ALT-LEFT key is mapped to (BLANK-TOP KEYBOARD DBK-META) according to (KEY-WINDOW).  For the above tests, I tried changing the code to use 'DBK-META instead of 'META, but no luck - my guess is because of the system capture issues.

3.  I did have success changing 'META to 'CTRL in the code - so CTRL-TAB does do the expected window switching (sort of, see below).  CTRL-TAB is the right choice anyways - like Chrome does to switch tabs on Windows and Linux (and I suppose on Mac).

4.  But lots of issues with the windows switching:

       4.1  The CTRL and TAB characters make it thru to the process that has the TTY so that , for example, if the EXEC has the TTY you see the caret skipping across the EXEC window as you switch windows.   Typically shortcuts absorb the characters and don't pass them on.   Not sure if this is at all possible with the current architecture of having a separate process polling for the shortcut.

      4.2  It can get into a state where its just switching between two windows and "ignoring" the other 10 windows on the screen.  I didn't try to debug this, but my guess is that (OPENWINDOWS) is getting reordered in unexpected ways.

     4.3  It can get into a state where a single Window is "ignored".  You can switch among all the other windows, but that one window just won't come to the top.  Haven't been able to get a replicable case for this - it just happens.

     4.4  My preference would be for it to switch just among "application" windows.  If you have, for example, a lot of buttons on the screen, you spend a lot of time switching between the various buttons.  Maybe treat all the buttons (or maybe all the non-application windows) as one and bring them to the top all at once.

     4.5 There is a frequent issue of the window that has the blinking cursor (i.e, the TTYPROCESS) bumping immediately back to the top of the Z-order.   It probably would make sense to switch the TTY.PROCESS along with bringing the window to the top.

masinter:
aat the 7/29/24 meeting we discussed moving this to draft. The problem is that there isn't a set of key bindings that willworkfor eeryone on allplatforms.

It's not clear that you can pick a set of key strokes that will work for everyone

More generally, Consider merging this with "Slaam" as far as enumerating open windows of a particular type and doing something to them.  Ron has some nascent theories of how this kind of facility can be implemented generally.
We spent a lot of time exploring he space ofpossibledesins.
The sense was this might be useful but not enough to approve it without at least some way to make it useful in other configurations.

This summary doesn't capture the breadth of the discussion.

hjellinek:
I will mark it as a draft.

I should mention that I have this in my `INIT.LISP`:
````lisp
(KEYACTION 'EXPAND '(METADOWN . METAUP)) (* "map the option key to meta")
````

hjellinek:
We discussed this earlier in October 2024.  It's impossible to handle every combination of keyboard, OS, and window system in the code, and unnecessary in a lispusers package anyway.  Instead, we agreed that I'd document such problems.

hjellinek:
@masinter suggests changing the hot key from "meta" + TAB to a key combo that's not used for any other purpose.
I will try this.

masinter (approved):
as per 1/20/25
should open issue to fix keystroke

PR #1811, opened 2024-08-24
Added new module READ-BDF.

MattHeffron:
This will parse a bdf font file into a BDF::BDF-FONT structure. It does NOT create (convert into) an IL:FONTDESCRIPTOR instance. Minimal error checking!


MattHeffron:
**NOTE:** This **will not work** _until_ PR #1833 is approved, merged, and included in any loadup/sysout where READ-BDF is loaded.

masinter (approved):
As a new lispusers package that has nothing that depends on it, it's hard to test and will be tested. I'm approving it so everyone can use it.

A
There are some design choices that need to be made before this will be useful, it seems. 
1) What is the application? 
   What fonts do we use? Where do we get them? Do we add another kind of font-type? Change the font-reading code to read BDF too? or convert from BDF fonts to .strike? Will we handle _ and ^ as left-arrow up-arrow? Convert the fonts to be in XCCS order or fix medley to know about unicode?

etc. There are (Open) Issues for the font use, though.

MattHeffron:
> 1. What is the application?
>    What fonts do we use? Where do we get them? Do we add another kind of font-type? Change the font-reading code to read BDF too? or convert from BDF fonts to .strike? Will we handle _ and ^ as left-arrow up-arrow? Convert the fonts to be in XCCS order or fix medley to know about unicode?

This module is a prerequisite to _any_ use of BDF fonts.

There a lot of BDF files/collections on the net; many on GitHub. 
For example: [Tecate / bitmap-fonts](https://github.com/Tecate/bitmap-fonts.git)
I think that we can only point to whatever we find; redistribution could become quite a headache.

The simplest would be to add to this module functions for converting the fonts to `.strike`, as that doesn't risk breaking the existing font file handling. Rearrange the glyphs from Unicode to XCCS, and group them into the appropriate charset files. It would depend on the `library/UNICODE` module.

Unfortunately, I think that we must handle _ and ^ as left-arrow up-arrow, or else font selection could cause significant user confusion. 
I think that it would require **extensive** changes within Medley to switch to Unicode, _and_ require different character codes for left-arrow and up-arrow, _and_ to have a mechanism for one-time conversion of source files that includes some marker that the conversion already has been done (or else _ and ^ could not be used safely and it becomes somewhat pointless to do).

phantomics:
I was just testing this and had a problem. I downloaded the sample BDF file from `https://en.wikipedia.org/wiki/Glyph_Bitmap_Distribution_Format` and put it in `/tmp/sample.bdf`. Then, at an XCL REP I evaluated:

`(bdf:read-bdf "<tmp>sample.bdf")`

I got the error message:

`Invalid BDF file - STARTPROPERTIES count (STARTPROPERTIES) is invalid or missing.`

I tried CL as well with the same result. Am I missing something here?

MattHeffron:
@phantomics This needs the changes from PR #1833. They've been merged into master but they aren't in the latest release on GitHub. 
You could build the loadups yourself. See `scripts/README.md`.
Or you must `(LOAD (MEDLEYDIR "sources" "CMLREAD.LCOM") 'SYSLOAD)` (in an Interlisp Exec or add `IL:` prefix to the 3 symbols).
Then try `READ-BDF`.
It only creates the `BDF::BDF-FONT` structure. It doesn't yet create a Medley FONTDESCRIPTOR` or write `.strike` files.
So, it's not quite _useful_ yet, but it's the first step.

phantomics (commented):
Looks like it works.

phantomics:
I ran `(LOAD (MEDLEYDIR "sources" "CMLREAD.LCOM") 'SYSLOAD)` in an IL exec but I still get the same error. I can try updating Medley, the one I'm running is from July or so.

nbriggs:
Make sure your medley directory is updated from github...

phantomics:
Got Medley updated and I was able to read the font into the struct.

Issue #1814, opened 2024-08-26
Create an HTML image outputstream


hjellinek:
Create an image output stream implementation that writes HTML.

rmkaplan:
@hjellinek , I think I didn't quite understand what you were asking wrt utf-8, and I may have thrown you off the track.

If you want to create an image stream for HTML with UTF-8 as its character encoding, I think all you have to do is open the backing stream with the parameter (FORMAT :UTF-8), and then simply write to the backing stream with ordinary Medley character functions (PRIN1, TERPRI, PRINTCCHODE...).  You shouldn't need to worry about decoding and recoding, that should happen automatically.  (If you want to put out binary data, use BOUT to avoid any coercions.) Unlike Tedit, you shouldn't need to deal with the internals of the UNICODE file.

You might also want to specify the parameter (LINELENGTH T) so that the Medley functions don't thrown in unexpected EOL's in an attempt to do their own line-breaking (essentially treating PRIN1 as PRIN3 etc.)/

hjellinek:
Hi @rmkaplan, thanks for clarifying.  I'll be able to take advantage of all of that information in my eventual implementation.  I'm pretty sure I'll still need to perform my own conversion from XCCS to Unicode in one case, though: I've got a collection of font metrics that are indexed by Unicode character code.  To look up the width of a given XCCS character, for instance, I first need to convert it to the equivalent Unicode code point and use that value as the index into the font metrics table.

It's good to know I can use the native XCCS codes for everything else.

rmkaplan:
You may be thinking of this, but wrt fonts, you could take the problem off-line by separately shuffling the widths vectors to make new XCCS-organized font files.  That's where you would have to use some of the translation macros in Unicode.

hjellinek:
Yes, I did think of that, and I may pursue it.  At the moment, simply collecting the metrics from the fonts I've chosen (Google Web Fonts) is proving to be the challenge.

nbriggs:
Does https://github.com/drwpow/google-font-metrics provide any useful extraction of the metrics you need?

hjellinek:
The code I wrote uses a similar technique.  The insane thing is that none of the tools I've found actually extracts metrics from the font files themselves.  Instead, they draw the glyphs offscreen and measure them!

I should look for and use a WOFF2 file parser instead, but this approach is working for me for the moment.

nbriggs:
[...]

PR #1816, opened 2024-08-27
XTOUCODE doesn't fail for nonexistent mappings

rmkaplan:
assigns a unique otherwise unallocated Unicode code.  Addresses bug reported in #1814


hjellinek (approved):
+1

hjellinek:
By the way, `UNICODE.TEdit` mentions a `UNICODE.TRANSLATE` macro, but I couldn't find it in the `UNICODE` sources.

rmkaplan:
It’s at the end of the file, declare dontcopy

PR #1822, opened 2024-09-06
Extend DEFPACKAGE to accept uninterned symbols as the names to export…

MattHeffron:
…. They are treated the same as strings: the `symbol-name` is first interned in the package before being exported.
This is defined as the behavior in CLtL2, and using uninterned symbols appears to be common practice in other Common Lisp code.


masinter (approved):
discussed 9.23.24

PR #1823, opened 2024-09-06
Import XCL:DEFPACKAGE into LISP package, and then export it.

MattHeffron:
Make DEFPACKAGE accessible from LISP (aka CL:) package.
CLtL2.
See Issue #1684


masinter:
i've been looking for the changes I had made when introducing the CL:LOOP macro, but haven't found it yet. From wht I remember, there might be long-term problems in integrating the CLTL2 changes doing them one at a time.   It seems like there is a file that contain a list of the names of all of the symbols to be interned and exported from the LISP package, and the package setup code does LISP first. Then there is no need to intern the symbols in the XCL package first.  Is it the case that DEFPACKAGE is special? 

Anyway, i thought i should write up why i've not approved this PR yet.

Anzus:
Larry, I think the file you're looking for might be PACKAGE-STARTUP?

masinter:
there are likely some constraints on code used early in the loadup process that are not documented and might not be obvious.  I think it is the case that the MAKEINIT process that "loads" code into the initial virtual memory file still only works with LCOMs and not DFASLs. But the Interlisp compiler doesn't compile DEFUNs and DEFMACROs but just leaves them as "expressions" .
If DEFPACKAGE is needed early in the loadup before the runtime is defined then the NLAMBDA function will run compiled while a DEFMACRO definition will postpone evaluation until more of Common Lisp is loaded.
 
This is from memory, I don't think I've run into any documentation of this level of stuff other than my recent notes.

masinter (approved):
discussed 9/23/24 mtg

Issue #1812, opened 2024-08-25
CL:READ-FROM-STRING returned position is incorrect


MattHeffron:
**Describe the bug**
The `position` value returned from `CL:READ-FROM-STRING` is twice the value it should be. (It is returning the byte position, not the character position.)

**To Reproduce**
1. Run `full.sysout`
2. In the XCL Exec, enter: `(multiple-value-list (read-from-string "ABCDEF X"))`
3. The value `(ABCDEF 14)` is returned and displayed

**Expected behavior**
In step 3, the value returned should be `(ABCDEF 7)`.
Note that the [Common Lisp HyperSpec (CLHS)](https://www.lispworks.com/documentation/HyperSpec/Body/f_rd_fro.htm) page for `read-from-string` notes:
> _position_---an _integer_ greater than or equal to zero, and less than or equal to one more than the _length_ of the _string_.

The `14` is not within that range. Also from the CLHS:
> The _secondary value_, _position_, is the index of the first _**character**_ in the _bounded_ string that was not read.

**Context (please complete the following information):**
 - IL:MAKESYSDATE: 31-Jul-2024 02:24:38

**Other info
The second returned value from `read-from-string` appears to be used in **SEDIT-COMMANDS** in `extract-current-selection`. I don't know why this isn't affected by this.

tfeb:
On 25 Aug 2024, at 17:40, rmkaplan ***@***.***> wrote:
> 
> Separately, the spec for CL:READ-FROM-STRING is a little strange, saying "If the entire string was read, the position returned is either the length of the string or one greater than the length of the string.” Which is it?

I think the spec talks about this: there can (it claims) be cases where you (might) want to simulate an extra character at the end of the string and the thing is then allowed to return the index it would if that character was there.

That strikes me as a ludicrously poor argument: even for an implementation that does this (I presume there was one) then instead of making the implementation make at worst a call to MIN with one of the arguments being the string length, every program has to be careful.  That's C-level design.

--tim

MattHeffron:
> The second returned value from `read-from-string` appears to be used in **SEDIT-COMMANDS** in `extract-current-selection`. I don't know why this isn't affected by this.
SEdit didn't have an issue with this because `read-from-string` was using the `start` value with the same interpretation as byte position, not character position. This is corrected in commit 07e858d of PR #1833.

PR #1833, opened 2024-09-16
CL READ-FROM-STRING returns byte position instead of character position.

MattHeffron:
Fix for issue #1812


MattHeffron:
From issue #1812 
> The second returned value from `read-from-string` appears to be used in **SEDIT-COMMANDS** in `extract-current-selection`. I don't know why this isn't affected by this.

SEdit didn't have an issue with this because `read-from-string` was using the `start` value with the same interpretation as byte position, not character position. This is corrected in commit 07e858d.
**Ready for review.**

masinter (approved):
I think this has been thoroughly tested and reviewed.
it's a category of bug that could use some explanation.

Issue #1841, opened 2024-09-29
TigerVNC server not installed error with medley -v + on Raspberry Pi


pamoroso:
On my Raspberry Pi 400 running 64-bit Raspberry Pi OS (Debian Bookworm 12.7) I used Clang to build Medley X11 from the latest sources. When I launch Medley with the `-v +` option I get this error (BTW the error has a typo: `that that`):

```
pi@raspberrypi:~ $ medley -a -e -n -v +
Error: The -v or --vnc flag was set.
But it appears that that TigerVNC server \(Xvnc\) has not been installed.
Please install the TigerVNC server and try again.  On Debian and Ubuntu, use:
"sudo apt install tigervnc-standalone-server". On most other Linux distros, use the
distro's package manager to install the "tigervnc-server" package.
Exiting.
```

All the required TigerVNC components are installed (installato: installed):

```
pi@raspberrypi:~ $ apt list tigervnc* 
Elencazione... Fatto
tigervnc-common/stable,now 1.12.0+dfsg-8 arm64 [installato, automatico]
tigervnc-common/stable 1.12.0+dfsg-8 armhf
tigervnc-scraping-server/stable 1.12.0+dfsg-8 arm64
tigervnc-scraping-server/stable 1.12.0+dfsg-8 armhf
tigervnc-standalone-server/stable,now 1.12.0+dfsg-8 arm64 [installato]
tigervnc-standalone-server/stable 1.12.0+dfsg-8 armhf
tigervnc-tools/stable,now 1.12.0+dfsg-8 arm64 [installato, automatico]
tigervnc-tools/stable 1.12.0+dfsg-8 armhf
tigervnc-viewer/stable,now 1.12.0+dfsg-8 arm64 [installato]
tigervnc-viewer/stable 1.12.0+dfsg-8 armhf
tigervnc-xorg-extension/stable,now 1.12.0+dfsg-8 arm64 [installato]
tigervnc-xorg-extension/stable 1.12.0+dfsg-8 armhf
pi@raspberrypi:~ $ vnc
vncagent                vncpamhelper            vncserver-x11
vncconfig               vncpasswd               vncserver-x11-core
vncfopshelper           vncserver               vncserver-x11-serviced
vncinitconfig           vncserverui             vncsession
vnclicense              vncserver-virtual       vncviewer
vnclicensewiz           vncserver-virtuald      
```

However, no `Xvnc` program is in the executable path.

fghalasz:
[...]

PR #1842, opened 2024-09-30
Fix Issue#1841: using medley --vnc option on RaspberryPi OS

fghalasz:
In scripts/medley/medley_vnc.sh, update references to Xvnc and vncviewer to their tigervnc-specific versions Xtigervnc and xtigervncviewer.  Fixes Issue#1841.

### DIFF
--- a/scripts/medley/medley.command
+++ b/scripts/medley/medley.command
@@ -1489,10 +1489,10 @@ then
   #
   # Make sure prequisites for vnc support are in place
   #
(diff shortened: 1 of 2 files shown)


masinter (approved):
This is reported to not solve the problem in issue #1841, although the PR is innocuous otherwise and it's a check we want (to call explicitly tigervnc isntead of whatever Xvnc you have installed).

Issue #1806, opened 2024-08-16
\TEDIT.COPYINSERTFN is an undefined function error when copying bitmap to TEdit


pamoroso:
**Describe the bug**

When copying a bitmap to TEdit by shift-clicking I get this error in a break window:

```
In OLDFAULT1:
\TEDIT.COPYINSERTFN is an undefined function
```

**To Reproduce**

1. sign into your Interlisp Online account
2. under `Initial Exec`, select `Interlisp`
3. click `Run Medley`
4. open TEdit
5. click on the TEdit buffer
6. on the background, SHIFT+right-click
7. select `SNAP`
8. swipe a frame around the Medley logo window

**Expected behavior**

The bitmap is inserted in the TEdit buffer with no errors.

**Screenshots**

The full desktop with the break window:

![shift-click-error](https://github.com/user-attachments/assets/0510fabd-1cb6-40ca-8c8e-fc8c3fd55a05)

**Context (please complete the following information):**

 - OS: Linux Mint Cinnamon
 - OS Version: 22
 - Host arch: x86_64 (System76 Merkaat)
 - Maiko version: N/A
 - IL:MAKESYSDATE: 31-Jul-2024 02:24:40

**Additional context**

A debugger session:

```
MOUSE/5(debug)BT
OLDFAULT1
NEWFAULT1
FAULTAPPLY
COPYINSERT
BITMAPOBJ.SNAPW
EVAL
DEFAULT.BACKGROUND.COPYFNA0011
DEFAULT.BACKGROUND.COPYFN
WINDOW.MOUSE.HANDLER

MOUSE/6(debug)?=
  FAULTX = \TEDIT.COPYINSERTFN
  FAULTARGS = ({IMAGEOBJ}#144,171256 {WINDOW}#122,40474)
  FAULTAPPLYFLG = T
```

After making the `COPYINSERT` frame current:

```
MOUSE/7(debug)?=
  IMAGEOBJ = {IMAGEOBJ}#144,171256
MOUSE/8(debug)PB FAULTX
bindings for FAULTX: 
 @ OLDFAULT1 : \TEDIT.COPYINSERTFN
 @ TOP : NOBIND
MOUSE/9(debug)BTV
   IMAGEOBJ {IMAGEOBJ}#144,171256
   TTYW {WINDOW}#122,40474
   INSERTFN \TEDIT.COPYINSERTFN
   IMOBJ {IMAGEOBJ}#144,171256
COPYINSERT
   SCALE NIL
[...]

Issue #1738, opened 2024-05-28
TEDIT feature:  Quiet variant of READONLY


rmkaplan:
In Quiet mode, when the user tries to type into the document, Tedit would simply ignore the typein without putting up a warning message in the prompt window.

This would be useful for viewing applications, like @pamoroso hex dumper.  Also, DINFO man windows could be run with a Tedit process so that meta commands like meta-F (Find) etc. would go through, as in other windows.

pamoroso:
Addressed by #1857.

Issue #1697, opened 2024-05-05
Corrupted text in TEdit when printing long output to TEXTSTREAM


pamoroso:
**Describe the bug**
Suppose you create a `TEXTSTREAM` and associate it with a new TEdit window. If you print at least several dozen lines of output to the `TEXTSTREAM`, the text eventually ends up corrupted in the TEdit window.

**To Reproduce**
1. sign into your Interlisp Online account
2. under `Initial Exec`, select `Interlisp`
3. click `Run Medley`
4. evaluate `(SETQ TS (OPENTEXTSTREAM))`
5. evaluate `(TEDIT TS)`
6. evaluate `(FOR I FROM 1 TO 200 DO (PRINTOUT TS "Line no. " I T))`

**Expected behavior**
All the text appears in the TEdit window with no corruption or alterations.

**Screenshots**
The TEdit window:

![textstream1](https://github.com/Interlisp/medley/assets/10342479/0effdd08-3161-4b44-b145-9b5fe56cc455)

The full desktop:

![textstream2](https://github.com/Interlisp/medley/assets/10342479/c2907af2-1730-46d6-a290-9255399721c3)

**Context (please complete the following information):**
 - OS: chromeOS Stable
 - OS Version: 124.0.6367.95
 - Host arch: x86_64 (ASUS Chromebox 3)
 - Maiko version: N/A
 - IL:MAKESYSDATE: 22-Apr-2024 02:03:43

**Additional context**
Sometimes the text corruption appears when printing only a couple dozen lines of text. The issue occurs also when printing to the `TEXTSTREAM` from Common Lisp, for example with `format`.

pamoroso:
Does `OPENTEXTSTREAM` require the first argument, i.e. a stream? Can it be omitted?

pamoroso:
Passing a stream to `OPENTEXTSTREAM` like this makes no difference as the corrupted text issue still occurs:

```lisp
(SETQ S (OPENSTREAM '{NODIRCORE} 'BOTH))

(SETQ TS (OPENTEXTSTREAM S))
```

rmkaplan:
Thanks for finding this so clearly, I’ve seen this sporadically but haven’t been able to track it down.

It was an order of execution error in an early attempt to speed up the case of appending to the end of a stream. When it was updating the window as well as just updating the text, it was not properly reasserting the file position when the line formatter was trying to determine whether it was running off the bottom of the window.

[...]

Issue #1677, opened 2024-04-21
scrolling TEDIT window when left side off-screen updates incorrectly


nbriggs:
Open a TEdit window on some file, move the window so that the left side is off-screen, use the trackpad or mouse wheel to scroll the window - the wrong portion updates.

https://github.com/Interlisp/medley/assets/8138841/5bb8dc0e-5337-475a-8ce8-0be588b9938f

pamoroso:
Is this issue resolved by #1857?

rmkaplan:
Should be

rmkaplan:
Fixed in #1857

PR #1857, opened 2024-10-23
Rmk36  tedit fifth round

rmkaplan:
This is the long awaited, next major release candidate for Tedit.  Many changes to address many lingering issues, improve maintainabilty, robustness, and performance in some areas.   TEDIT-RELEASENOTES.TEDIT highlights some of the changes.

This is a draft, may have lingering problems--testing of screen/mouse interactions is difficult.  But please give it a try.


nbriggs:
Try opening the attached Slides.tedit.txt (remove the .txt) and scrolling it.  Older TEdit scrolls OK, latest TEdit dies with an arg not non-complex number `(IQUOTIENT 0 NIL)` under SCROLLW
[Slides.tedit.txt](https://github.com/user-attachments/files/17497325/Slides.tedit.txt)

The Slides file is from https://xeroxparcarchive.computerhistory.org/phylum/ctamarin/doc/talks/.index.html

rmkaplan:
> Try opening the attached Slides.tedit.txt (remove the .txt) and scrolling it. Older TEdit scrolls OK, latest TEdit dies with an arg not non-complex number `(IQUOTIENT 0 NIL)` under SCROLLW [Slides.tedit.txt](https://github.com/user-attachments/files/17497325/Slides.tedit.txt)
> 
> The Slides file is from https://xeroxparcarchive.computerhistory.org/phylum/ctamarin/doc/talks/.index.html

Default tab was not always initialized properly, now fixed.  File was a little bit odd: a run of tabs coded as 2-bytes (255 255 0 0 9), but that was more amusing than problematic.

nbriggs:
After this fix, again using Slides.tedit, if the TEdit window is not as tall as the image object (e.g., shorter than the Internal Stack Frame sketch) then it is impossible to scroll past the image object - either in the scroll bar, or using the scroll wheel/trackpad.

rmkaplan:
> After this fix, again using Slides.tedit, if the TEdit window is not as tall as the image object (e.g., shorter than the Internal Stack Frame sketch) then it is impossible to scroll past the image object - either in the scroll bar, or using the scroll wheel/trackpad.

This is somewhere on the list--I think Frank raised it quite some time ago.  The venuesysout shows a little bit more, but is goofy in a different way.

The question is whether to show a line that would be partially out of the window, either above or below.   The current strategy is to suppress the line completely, so that you don't see half of the characters on an overflow  line. That's pretty clean because most windows are much higher than the typical fonts, so a line is all or nothing.  But doesn't work well with characters in a very high font...and a high image object is treated essentially as a character in a very high font.  A line doesn't show at all if any part of its image would be outside the window.

I'm not sure what the right heuristic should be.  Turn off the filter for any line containing an object?  Turn it off for lines above a given height, whether due to a font or an object?  Suggestions?

nbriggs:
> A line doesn't show at all if any part of its image would be outside the window.

That doesn't seem to be quite what it is doing -- as you scroll into the large sketch object it *does* display the top part of the image, but won't allow you scroll past it (you can scroll towards the beginning but not towards the end).  And when playing around with scrolling it after reshaping the window to be a little bigger (but not big enough to get the whole thing)  I got another `arg not non-complex number: NIL` -- this time in `\TEDIT.SCROLLDOWN`.  BTV! attached.
[tedit-scroll.txt](https://github.com/user-attachments/files/17510161/tedit-scroll.txt)

nbriggs:
If it knows that the total height of a line (font, or image object) is larger than the height of the window could it not filter it out? Looking at modern apps, they don't filter at all - they jump by the height of the font you're in for input, but once the (large) text is there you can scroll to expose only a partial character.
image

The other complication is that the size of the "visible area of document" in the scroll bar keeps changing in size as you scroll:
Screenshot 2024-10-24 at 10 02 44 AM
Screenshot 2024-10-24 at 10 03 24 AM

SAVE.SYSOUT
nbriggs, line 1: Is this SAVE.SYSOUT supposed to be checked in as part of this?

rmkaplan:
No, sorry about that.

nbriggs:
Try this:  Open a moderate sized TEDIT window and type
```
asdf
asdf
asdf
qwer
``
leaving the caret at the end of `qwer`, and then in the left margin area (so it flips to the right slanted cursor) shift-select the `qwer`, and without moving the cursor or caret, shift-select it (the line, or where the line might have been) again and again and again.

rmkaplan:
On the SAVE.SYSOUT:  I think the problem is that .gitignore only excludes *.sysout, not also *.SYSOUT

rmkaplan:
(Actually, it only excludes the specific lisp.sysout, full.sysout, fuller.sysout, and apps.sysout.  Should it exclude all sysout files?)

rmkaplan:
On the qwer, is the problem that it shows the highlighting of the insertion (maybe as a line below) but it doesn't actually display the characters?

nbriggs:
Screen recording attached...

https://github.com/user-attachments/assets/df1c9259-b1e0-48ef-89bf-3a7af61598f4

nbriggs:
> (Actually, it only excludes the specific lisp.sysout, full.sysout, fuller.sysout, and apps.sysout. Should it exclude all sysout files?)

It ignores the named sysouts in the loadups/ directory, but nothing at the top level.

rmkaplan:
Give it another try.  Also look at the large image objects in the slide example.

nbriggs:
I built with the latest changes.  Here's a screen recording of what's going on.  Notice where the mouse is when I'm doing line selections (often not within the range of the characters in the document).  Notice how sometimes it scrolls everything up off the top even though it would all fit on the screen.

https://github.com/user-attachments/assets/1a30521e-f1ed-4ba4-bae8-8d725bd44cab

rmkaplan:
Hard to see exactly what you did, but I think I was able to see the same things.  The copy was “normalizing" the caret (early code=move to top) instead of using the new (and hopefully more intuitive) caret-scroller.  I’m trying to avoid big jumps in the caret, it’s a little tricky, there may be more.

nbriggs:
Try scrolling backwards with smallish increments through the tall image objects in the Slides.tedit.  The scrolling feels a bit off - I'm using mouse click scrolling not scroll-wheel.  Screen recording attached.

https://github.com/user-attachments/assets/b5f4f99b-4bba-455c-9130-84dfaa29fa1d

rmkaplan:
@nbriggs, I wasn't able to see the behavior in you last screen movie, but that may be because I continued to rework the scrolling edge cases (large objects, eol's adding and deleting in different contexts, etc.).  With the current commit, the only odd thing I have seen with that very large bit map is that if it is scrolled in just the right way, the screen is blank--and it comes back in if you jiggle it just a little bit.  So I'm still looking at that.

If you want to test scrolling a little more systematically, you can load TEDIT-DEBUG.LCOM, and then run (TSCROLL* N).  Every time you type U it will scroll the default Tedit window up by N points, typing D will scroll it down by N. Holding either key down will give repeated scrolling.  I have wheel scroll set at 20, so (TSCROLL* 20) tends to simulate a spin of the wheel.

nbriggs:
Made a new full.sysout with latest TEdit.  Went to the end of Slides.tedit and then used `(TSCROLL* 4)` with `u` to scroll backwards -- things that work fine scrolling forwards jump around when going backwards, and certain table headings (like "Internal Stack Frame") get stuck at the top while the image object scrolls over the header leaving partial text at the top of the screen.  Also, clicking in another window and then hitting `u` generated a `CL::%%NOT-NONCOMPLEX-NUMBER`  error, but that's probably only relevant when using `TSCROLL*`.  Images attached.
Screenshot 2024-11-30 at 8 51 05 PM
Screenshot 2024-11-30 at 8 51 16 PM
Screenshot 2024-11-30 at 8 51 25 PM
Screenshot 2024-11-30 at 8 51 34 PM
Screenshot 2024-11-30 at 8 51 42 PM
Screenshot 2024-11-30 at 8 52 13 PM

nbriggs:
Also, scrolling back-and-forth (in Slides.tedit) over "Cons Cells" and the image object below it, at the bottom of the screen, makes the image object disappear at odd times, and when you scroll back, the spacing between the title and the image changes quite noticeably.
Screenshot 2024-11-30 at 9 01 44 PM
Screenshot 2024-11-30 at 9 02 08 PM
Screenshot 2024-11-30 at 9 02 21 PM

rmkaplan:
@nbriggs , I still haven't been able to reproduce the scrolling white-outs that you saw.  Maybe the region parameters were slightly different, in that (WINDOWPROP W 'REGION) doesn't actually return the parameter that you give to CREATEW--it doesn't account for the promptwindow.

So, if you can make it happen again, send me the value of the region parameter that you actually gave to CREATEW.

Here's a more elaborate theory of why the iterative scrolling slows down noticeably when it is dealing with the large imageobjects in the SLIDES file:

When such an image object is totally inside the window, there should be no difference in speed--the image of the imageobj is entirely within the bitmap that is blitted from one place to another, like all other lines.  (The Newman paper actually mentions using the screen bitmap in this way as another trick in Bravo).  But if the imageobj is partially off-window and then is scrolled even a little bit, the image object's displayfn has to be called to produce the part of the imageobject that was previously invisible and now comes into the window.  In this case it has to invoke all of the sketch machinery to recreate the whole image--the displayfn doesn't know how to produce just the newly exposed top or bottom.

The slowdown isn't really apparent when there is user-interaction between the scrolling events, but pretty obvious when you hold down the u or d keys in TSCROLL*.

nbriggs:
It happens every time, for all the window sizes that I've swept out by hand, doing `(TEDIT "Slides.tedit")` -- here's a trace of the calls to `CREATEW` -- and screenshots showing the problem again

[createw.txt](https://github.com/user-attachments/files/17988534/createw.txt)

Screenshot 2024-12-02 at 10 39 38 PM
Screenshot 2024-12-02 at 10 40 07 PM
Screenshot 2024-12-02 at 10 40 21 PM
Screenshot 2024-12-02 at 10 40 31 PM

As far as the speed of scrolling goes - I don't think it's the performance, it's the amount that it actually scrolls the window for each "u" or "d" it's shifting from scrolling about 1 line per u/d to somewhere around 1-4 pixels per u/d when it's moving slowly.

rmkaplan:
OK, I can see it. You scroll down and the object disappears, you do it again and a piece of it comes back.  Probably another case where the image is all set up, but the  clear-below-line clean-up function is called on the wrong line and wiping the image out.

The scrolling distance that you specify is really a minimum, since it is still trying to honor the constraints that it doesn't show partial lines and moves at least one line no matter what distance is specified (even 1 point, although I haven't tried that).  So it will move at least 4, but if it takes 10 points to get to the next full line, it will move 10 (or more).  But if the line is "tall" (= won't fit in the pane at all and/or is more than 50 points) and isn't already completely in the window, it gives up on the line constraints and takes the specified distance literally.  So if you specified 4, that's what you get.  And it then takes a lot more u's and d's to move it along, until it clears.

rmkaplan:
(BTW, if you break \TEDIT.CLEARPANE.BELOW.LINE just before the "d" that gives the incorrect white space, you will see what is supposed to be showing)

rmkaplan:
Clearing under the bitmap before printing the bottom lines seems to eliminate the last reported scrolling anomaly.

pamoroso:
Here are some issues round 5 likely fixes but there may be others:

* #1677
* #1697
* #1738
* #1806

masinter (approved):
TEdit fifth round extensively tested.
12/23/2024 meeting

PR #1878, opened 2024-11-11
Update github action workflows to account for deprecation of macos-12 runner and move of ubuntu-latest runner to ubuntu-24.04

fghalasz:
Update github action workflows to account for deprecation of macos-12 runner and move of ubuntu-latest to ubuntu-24.04.  

Removed use of ubuntu-latest and macos-latest in favor of explicit versions (macos-14, ubuntu-24.04) in order to prevent issues when *-latest changes without our notice.


hjellinek (approved):
These changes look sensible.

Issue #1868, opened 2024-10-28
Windows installer doesn't escape spaces in file paths.


kotenok2000:
**Describe the bug**
When i tried to install medley on Windows i have set install path to E:\Program files\interlisp
It installed cygwin to E:\Program
**To Reproduce**
1. Run https://github.com/Interlisp/medley/releases/download/medley-241022-5fef8528/medley-full-cygwin-x86_64-241022-5fef8528_240513-4becc6ad.exe
2. Use  install path cotaining spaces.
3. It installed cygwin to E:\Program and couldn't find bash.

**Expected behavior**
Interlisp handles file paths containing spaces without errors.
**Screenshots**

**Context (please complete the following information):**
 - OS: Windows 11
 - OS Version: 10.0.22631.4317
 - Host arch: x86_64
 - Maiko version: [e.g. commit ID from `git log | head`]
 - IL:MAKESYSDATE: [ date ]

**Additional context**

masinter:
probably should go through all the scripts and make sure parameters are quoted

fghalasz:
Fixed in Pull Request #1880.

nbriggs:
https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

PR #1880, opened 2024-11-15
Fix Issue #1868: Cygwin installer now handles install directory with spaces in its name.

fghalasz:
Cygwin installer now handles install directory with spaces in its name (see Issue 1868)

Also fixed the uninstall script do that it actually deletes all of the installed files.  Previously was deleting just some of the files.

Also added a powershell script that makes the local testing of the cygwin installer easier.

A cygwin installer that includes this PR can be found here: [https://github.com/Interlisp/medley/releases/download/untagged-532549860f120d8f9812/medley-full-cygwin-x86_64-241114-28315513_241110-e68fb0dd.exe](https://github.com/Interlisp/medley/releases/download/untagged-532549860f120d8f9812/medley-full-cygwin-x86_64-241114-28315513_241110-e68fb0dd.exe)

### DIFF
--- a/installers/cygwin/.gitignore
+++ b/installers/cygwin/.gitignore
@@ -4,3 +4,4 @@ maiko*.tgz
 setup-x86_64.exe
 medley.bat
 
+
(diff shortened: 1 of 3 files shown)


masinter (approved):
Discussed in recent meeting -- let's accept the change.

PR #1884, opened 2024-11-24
Rmk37  git ignores .sysout files

rmkaplan:
Previously, only specific sysouts in the loadups/ directory were ignored.  But is it ever right to upload a random sysout to git, even by accident?


nbriggs:
The TEDIT-CHAT patch probably shouldn't be here.  We do have `internal/loadups/starter.sysout` checked in.  If you have a pattern in .gitignore even if you explicitly try to commit something matching it won't.  I'm not sure what it does with an attempt to commit something that has previously been committed, but now is ignored.

rmkaplan:
In TEDIT-CHAT, I wanted to remove the reference to STREAMHINT

rmkaplan:
... but it slipped in in the wrong commit.  I'll remove it.

rmkaplan:
> We do have `internal/loadups/starter.sysout` checked in. If you have a pattern in .gitignore even if you explicitly try to commit something matching it won't. I'm not sure what it does with an attempt to commit something that has previously been committed, but now is ignored.

I assume that it won't delete the file, either locally or remotely.  With this in .gitignore, I'm still able to make loadups.

Much more likely that a sysout would be accidentally created and slip in to a commit, than that we want to update our starter sysout.  If we did want to do that, the .gitignore could be temporarily reverted.

masinter:
I wonder about this. Doing a "git add" without checking for and examining new files seems like a bad practice. But all this does is let you "git add" without looking to see if you really want them.  Sure there are work-arounds but this just adds some complexity.... 
Don't git add new things without checking.

PR #1892, opened 2024-12-03
Update .ignore to suppress uploading of .sysout files

rmkaplan:
Except for special purposes, we don't want to pollute the repo with accidental sysouts.


nbriggs (approved):
LGTM.

PR #1896, opened 2024-12-07
Rmk37  versiondefs

rmkaplan:
VERSIONDEFS provides basic tools for dealing with the definition of an item on different medley file-versions.  VERSIONDEFS itself also includes some version-editing functions.   EXAMINEDEFS and TEDIT-PF-SEE piggy-back on VERSIONDEFS to also operate on different versions.


rmkaplan:
This wouldn't merge, so replaced by  #1931

Issue #1900, opened 2024-12-10
RENAMEFILE between {DSK} and {UNIX}:  Is it necessary to copy?


rmkaplan:
**Describe the bug**
This showed up as one of the failures in the HCFILES log.  The file docs/user-guide-v1.2/10-USER-CARDS.TEDIT fails not in Tedit but afterthe shell conversion from PS to PDF, when it is trying to rename the pdf in /tmp/ back to its ultimate destination in {DSK}.

RENAMEFILE in this situation thinks that it is renaming the file from one file-device to another, not recognizing that these 2 devices are really just 2 versions of the same underlying file system.  The copyfile decides (either because it thinks the file is of type TEXT or the external format of files on the different devices is different) that it needs to do a COPYCHARS instead of a COPYBYTES.  

It is a separate issue why that fails at a low character reading level.  The question here is:  does it really need to do the copyfile at all?  Is there a better test for whether or not the low-level device rename operation can be invoked when going between DSK and UNIX?

For example, instead of just testing whether the devices are the same, test for whether they are somehow equivalent wrt the rename operation (\UFSRenameFile in this case).  Or, perhaps have a separate list somewhere that lists pairs of devices that can rename without copying.

**To Reproduce**
Hardcopy that notecards file, you should get a break.

This was on Macos.

rmkaplan:
To resurrect a previous discussion, if we could begin to shift the actionable-extension around the Medley version encoding, pdf conversion wouldn’t have to deal with tmp/ and renaming. Has your rework of the version/file-name code made that shift any more feasible?

That would fix this particular application, but still there is a general question about what rename needs to do (and a separate issue as to why the copychars is failing—a glitch somewhere in the low-level external format interface.

rmkaplan:
Does \GENERIC.RENAMEFILE really need to call \COPYOPENFILE if it is always going to just do COPYBYTES?  Other than the opening and the standard error-reset stuff, isn’t the rest of it useless or wrong for the rename case?

A separate question whether some of that stuff is still good when \COPYOPENFILE is called from COPYFILE. \COPYOPENFILE is only called from those 2 places.

nbriggs:
No idea why it does it that particular way, or whether it is a good way to do it.

rmkaplan:
[...]

Issue #1830, opened 2024-09-13
PDFs missing from NoteCards HC files at files.interlisp.org


pamoroso:
Some TEdit files of the NoteCards documentation are missing the matching PDFs the HC file generation process should make from them at `files.interlisp.org`. In directory [{MEDLEY}/notecards/docs/user-guide-v1.2/](https://files.interlisp.org/medley/notecards/docs/user-guide-v1.2/) the following TEdit files have no matching PDFs:

* 03-SOFTWARE-INSTALLATION.TEDIT
* 04-SYSTEM-USE-ISSUES.TEDIT
* 05-NOTECARDS-BASICS.TEDIT
* 11-SYSTEM-CARDS.TEDIT
* APP-Z-PROG-INTERFACE.TEDIT

Most or all PDFs are missing from [{MEDLEY}/notecards/docs/user-guide-v1.2/from_envos/](https://files.interlisp.org/medley/notecards/docs/user-guide-v1.2/from_envos/) and [{MEDLEY}/notecards/docs/misc/from_envos/](https://files.interlisp.org/medley/notecards/docs/misc/from_envos/).

masinter:
these are all files that failed in some way (an error or call to HELP) when run through TEdit or (for Lisp source files) have package problems. 

The Notecards files might have missing image object sources, because Notecards isn't in WHEREIS.HASH for TEdit to pick up.

There may be other reasons that all need to be fixedd in the HCFILES process or perhaps some TEdit bugs.

I think the way to address these is to take the files one at a time, try to TEdit hardcopy the file, fix the problems encountered and then try again. Since many of the problems affect more than one file.

masinter:
@pamoroso @rmkaplan @fghalasz The task here is to figure out why HCFILES fails on these files by seeing what errors arise when you just try to TEDIT the files and "hardcopy" to pdf or postscript.

@fghalasz wrote:

The hcfiles-fails.txt and hcfiles.dribble files from the weekly builds are stored along with all the hcfiles-generated PDF files on files.interlisp.org.  Unfortunately they are not included in the html index pages that HCFILES creates because they are not available until after HCFILES is run.  

But they are available at https://files.interlisp.org/medley/loadups/hcfiles-fails.txt and https://files.interlisp.org/medley/loadups/hcfiles.dribble either in a browser or via wget.

hcfiles-fails.txt is just a selection from hcfiles.dribble of every line that ends in IL:FAIL.

pamoroso:
[...]

PR #1902, opened 2024-12-10
Make PDF BINARY on DEFAULTFILETYPELIST

rmkaplan:
This prevents RENAMEFILE from trying to do COPYCHARS from UNIX to DSK, at least for PDF files.

Not sure whether PS files should also be changed from TEXT to binary.

It's odd that UFS specifies the type for these add-on file types.  I kept that, but also added it to PDFSTREAM where it really belongs.

See the discussions in #1830 and #1900.

HCFILES should be rerun with this, it may resolve some number of failures.


masinter (approved):
12/23/24 meeting

Issue #1897, opened 2024-12-08
Remove unused field from FONTDESCRIPTOR


rmkaplan:
FONTDESCRIPTOR has a bunch of fields that are not set or fetched in any code that the Masterscope database knows about.  It seems that these are left over from the days of fonts with only 8-bit characters (Altofonts?) and were not cleaned out when the information moved down into the per-character-set CHARSETINFO's.

The fields are
(FBBOX SIGNEDWORD)
(FBBOY SIGNEDWORD)
(FBBDX SIGNEDWORD)
(FBBDY SIGNEDWORD)
(\SFLKerns POINTER)
(\SFRWidths POINTER)

It is confusing to have those fields hanging around.  As a first step (which avoids needing to recompile all the clients), I propose to replace all those field-names with NIL so it is clear that they should eventually disappear.

There is a problem with kerning apart from the useless \SFLKerns field.  The CHARSETINFO has a LEFTKERN field, which is referenced only by \READACFONTFILE.  That would be useful if Tedit or anybody else wanted to do a better job of formatting.  But the function \FGETLEFTKERN takes an 8-bit code and a kern-vector as its argument, assuming that the caller has already landed on the kern vector by getting the CHARSET info for the character's high-order byte.  In that respect it is not parallel to e.g. \FGETCHARWIDTH.  \FGETCHARWIDTH takes a fontdescriptor and 16-bit code and does a 2-step retrieval. I think \FGETLEFTKERN should follow the same protocol (whether or not it is ever used).

nbriggs:
There seems to be a bunch of commented out code that used to use these fields, so I agree that they're probably obsolete.  I wonder if we have sufficiently old copies of the font code that it would predate the introduction of the NS character coding for FONTDESCRIPTORs.

masinter:
There are a huge number of cleanup tasks that would individually alone get us closer to the project goals, without risking introducing backward incompatibilities.

* For example, currently you CANNOT RELY on FULLER.DATABASE to tell you where a record field might be used. It's close, but it's not 100%. Fixing that would help. 
* Computing diffs of compiled code to insure that GLOBALVARS declarations vs. SPECVARS would help with insuring we can recompile every file without loss.
* There are a few records which have different definitions for the same record name -- these need to be eliminated to avoid unwittingly redefining the record at the wrong time.
* there are some files included in the loadup with 90% or 99% of the code unused and not useful. Sorting these out will remove some sources of conflicts. 

[...]

PR #1910, opened 2024-12-14
Remove outdated FONTDESCRIPTOR fields

rmkaplan:
As per #1897.  I also added a DEFPRINT function for FONTDESCRIPTOR.

I'd like to get kerning to work (for fonts that have that information), but I want to confirm:  A kerning table is logically indexed by 2 characters, and it indicates how the distance between them should be modified when they appear in sequence.  I assume that most entries in such a table are 0's--no adjustment--so that a representation as a 2-level Alist in each charset would be reasonable.  Is that true?


masinter:
i would wonder what fonts we have with accurate kerning information.
Without the fonts and an infrastructure to create and validate them, isn't the whole enterprise moot?
My impression is that modern current fonts use some other more flexible data structures and algorithms to manage the equivalent of kerning, with more context than just the two adjacent characters. This is, for example, the handling of characters that change shape when they occur at word endings.

masinter (changes requested):
This just removes the field names. The argument is that by itself its mainly harmless because we believe those field names aren't used, and that any code that might be around that used them would be broken anyway. 

Since the goal of this is to reduce confusion for people who later try to understand the font code, a comment explaining the situation would really be helpful.

rmkaplan:
I think changing shape (presumably character codes) at the ends of words is a different problem than changing the spacing between adjacent characters.

rmkaplan:
I think this is ready to merge

masinter (approved):

PR #1930, opened 2024-12-24
Clicking See for pdf files will do the ShellOpen

rmkaplan:
instead of crashing into Tedit.  prc will inherit this behavior


masinter (approved):
remaining problems have to do with PDFSTREAM

nbriggs (approved):
Looks good.

PR #1931, opened 2024-12-24
Manipulate versions of definitions by their ordinal file numbers

rmkaplan:
This replaces the original versiondefs before (closed), which git thought had some conflicts


MattHeffron (approved):
Per 1/20 meeting

PR #1933, opened 2024-12-24
Open man page with TEDIT READONLY-QUIET instead of OPENTEXTSTREAM

rmkaplan:
This creates a Tedit process along with the usual textstream and window so that some of the Meta command keys work in the expected way.  meta-F and meta-G for searching, without having to bring up the menu, meta-D for other man pages, meta-O to open selected function in Sedit, meta-T to display the function definition (if TEDIT-PF-SEE is loaded).


masinter (approved):
LGTM discussed 1.20.25

PR #1937, opened 2024-12-25
Tedit - a few rough edges

rmkaplan:
-  Scaling was mixed up for truehardcopy vs hardcopydisplay (symptom: page numbers in the wrong place)
-  Redundant promptwindow message after Put
-  Guard against page-looks menu without selecting a page type
-  Guard against empty page-heading specs in page-looks menu
-  Added flag to FONTDESCRIPTOR to say that font has no kerning information (so Tedit can reference it)


masinter (approved):
Tedit 5gh round cleanup

PR #1939, opened 2024-12-25
SEE-PDF searches for PDF file

rmkaplan:
Causes a FILE-NOT-FOUND error if it can't find the PDF on DIRECTORIES


pamoroso:
On Linux Mint 22 Cinnamon I switched to this branch and built Medley with `scripts/loadup-all -apps`. I evaluated `(SEE-PDF 'NONEXISTENT)` to test the PR but the expression returns `SEE-PDF is an undefined function.` and `(APROPOS "SEE-PDF")` returns `NIL`.

`SEE-PDF` is available and works on master.

rmkaplan:
SEE-PDF is in library/PDFSTREAM.  Is it loaded?

pamoroso:
I didn't realize I had to load PDFSTREAM manually as I seem to remember it was preloaded in master. Anyway, once I load it and evaluate `(SEE-PDF 'NONEXISTENT)` I get the expected break window with the error:

```
In ERROR:
FILE NOT FOUND
NONEXISTENT
```

masinter (approved):
LGTM

Issue #1943, opened 2024-12-29
Errors when pressing M-d or M-o in TEdit with no selected text


pamoroso:
**Describe the bug**

The 5th round of TEdit changes bound the `Meta-d` and `Meta-o` keychords to commands for looking up in DInfo and editing the selected word. But pressing `Meta-d` or `Meta-o` when no text is selected brings up a break window with this error for `Meta-d`:

```
In CL::%%NOT-NONCOMPLEX-NUMBER-ERROR:
NIL is not a NUMBER.
```

and this for `Meta-o`:

```
In \EOSERROR:
End of file #<IO Text Stream/127,135400>
```

**To Reproduce**

1. launch Medley
2. open TEdit
3. press `Meta-d` or `Meta-o`

**Expected behavior**

TEdit does nothing, or without breaking into the debugger warns that no word was selected.

**Screenshots**

The break window after pressing `Meta-o`:

![tedit-meta-o-error1](https://github.com/user-attachments/assets/3cfaf5d5-2ca4-4e66-96fb-8ea174aa75b2)

The full desktop:

![tedit-meta-o-error2](https://github.com/user-attachments/assets/06289d50-e5e3-4044-ac6b-85696181fc28)

**Context (please complete the following information):**
 - Are you using online.interlisp.org? no
 - OS: Linux
 - OS Version: Mint 22 Cinnamon
 - Display/window system: X11 + VNC
 - Host arch: x86_64 (System76 Merkaat)
 - Maiko version: f23a43f1a0cffebdfa08660dbb5cd1a1dac7f7c3
 - `IL:MAKESYSDATE`: 24-Dec-2024 08:36:18

**Additional context**

The backtrace after pressing `Meta-o`:

```
TEdit/5(debug)BTV
\EOSERROR
\TEDIT.TEXTBIN
\BIN
   STREAM #<IO Text Stream/127,135300>
   COUNTP NIL
\TEDIT.TEXTINCCODEFN
   STREAM #<IO Text Stream/127,135300>
   EOLC 0
   BYTECOUNTVAR NIL
   BYTECOUNTVAL NIL
   A0147 \TEDIT.TEXTINCCODEFN
\INCCODE.EOLC
\SUBREAD
   SI::*CLEANUP-FORMS* \TOP-LEVEL-READA0003
SI::*UNWIND-PROTECT*
   STREAM #<IO Text Stream/127,135300>
   EOF-SUPPRESS NIL
   EOF-VALUE NIL
[...]

PR #1944, opened 2024-12-29
Proper error messages for meta-O and meta-D on empty selection

rmkaplan:
Addresses #1943


fghalasz (changes requested):
TEDIT.PROMPTPRINT clears and flashes for Meta-D error message.  It does neither for Meta-O error message.  Should at least clear before the Meta-O error message.

rmkaplan:
I think I have made the requested changes and I hope this is mergeable.  It's getting harder to track with the unapproved PR's

pamoroso:
I tested this PR.

`Meta-D` with no text selected: `Please select a man-page key`.

`Meta-D` with text selected: brings up relevant DInfo and HyperSpec entries.

`Meta-O` with no text selected: `Please select a symbol to edit`.

`Meta-O` with text `times` selected (similar for `myfun`): `times has no definition to edit`. I would expect SEdit to be opened when no definition is found.

rmkaplan:
If symbol has no definition, there would be nothing for Sedit to edit.

Applying ED to an undefined symbol brings up a menu to set up a definer for that symbol.  That may be a useful crutch for a first-time or naive user, but it almost always results in requiring a useless and distracting mouse click—the problem is almost always either a typo or package confusion.  I don’t think that behavior should be replicated through this little Tedit interface.

(Separately, as I have mentioned, I think ED should be revised to be more helpful and less intrustive in the no-definition case.)

fghalasz (approved):
Ok. Changes to prompt messages work as expected.  Ready to go IMHO.

PR #1947, opened 2024-12-31
TRUEDEVICE behaves more like \GETDEVICEFROMNAME

rmkaplan:
tries to create a device if it doesn't yet exist


nbriggs (approved):
Looks good.

MattHeffron:
I **just** noticed that this PR put `PSEUDOHOSTS` (and `.LCOM`) into `library/` but left the previous version (and corresponding `.TEDIT`) in `lispusers/`.  **This should be corrected!!**

MattHeffron:
@rmkaplan Was your intent to **move** `PSEUDOHOSTS` to `library/`? 
Did you use the shell, or Medley to move/copy the file, instead of the `git` command to move the file(s) and preserve their git histories?

nbriggs:
I believe that it should be `git rm'd` from Lispusers - pending Ron's response.

MattHeffron:
> I believe that it should be `git rm'd` from Lispusers - pending Ron's response.

That would lose any git history of `PSEUDOHOSTS`. 
It _seems_ the correct thing would be to go back to before the files were put into `library/`, `git mv lispusers/PSEUDOHOSTS library/` (and likewise for the `.LCOM` and `.TEDIT` files), then remake the changes in the files in the new location (and recompile...).

nbriggs:
Ideally, yes.  I suppose since it's a wholesale replacement of a file that would be not too hard to do -- I've had a *lot* of trouble when I was trying to merge just a couple of functions from my file (ADIR) into an updated version - and it wasn't git that was the problem, it was Lisp being too "smart" when dealing with the file names (that were totally hosed with incorrect version numbers and paths) and doing LOADFROMs to get just the functions I wanted into the image to remake the file which had been updated.

rmkaplan:
Yes, I intended to remove it. This was a first step in what we have talked about, getting more of the lispusers files in full into the library.

I believe I used RENAMEFILE in Medley to move it over in my working directory, although maybe I just did a cleanup in library to create a new version there.  And then used GITFNS to copy the library version to the clone, and (but now I can’t be sure) I did the separate delete command to remove any versions hanging on in the clone’s lispusers.  

Maybe GITFNS should figure out that it sees an add and delete of files with the same content under different names or directories, and then offer to coerce the menu action on either the add or the delete into a rename, with corresponding version cleanup.

MattHeffron:
@rmkaplan If you're moving files around **within** the repository, or renaming, that must be done with `git` operations, not just file operations. If you do an add and delete as separate file operations, it's already too late to do the `git mv`. (Unless you revert the add and delete, and then do the `git mv`, **before** doing a commit.)

@nbriggs See my PR #1980 that I _think_ does this correctly.

rmkaplan:
That would do the right thing wrt the repository, but I don’t think that by itself would also keep the local Medley history = the previous Medley versions in the source directory.  If you just mv the top one, the others will hang around in the old place or under the old name in a confusing way.

So, do we need a Medley-level function that wraps up both the git-level mv for the newest version but Lisp-level renames for the older ones (and gets things linked properly)?

Issue #1953, opened 2025-01-04
TEdit Menus: MB.ADD of FIELD with (FIELDTYPE IMAGEOBJ) doesn't handle INITSTATE correctly


fghalasz:
`(MB.ADD '((FIELD (FIELDTYPE IMAGEOBJ)(INITSTATE <IMAGEOBJ>)(PREFIX 'A)(IDENTIFIER 'A))) TS)`

results in "{IMAGEOBJ}#765,123456}"  in the value of the A FIELD and not the IMAGEOBJ itself.

Issue is in:  MB.FIELD.CREATE, specifically the following lines:

```
(CL:WHEN (AND INITSTATE (NEQ INITSTATE '**EMPTY**))
   (TEDIT.INSERT MENUSTREAM (MKSTRING INITSTATE) CH# '(FONT ,FIELDFONT)))
```

But for (FIELDTYPE IMAGEOBJ): there should be no MKSTRING and its should be TEDIT.INSERT.OBJECT and not TEDIT.INSERT.

Also, why not use MB.FIELD.SETSTATEFN instead of TEDIT.INSERT here?

rmkaplan:
Fixed in #1957

Issue #1952, opened 2025-01-04
TEdit Menus:  MB.ADD of a FIELD causes TEdit window to always scroll so that end of field is on top line - even if this is not called for


fghalasz:
MB.ADD of a FIELD causes TEdit window to always scroll so that end of field is on top line - even if this is not called for.

Example:
```
> (SETQ MD '((FIELD (FIELDTYPE STRING) (PRELABEL "aaa") (IDENTIFIER aaa) (INITSTATE
"aaa")) EOL (FIELD (FIELDTYPE STRING) (PRELABEL "bbb") (IDENTIFIER bbb) (INITSTATE
"bbb")) EOL (FIELD (FIELDTYPE STRING) (PRELABEL "ccc") (IDENTIFIER ccc) (INITSTATE
"ccc")) EOL))
((FIELD (FIELDTYPE STRING) (PRELABEL "aaa") (IDENTIFIER aaa) (INITSTATE "aaa")) EOL (FIELD (FIELDTYPE STRING) (PRELABEL "bbb") (IDENTIFIER bbb) (INITSTATE "bbb")) EOL (FIELD (FIELDTYPE STRING) (PRELABEL "ccc") (IDENTIFIER ccc) (INITSTATE "ccc")) EOL)

> (SETQ TS (TEXTSTREAM (TEDIT)))
#<IO Text Stream/146,176100>

> (MB.ADD MD TS)
19
```

Here is the TEdit window immediately after the MB.ADD
![AA](https://github.com/user-attachments/assets/2db6d134-3570-4089-b592-b31e4598e809)

Here is the same TEdit window after scrolling up a bit
![BB](https://github.com/user-attachments/assets/08190879-095f-436b-a627-d27c249a8260)

Clearly the MB.ADD succeeded, it just scrolled the final EOL to the top line of the window.

I tried to track this down a bit.  For some reason \TEDIT.VISIBLECARETP (inside TEDIT.NORMALIZECURSOR inside TEDIT.INSERT.OBJECT inside MB.FIELD.CREATE) is returning NIL when (I think) it should be returning a value.  Looks like a timing issue of some sort - but way to complicated for me to track down further.

Also seems like this might be related to Issue# 1950.

Issue #1950, opened 2025-01-03
TEDit Menus: Display messed up immediately after MB.ADD of FIELD to existing TEdit Menu


fghalasz:
Immediately after MB.ADD of a FIELD + EOL to an existing TEdit menu of FIELDs, the display is messed up.  If you scroll up and then down, the display returns to the expected.  So it appears that the MB.ADD completes correctly, its just that the display immediately after the MB.ADD is messed up.

Example:
```
>  (SETQ MD '((FIELD (FIELDTYPE STRING) (PRELABEL "aaa") (IDENTIFIER aaa) (INITSTATE
"aaa")) EOL (FIELD (FIELDTYPE STRING) (PRELABEL "bbb") (IDENTIFIER bbb) (INITSTATE
"bbb")) EOL (FIELD (FIELDTYPE STRING) (PRELABEL "ccc") (IDENTIFIER ccc) (INITSTATE
"ccc")) EOL))
((FIELD (FIELDTYPE STRING) (PRELABEL "aaa") (IDENTIFIER aaa) (INITSTATE "aaa")) EOL (FIELD (FIELDTYPE STRING) (PRELABEL "bbb") (IDENTIFIER bbb) (INITSTATE "bbb")) EOL (FIELD (FIELDTYPE STRING) (PRELABEL "ccc") (IDENTIFIER ccc) (INITSTATE "ccc")) EOL)

>  (SETQ TS (TEXTSTREAM (TEDIT)))
#<IO Text Stream/146,176100>

> (MB.ADD MD TS)
19

> (MB.ADD '((FIELD (FIELDTYPE STRING)(IDENTIFIER 'ddd)(PRELABEL "ddd")(INITSTATE
"ddd")) EOL) TS 7)
13
```
TEdit display after first MB.ADD
![Start](https://github.com/user-attachments/assets/d20b8b49-8c0a-4d26-b77d-5c43bbce9eb5)

TEdit display immediately after second MB.ADD
![Before Scroll](https://github.com/user-attachments/assets/e1a61ed4-26c4-48e9-b176-74c922045c2b)

TEdit display after scrolling up and then down after the second MB.ADD
![After Scroll](https://github.com/user-attachments/assets/2e72acf4-b908-4c04-b3d4-c5b62351faa7)

rmkaplan:
I think it is a mistake to have 'ddd in the second identifier, probably I should check for stringp and litatom.

[...]

Issue #1949, opened 2025-01-03
TEdit Menus FIELDS:  Using a string as an IDENTIFIER makes the FIELD inaccessible to MB.GET


fghalasz:
(Branch: master)

Using a string as an IDENTIFIER in a FIELD makes the FIELD inaccessible via MB.GET.
Example: 
```
> (MB.ADD '((FIELD (IDENTIFIER "TEST")(PRELABEL "TEST"))) TEXTSTREAM)
3
> (MB.GET "TEST" TEXTSTREAM)
NIL
> (MB.GET 'TEST TEXTSTREAM)
NIL
```

Probably should either ERROR when IDENTIFIER is given as a string or MKATOM the string for all MB.* functions..

rmkaplan:
Fixed in #1957

PR #1957, opened 2025-01-05
Shakedown of field menu items

rmkaplan:
This addresses field menu-item issues:
#1953 
#1952 
#1950 
#1949 

It now validates the IDENTIFIER, raising an error if it is not a string or non-NIL atom.  And it uppercases, for consistency.

MB.ADD has an extra argument INCREMENTALUPDATES.  If defaulted to NIL, the screen will not be updated at all until all the items are added.  If T, the update will happen after each insertion, as before.

The last inserted items is not highlighted at the end of the MB.ADD, and the selection is disabled.


rmkaplan:
I added a small change to to non-field menu items to suppress copy-selecting, cf #1297

fghalasz:
@rmkaplan   The first two arguments to the call to TEDIT.INSERT.OBJECT in the INITSTATE portion of MB.FIELD.CREATE are in the wrong order.

Also, from Notecards standpoint, uppercasing all IDENTIFIERS is not ideal.  In the original code, Property Names are case sensitive.  And I don't know which of the many Notecards extensions and applications depend on that case sensitivity. Currently I am using the Property Names as IDENTIFIERS, so if IDENTIFIERS are all U-CASEd I loose case sensitivity.  My alternative is a relatively complicated mapping where I keep a permanent record of the Property Name to IDENTIFIER map.

rmkaplan:
OK, I’ll fix the argument order.  It’s unfortunate and confusing that the documented argument-orders for TEDIT.INSERT and TEDIT.INSERT.OBJECT are not the same.

I'll take out the U-CASE (I guess I was trying to compensate for my own typos).

Would a pop-up-menu button-type be useful?

I haven’t tried the lfg scripts yet, waiting until I can get an ordinary full loadup to go through.

fghalasz (approved):
The TEdit Menu Fields are all working as expected in Notecards.

Issue #1767, opened 2024-06-17
Resuming lisp.virtualmem has error if connected dir was a pseudohost


MattHeffron:
**Describe the bug**
If the connected directory at `(LOGOUT)` was, or included, a pseudohost, then an error occurs when resuming medley with that `lisp.virtualmem`.

**To Reproduce**
1. I hid my personal `INIT` file.
2. My `.medley_config` contains `-r /home/matt/GREETWRAP` (and other args, file is attached)
3. The shell connected directory is my home: `/home/matt`
4. Launch medley with command: `/home/matt/Interlisp/medley/scripts/medley/medley.command --full` (it's a defined `alias`)
5. Create an Interlisp Exec, and use it for the entries below.
6. `(FILESLOAD (SYSLOAD FROM LISPUSERS) PSEUDOHOSTS)`
7. `(PSEUDOHOST 'MEDLEY MEDLEYDIR)`
8. `(CNDIR MEDLEYDIR)`
9. `(LOGOUT)`
10. Launch medley to resume the saved VM: `/home/matt/Interlisp/medley/scripts/medley/medley.command --k vmem/lisp.virtualmem`
11. Medley resumes correctly with no error, `(LOGOUT)` returns `NIL` 
12. `CD {MEDLEY}`   (which returns `{MEDLEY}`)
13.  `(LOGOUT)`
14. Again resume the saved VM:  `/home/matt/Interlisp/medley/scripts/medley/medley.command --k vmem/lisp.virtualmem`
15. This time it gives the error: `File not found:  /home/matt/GREETWRAP`
16. The error opens in a BREAK window if the `(LOGOUT)` executes in other contexts, such as from a button in the Lispusers BUTTONS module. (This is my typical case.)

**Expected behavior**
Medley resumes without error.

**Screenshots**
![image](https://github.com/Interlisp/medley/assets/8261874/47b60325-7ebd-4828-a1a4-210e34d3f92b)

**Context (please complete the following information):**
 - OS: Ubuntu on Win 11 WSL
 - OS Version: Ubuntu 22.04.3 LTS
 - Host arch: x86_64
 - Maiko version: fd16058ca7debb9e9620b99a512c93937300fb5c
 - IL:MAKESYSDATE: 6-Jun-2024 19:53:48

**Additional context**
My `.medley_config` and `/home/matt/GREETWRAP` files (renamed so github allows attaching):
[...]

PR #1963, opened 2025-01-09
Use INFILEP to get unambiguous path in INTERPRET.REM.CM

MattHeffron:
This is necessary because a path of the form "/home/matt/GREETFILE" is treated as "home>matt>GREETFILE".
OPENSTREAM in INTERPRET.REM.CM tries to open that effectively as relative to (CNDIR T). 
This is seen when resuming from a saved lisp.virtualmem.


hjellinek:
I tried this using macOS.  While connected to `{DSK}<Users>hjellinek>Projects>blablabla>`, I tried something similar:
```
(openstream "/tmp/foo.txt" 'input)
```
The result was a normal input stream on `{DSK}<tmp>foo.txt;1`.

I don't think `OPENSTREAM` treats the path in my example differently than the one in your example, so what accounts for the difference in results?

nbriggs:
For background, in the Alto days (Alto Executive from the Alto OS Reference Manual, 1977):
```
Commands can be separated by semi-colons.
If there is more than one command in a command line, everything following the
first command is saved (after the interpretation described above)
on a file called Rem.Cm, which will be examined the next time the Executive is run.
```

I think that our current problem got started when we switched from having separate "Rem.Cm" file and site init file to overloading LDEINIT for naming both Rem.Cm and site init files.

Piecing the story together from my fragmentary knowledge - if Lisp wanted to chain to another program (or itself), it could append to (or overwrite) Rem.Cm with "Lisp; (something interesting to Lisp)" with the expectation that the Executive would start "Lisp" and stuff the remaining command line back into Rem.Cm -- Lisp was set up to read an appropriate part of the Rem.Cm contents, and being a good citizen, remove it, and let the Alto executive see the leftovers.

Interpreting Rem.Cm, if it existed and contained anything of interest to Lisp, was completely separate from the site and user greetfiles. I think that interpreting Rem.Cm happens at a different time in the startup than the processing of site and user greeting.  Linking the two names together means that you get attempts from *both* places in the startup to do something with the site init file.

```
(GREETFILENAME
  [LAMBDA (USER)                                             (* lmm "13-Apr-84 08:42")
                                                             (* Returns name of an existing greeting file, or NIL)
    (DECLARE (GLOBALVARS USERGREETFILES LOGINHOST/DIR COMPILE.EXT))
    (SELECTQ USER
	     [T (OR (INFILEP (QUOTE {DSK}INIT.LISP))
		    (bind FILE while (SETQ FILE (PROMPTFORWORD (QUOTE 
	       "Please enter name of system init file
(e.g. {server}<directory>INIT.extension): ")))
		       until (SETQ FILE (INFILEP (MKATOM FILE))) finally (RETURN FILE]
	     (NIL)
	     (COND
	       ((LISTP USERGREETFILES)
		 (PROG [(POS (STRPOS (QUOTE %.)
				     (SETQ USER (U-CASE USER]
etc. etc.
```
In the same timeframe, the INTEPRET.REM.CM looked like:
```
(INTERPRET.REM.CM
  [LAMBDA (RETFLG)                                           (* bvm: "22-Aug-84 14:16")
    (DECLARE (GLOBALVARS STARTUPFORM))

          (* * Looks at REM.CM and evaluates the form there if the first character of the file is open paren or doublequote.
	  If it's a string, it will be unread,, else the form will be evaluated at the next prompt. For use in INIT.LISP, 
	  among others. If RETFLG is true, the expression read is simply returned)


    (PROG ([FILE (CAR (NLSETQ (OPENSTREAM (QUOTE {DSK}REM.CM;1)
					  (QUOTE BOTH]
	   COM AUXFILE)
          (OR FILE (RETURN))
          [COND
	    ([AND (IGREATERP (GETFILEINFO FILE (QUOTE LENGTH))
			     0)
		  (SELCHARQ (\PEEKBIN FILE T)
			    ((%( %")
			      T)
			    NIL)
		  (SETQ COM (PROGN (SETFILEINFO FILE (QUOTE ENDOFSTREAMOP)
						(FUNCTION ERROR!))
				   (CAR (NLSETQ (READ FILE T]
	      (COND
		(RETFLG                                      (* Save it to return))
		((LISTP COM)                                 (* make it happen at next prompt)
		  (SETQ STARTUPFORM (LIST (QUOTE PROGN)
					  (QUOTE (SETQ PROMPTCHARFORMS (DREMOVE STARTUPFORM 
										PROMPTCHARFORMS)))
					  (LIST (QUOTE PRINT)
						(LIST (QUOTE LISPXEVAL)
						      (KWOTE COM))
						T T)))
		  (SETQ PROMPTCHARFORMS (CONS STARTUPFORM PROMPTCHARFORMS)))
		(T                                           (* Unread a string)
		   (BKSYSBUF COM)))
	      (\SETEOFPTR FILE (COND
			    ((NOT (\EOFP FILE))
			      (SELCHARQ (\PEEKBIN FILE)
					((CR ;)              (* Eat up the command terminator)
					  (\BIN FILE))
					NIL)                 (* Need to rewrite REM.CM with remainder of text)
			      (SETQ AUXFILE (OPENSTREAM (QUOTE {NODIRCORE})
							(QUOTE BOTH)
							(QUOTE NEW)))
			      (COPYBYTES FILE AUXFILE)
			      (SETFILEPTR FILE 0)
			      (COPYBYTES AUXFILE FILE 0 (GETFILEPTR AUXFILE))
			      (CLOSEF AUXFILE)
			      (GETFILEPTR FILE))
			    (T 0]
          (CLOSEF FILE)
          (RETURN (COND
		    (RETFLG COM)
		    (COM T])
```

MattHeffron:
It just occurred to me that this was "familiar".
This is issue #1767!

nbriggs:
This is how it used to work:
```
// LoadFullFromLisp.cm Edited 5-Sep-85 -- van Melle --

Copy Init.save ← Init.Lisp
Del INIT.LISP!*
FTP/-E ERIS Login/C Dir/C Lisp>koto>basics Ret/<>A Lisp.run DoradoLispMc.eb DolphinLispMc.eb Ret/S Init.nogreet Init.Lisp

Lisp [ERIS]<Lisp>KOTO>BASICS>Lisp.sysout ;"  
SHH(AND (INFILEP '{DSK}INIT.SAVE) (COPYFILE '{DSK}INIT.SAVE '{DSK}INIT.LISP;1) (DELFILE '{DSK}INIT.SAVE))
SHH(LOAD '{ERIS}<LISP>KOTO>CM>LOADFULL.LISP)
SHH(MAKESYS '{ERIS}<LISP>KOTO>BASICS>FULL.SYSOUT]
SHH(LOGOUT T] "
```
So it saves then replaces your Init.Lisp with Init.nogreet, fires up the Lisp sysout, where the Rem.Cm will contain the "..." commands to do what needs to be done... which is to put back the saved copy of the site init file and then load the Lisp file to load full from lisp, etc.

MattHeffron:
Superseded by #2027

Issue #1968, opened 2025-01-14
GREET just wished me "Happy Valentine's Day" a month early


MattHeffron:
`GREET` just wished me "Happy Valentine's Day" a month early!
![image](https://github.com/user-attachments/assets/0b74cdcf-e325-4ae0-8653-a4a525769e60)
`GREETDATES` correctly has Valentines Day as "14-FEB".
In `GREET0` it is using
```common-lisp
(CL:MULTIPLE-VALUE-BIND (SECONDS MINUTES HOUR DAY MONTH YEAR)
  (CL:GET-DECODED-TIME) 
``` 
to get the parts of the date, and then constructing the _day-month_ string to `SASSOC` into `GREETDATES` 
_However_, it is using 
```common-lisp
(CL:NTH MONTH '("JAN" "FEB" "MAR" ; etc
```
to get the month text, but `CL:NTH` is zero-based, so `MONTH` = 1 gets "FEB"

rmkaplan:
I got Lincoln’s Birthday on Sunday.

Issue #1954, opened 2025-01-04
Apps loadup fails with missing CLOS package error


pamoroso:
To test [Maiko PR #523](https://github.com/Interlisp/maiko/pull/523) I freshly cloned the full Maiko repo, as so far on my local Linux system I had initially cloned it with the `--depth 1` option as well as the Medley and Notecards repo. To have matching, full repos of the latter I freshly cloned Medley and Notecards too.

Then I tried to rebuild everything as explained [here](https://journal.paoloamoroso.com/how-to-build-medley-interlisp). The Maiko build completed successfully but the apps loadup `scripts/loadup-all.sh -apps` failed with the error:

```
STACK OVERFLOW
Can't find package CLOS to look symbol LOAD-CLOS
```

This is the error:

![loadup-error](https://github.com/user-attachments/assets/abbf125b-b298-44fb-a277-7ca37c61af5c)

And the console output:

```
[...]
>>>>> START loadup-apps-from-full
"/home/paolo/medley/medley/../maiko/linux.x86_64/lde" "/tmp/loadups-45336/full.sysout" -id "loadup_1" -title "Medley Interlisp :: loadup_1" -g 1024x768 -sc 1024x768 -noscroll
MEDLEYDIR: "/home/paolo/medley/medley"
LOGINDIR: "/tmp/loadups-45336/logindir"
GREET FILE: "/tmp/loadups-45336/loadup-apps-from-full.cm"
VMEM FILE: "/tmp/loadups-45336/logindir/vmem/lisp_loadup_1.virtualmem"

*Error* Stack Overflow, MUST HARDRESET!
Enter the URaid
CL:NIL

< 
```

I'm using amd64 Linux Mint 22 Cinnamon.

pamoroso:
The head of my Medley local repo from `git log`:

```
commit 5b37dd09db4bba3633a5bed30de4bb53a7ced539 (HEAD -> master, tag: medley-241225-5b37dd09, origin/master, origin/HEAD)
Author: rmkaplan <69548581+rmkaplan@users.noreply.github.com>
Date:   Wed Dec 25 13:06:35 2024 -0800
```

masinter:
there is a DRAFT PR that addresses this and other problems. 
But I still haven't figured out why loadup-apps-from-full.sh succeeds in some situations and not in others.   I have two clones of the medley, maiko, notecards, repos that are the same according to "diff" (excluding .git) both under Debian distro and one fails and the other succeeds.

[...]

PR #1969, opened 2025-01-14
Fix an off by 1 error in GREET0.

MattHeffron:
See Issue #1968

This is a one line change on line 2624 (new file). I have *no idea* how the FILEINFO READTABLE was changed from "XCL" to "INTERLISP". That's what's causing all of the other diffs.
(Anyone have ideas why?)


pamoroso (dismissed):
It looks good to me but I still can't build Medley to test the code because of #1954.

nbriggs:
NO.  Don't change the read table.

nbriggs:
@pamoroso - Build full instead of apps

nbriggs:
I'm guessing that there was a MAKEFILE ... 'NEW there?  If you let it loadfrom prop the GREET0 function, add a SUB1 for the month, and then (CLEANUP HIST) it remakes the file with only the one function changed and retains the original XCL readtable, and doesn't mess with the case of all the other code in the file.

MattHeffron:
Replaced by PR #1970 with clean file diffs.

PR #1970, opened 2025-01-14
Fix GREET0 off by 1 error. Without changing the file's READTABLE

MattHeffron:
Replaces mth26 branch (PR #1969). This doesn't change the file's READTABLE.


nbriggs (approved):
Much improved!  Yes, that'll fix the problem.  I think you can merge it, or if you want, wait for any other reviews.

Not to derail this fix, but I do wish we had a process by which we could keep the version numbers incrementing in at least a somewhat orderly manner -- perhaps we should document doing a "medley/scripts/restore-versions.h FOO" before making changes to lisp file FOO - I think that would get things at least a bit better.

PR #1973, opened 2025-01-15
Add a call to check for orphaned versions after any checkout

masinter:


scripts/find-orphaned-versions.sh
```diff
@@ -0,0 +1,9 @@
+#/bin/sh
+# $1 is a versioned file name
+
+ver="$1"
+base="${ver%%.~[1-9]*~}"
```
MattHeffron, line 5: This will fail to get base from versions 0 mod 10.
Is this what's needed: `~[1-9][0-9]*~`

scripts/post-checkout
```diff
@@ -0,0 +1,16 @@
...
+is_branch_checkout=$3
+
+# Print some info
+echo "Checkout complete!"
+# echo "Previous HEAD: $previous_head"
+# echo "New HEAD: $new_head"
+# echo "Branch checkout: $is_branch_checkout"
+
+if [[ "$is_branch_checkout" == "1" ]]; then
+    find . -name "*.~[1-9]*~" -exec ./scripts/find-orphaned-versions.sh {} \;
```
MattHeffron, line 15: Like above re 0 mod 10

scripts/post-checkout.sh
```diff
@@ -0,0 +1,3 @@
+#/bin/sh
+
+find . -name "*.~[1-9]*~" -exec if \[ ! -f {}:h \]\; then echo "{}" " with no versionless" \;
```
MattHeffron, line 3: Like above re 0 mod 10

masinter:
to install this you need to 
```
cd .git/hooks/
ln -s ../../scripts/post-checkout
```

scripts/find-orphaned-versions.sh
MattHeffron: I don't see that this is resolved. What am I missing?

scripts/post-checkout
nbriggs: I fixed that just now and pushed it back to Larry's branch.

masinter: I'm sorry, but the "fix" isn't correct. The -name filter to "find" is not a regular expression, it's a shell filename expansion. So `foo.ext.~5~`  doesn't match `*.~[1-9][0-9]*~` but does match `*.~[1-9]*~`.

nbriggs: I've undone the change I made and pushed it back.

nbriggs (dismissed):
Works for me.

masinter:
I might have fixed the loophole but in any case I can't merge this even though  @nbriggs approved

nbriggs (approved):
OK.   I'd note that the install-repo-checks.sh must be run at the top level of a repository

nbriggs:
Oops - that should be `chmod +x .git/hooks/post-checkout` unless you're specifically trying to have git not use the hook unless the user adds the executable bit.
```
Your branch is up to date with 'origin/master'.
hint: The '.git/hooks/post-checkout' hook was ignored because it's not set as executable.
hint: You can disable this warning with `git config advice.ignoredHook false`.
```
and if you switch from a branch which has the post-checkout hook installed to one that doesn't, you get a lot of
```
find: ./scripts/find-orphaned-versions.sh: No such file or directory
```

nbriggs:
I'm going to revert this.

Issue #1965, opened 2025-01-13
Reading a simple 7-bit ASCII text file with `RSTRING` and `((FORMAT :UTF-8))` produces an odd result


hjellinek:
**Describe the bug**
I have a text file containing only 7-bit ASCII.
I opened the file for input with `((FORMAT :UTF-8))`, called `RSTRING`, and displayed the characters in the resulting string.
The first character is correct.  The second and subsequent characters are blobs.

If I repeat the experiment without `((FORMAT :UTF-8))`, the characters appear correct.

**To Reproduce**
1. Load or paste the file `read-string.lisp.txt`, which is attached.  (GitHub won't allow `.lisp` file attachments.)
2. Copy the file `foo.txt`, also attached, to `/tmp`.
3. Evaluate `(READ-STRING-FROM-FILE "/tmp/foo.txt" T)` -- the `T` means to set the UTF-8 encoding
4. What happened? See [screenshot_760](https://github.com/user-attachments/assets/e6b9721a-f05a-4400-bede-419370002646)

**Expected behavior**
I expected identical results to what you'd see if you evaluate `(READ-STRING-FROM-FILE "/tmp/foo.txt")`, which does not specify UTF-8.
See [screenshot_761](https://github.com/user-attachments/assets/00f07fb6-f5c3-4654-be11-48169c419cf4).

**Context (please complete the following information):**
 - Are you using online.interlisp.org? no
 - OS: macOS
 - OS Version: Ventura
 - Display/window system: SDL
 - Host arch: x86_64
 - Maiko version: `commit 55d8504060d0d40505a2ddc39daca12f9ecb0303`
 - `IL:MAKESYSDATE`: [ date ] or `(il:print-lisp-information)`: 6-Nov-2024

**Attachments**

[foo.txt](https://github.com/user-attachments/files/18401546/foo.txt)
[read-string.lisp.txt](https://github.com/user-attachments/files/18401558/read-string.lisp.txt)

![screenshot_760](https://github.com/user-attachments/assets/e6b9721a-f05a-4400-bede-419370002646)
![screenshot_761](https://github.com/user-attachments/assets/00f07fb6-f5c3-4654-be11-48169c419cf4)

rmkaplan:
I'm having trouble running this.  The function by itself is not a loadable file, but I can get it to load if I put a semi-colon on the first line.  If I try to copy (meta-C/meta-X) into Lisp, the quotes are screwed up.

Can you put it in a loadable file?

[...]

PR #1974, opened 2025-01-16
Slightly better progress reporting in LOADUP.

MattHeffron:
Make it clear when files are skipped because they are already loaded.

This helps address the _apparent_ truncation of `full.dribble`. It wasn't truncated, it just appeared to be. 
See Nick's comment: https://github.com/Interlisp/medley/issues/1965#issuecomment-2594246610


sources/APUTDQ
```diff
@@ -75,7 +77,7 @@ Copyright (c) 1981-1988, 1990, 2021-2022 by Venue & Xerox Corporation.
                      (*COMPILED-EXTENSIONS* '(DFASL LCOM))
                      (SYSOUT.EXT 'SYSOUT]
         (LOCALVARS . T)
-        (DECLARE%: DONTEVAL@LOAD DOEVAL@COMPILE DONTCOPY COMPILERVARS (ADDVARS (NLAMA FAULTEVAL)
+        (DECLARE%: DONTEVAL@LOAD DOEVAL@COMPILE DONTCOPY COMPILERVARS (ADDVARS (NLAMA)
```
nbriggs, line 80: Why did we lose the FAULTEVAL from (ADDVARS (NLAMA FAULTEVAL) here?

sources/APUTDQ
```diff
@@ -132,12 +134,14 @@ Copyright (c) 1981-1988, 1990, 2021-2022 by Venue & Xerox Corporation.
...
-  [LAMBDA (FILES)                                      (* ; "Edited 12-Mar-2021 00:15 by larry")
-    (for X in FILES do (PRINTOUT T "loading " X T)
-                                  (OR (FMEMB X SYSFILES)
-                                      (DOFILESLOAD (LIST '(SYSLOAD)
-                                                         X)))
-                                  (SMASHFILECOMS X])
+  [LAMBDA (FILES)                                            (* ; "Edited 16-Jan-2025 13:35 by mth")
+                                                           (* ; "Edited 12-Mar-2021 00:15 by larry")
+    (for X in FILES do (if (FMEMB X SYSFILES)
+                           then (PRINTOUT T X " already loaded" T)
```
nbriggs, line 140: OK, and we should also fix the loadup files so that it does not generate this note -  we should load UNIXUTILS before we load things that depend on it.

sources/APUTDQ
```diff
@@ -251,91 +255,16 @@ Copyright (c) 1981-1988, 1990, 2021-2022 by Venue & Xerox Corporation.
...
 
-(ADDTOVAR NLAMA FAULTEVAL)
-
-(ADDTOVAR NLAML )
-
-(ADDTOVAR LAMA )
-)
-(PRETTYCOMPRINT APUTDQCOMS)
-
-(RPAQQ APUTDQCOMS
```
nbriggs, line 262: I'm confused about why the RPAQQ of the APUTDQCOMS seems to appear twice in the original file. @masinter ideas?

masinter:
The Interlisp-10 compiler needed to know ahead of time which functions were
NLAMBDA vs. LAMBDA and spread vs nospread, by maintaining LAMA, NLAMA NLAML
(lambda spread, nlambda spread and nospread).

These were set by COMPILERVARS -- if when you makefile you notice come
across a definition that doesn't match, it rewrites the COMS and prints the
COMS again with the corrected value.

I'm not sure what happens with Common Lisp... there aren't any NLAMDAs and
I'm not sure if there are any no-spread?

sources/APUTDQ
MattHeffron: There are, in fact, several "already loaded" modules reported in the `full.dribble`

MattHeffron:
> The Interlisp-10 compiler needed to know ahead of time

Does the Interlisp-D compiler (`IL:TCOMPL`, `IL:BCOMPL`, etc.) _need_ to know ahead of time? 
If it is to know how to compile _callers_ in the same source file that are located earlier in the file, then adding the modified COMS (and setting expressions of `LAMA`, `NLAMA`, `NLAML`) at the end of the file might not help.

If the `MAKEFILE` that updates the COMS is done in one sysout, but the compilation is done in a different sysout, without loading the (modified) source, the updated COMS and `LAMA`, `NLAMA`, `NLAML` wouldn't be known at compile time; unless compiling a file always loads any setting expressions of `LAMA`, `NLAMA`, `NLAML` before compiling anything.

rmkaplan:
Now that the issue as surface, I don’t see how  the NLAMA/L would have ever worked.   If you were compiling a file with a function FOO that called a function FIE defined as an NLAML on some other unloaded file BLAH, then how would it know how to compile that call?  Did this require that information about all the functions called by FOO had to be loaded in advance?

rmkaplan (approved):
LGTM

PR #1978, opened 2025-01-19
Adds STATUS argument to LOGOUT for process exit status

nbriggs:
Combined with Interlisp/maiko#525 allows passing an exit status (typically 0..255) as an optional second argument to LOGOUT.

The underlying Maiko implementation will accept any integer value (or NIL) and pass it to the system exit() procedure, however the behavior for numeric values outside the range 0..255 is typically OS dependent and often unspecified.  The default exit code is 0 (EXIT_SUCCESS). Passing non-numeric values (other than NIL) will result in an exit with status `EXIT_FAILURE`, typically defined by the OS as 1.

While both the change in this branch and the corresponding maiko branch are required for full functionality, they can be deployed in either order -- old sysouts will not change behavior on the newer maiko emulator, and new sysouts will run on an older emulator but the exit status will be ignored.


fghalasz (dismissed):
I merged into this branch a change to \LOGOUT0 that ensures that the SUBRCALL LISPFINISH is called with only a single argument if STATUS is not a SMALLP (e.g., in the (LOGOUT) and (LOGOUT T) cases).  If STATUS is a SMALLP, then the SUBRCALL LISPFINISH has 2 args.  This matches the logic in the Maiko PR#525 lispfinish subr.  Without this change, (LOGOUT) and (LOGOUT T) will cause lde to exit with status code 1 (and not 0).

With this change all seems to work OK.  Tested only on WSL/Ubuntu22.04.

nbriggs:
I've reverted the last change to ADIR and updated the Maiko code to have the correct check.

nbriggs:
Git thinks that my ADIRs conflict - but they don't, since I incorporated Ron's changed ADIR.

fghalasz (approved):
LGTM.

PR #1979, opened 2025-01-21
Expand PSEUDOHOST as destination in FILECREATED info written

MattHeffron:
This change uses TRUEFILENAME to expand the FULLNAME of the output stream.

Some files are being saved with PseudoHosts in the FILECREATED expression. The PseudoHost will prevent some utility scripts from correct behavior.

(E.g., restore-versions.sh looks for "{DSK}" to identify the file path and get the Medley file version number. This won't fix every case, for example if the PseudoHost is defined in terms of "{UNIX}", restore-versions.sh still wouldn't identify the file version.)


masinter (commented):
I don't think we can resolve the problems of version numbers and files through a sequence of patches like this one, which doesn't address the fundamental problem that different people will have different version numbers in the file system than in the FILECREATED expressions.

rmkaplan:
The pseudohost part of the name in the filecreated expression is an approximation to a logical path name that in principle should abstract away from the full pathname that includes the location of the clone in the originator’s file system.  The name should be good when the file is downloaded into a clone on another system.  

But the filecreated version numbers are system-dependent, as noted, and my style of copying (as opposed to remaking) files from my working pseudohost {WMEDLEY} to my clone pseudohost {MEDLEY} also screws that up.  

I’m not sure what kind of consistency guarantee we should or can provide between what is written in the FILECREATED  expression (version, fullpath/pseudopath…) and the name/location of the file as it appears when downloaded to a new system.

MattHeffron:
> I don't think we can resolve the problems of version numbers and files through a sequence of patches like this one

I agree that this is a _stop gap_ solution so that `restore-versions.sh` has a better chance of working. A complete solution would be much better, of course.

masinter:
What is the problem? 
The version numbers in the FILECREATED expression are meaningless for anyone not running in the clone of the repo used by anyone else. 
We can support version numbers, but this PR doesn't use absolute version numbers but rather relative ones, and that helps.

Maybe we could work out some use cases? the main one I want to support is to ensure that if I make a new clone of the repo or remove all  version files (`find . -name "*~" -exec rm {} \;`) that the file package will not assume that version numbers are valid. Right now the ilemaps are cached by the version numbered file names. Also whereis.hash when it reopens should get the latest version number.

MattHeffron:
Per discussion 1/27, will not work as intended. Closed without merging.

PR #1980, opened 2025-01-21
Fix move of PSEUDOHOSTS from lispusers to library

MattHeffron:
The files were **copied** from `lispusers` to `library`, and then modified there.
They should have been **moved** so the git history was preserved.
I replaced these files _with themselves_ by a 3 step process that _appears_ to have gotten the git history to be correct.
(And the `PSEUDOHOSTS.TEDIT` from `lispusers` was also moved. It had been left behind previously.)


nbriggs (approved):
Looks fine.

PR #1984, opened 2025-01-22
Fix a glitch in Tedit screen-update

rmkaplan:
I don't know if this addresses the pointer problems I've seen, those are nondeterministic, and I haven't yet understood them.  But I saw this when trying to track down that problem, and after this fix I haven't had a crash, so fingers crossed.


masinter (approved):
1/27/25 meeting -- taking Ron's assurances -- hasn't crashed

Issue #1985, opened 2025-01-22
Meta-, does nothing in TEdit with an Italian keyboard on Linux


pamoroso:
**Describe the bug**

In TEdit, pressing `Meta-,` moves the caret left and `Meta-.` right by one character. But pressing `Meta-,` on [my italian keyboard](https://i.snap.as/wAoI3ouY.jpg) does nothing and the caret doesn't move. `Meta-.` works as expected.

**To Reproduce**

1. launch Medley
2. open TEdit
3. type one word
4. place the caret at the center of the word
5. press `Meta-,` on an Italian keyboard

**Expected behavior**

The caret moves left by one character.

**Screenshots**

N/A

**Context (please complete the following information):**
 - Are you using online.interlisp.org? no
 - OS: Linux
 - OS Version: Mint 22 Cinnamon
 - Display/window system: X11 + VNC
 - Host arch: x86_64 (System76 Merkaat)
 - Maiko version: 3c4d9f5393e56857d1e75352f8d8d0a0c9d91e01
 - `IL:MAKESYSDATE`: 21-Jan-2025 04:48:26

**Additional context**

The output of `xev` when `,` is pressed:

```
KeyPress event, serial 28, synthetic NO, window 0x4200001,
    root 0x5cf, subw 0x4200002, time 39556239, (45,22), root:(110,825),
    state 0x0, keycode 59 (keysym 0x2c, comma), same_screen YES,
    XLookupString gives 1 bytes: (2c) ","
    XmbLookupString gives 1 bytes: (2c) ","
    XFilterEvent returns: False

KeyRelease event, serial 28, synthetic NO, window 0x4200001,
    root 0x5cf, subw 0x4200002, time 39556351, (45,22), root:(110,825),
    state 0x0, keycode 59 (keysym 0x2c, comma), same_screen YES,
    XLookupString gives 1 bytes: (2c) ","
    XFilterEvent returns: False
```

This issue may be related to the behavior of the `<` key on my keyboard described in #1828.

rmkaplan:
The default Tedit readtable arms both Meta-, and Meta-<  to go back one character.  If the Italian keyboard doesn’t produce the expected key-bits for those Meta combinations, the backup won’t be triggered.

If you know what bits are showing up for those combinations, then you could add those to the Tedit readtable using TEDIT.SETFUNCTION.  Something like

(TEDIT “Meta,???" (FUNCTION \TEDIT.ONECHAR.BACKWARD))

[...]

PR #1988, opened 2025-01-24
TEDIT-FNKEYS makes sure that the new caret-point after onechar forward/backward is always LEFT

rmkaplan:
I hope this fixes the observations in #1985 , although perhaps there is still an Italian keyboard issue.

The odd behavior with respect to the selection had to do with the fact that the point of the resulting selection dependede on the point of the input selection, e.g., whether the caret was where it was because of a click after the middle of the character before (RIGHT) or a click before the middle of that character (LEFT).  Now the result is always LEFT


nbriggs (approved):
This works much better.  There are some circumstances, not sure if they are pre-existing, and I can't reproduce it on demand, where having one or more characters blue-pending-delete and moving around with meta-whatever and then reselecting to generate another blue-pending-delete gets the inversion incorrect, leaving an inverted character.  Redisplay fixes it.

rmkaplan:
I played around with meta-< and meta-> with blue-pending-delete selections and I haven’t seen anything odd wrt the inverted display—it always goes back to normal and then the caret moves.  Depending on where the caret is to begin and when forward or backward, the caret might move into what was the inverted region, but that seems right.

I have seen something that doesn’t seem quite right with BPD word selections, but not related to forward or backward selection.  If you select word by clicking the middle button to the left of the mid point of a word, the word inverts and the caret moves to the beginning of the word, which is good.  But then when you click the right button at the same position in that word, the selection is extended vacuously—it doesn’t grow or contract—but the caret move to the right side.

Shouldn’t the caret stay on the left, if you click towards the left?

nbriggs:
I can get the BPD problem to happen occasionally, but I agree it's probably not the meta-movement that's involved.  Is there a way to dump TEdit's idea of exactly what I've done, after it happens, in a way that will help you debug it?

rmkaplan:
Possibly the history list.

If you load TEDIT-DEBUG.LCOM, then immediately after you see the issue, you can execute (IHIST) (maybe with the mouse over the target Tedit window, if there are several on screen).

The last event should be the top of the inspector, you should at least see what you did and where.  But it may be more internal.

PR #1994, opened 2025-01-29
fix GITFNS  prc command

rmkaplan:
I think I broke the non-menu single-PR case when I put in the JSON interface and the menu.  prc of a string that matches a single PR should do that comparison without bringing up a menu.  If it matches more than one, the menu should still come up.  E.g prc rmk should bring up a menu of all my PRs.


rmkaplan:
Should fix #1991

pamoroso:
I tested this PR and now `prc mth3--Extend-file-COPYRIGHT-for-succesive-owners drafts` works as expected with no errors.

The PR fixes #1995 as discussed in #1991.

pamoroso (approved):
LGTM.

PR #1998, opened 2025-01-31
Improved REGIONMANAGER reference coordinates

rmkaplan:
Additional ways of specifying the anchor for the reference (window, region, position)


fghalasz (approved):
LGTM

PR #1999, opened 2025-01-31
PSEUDOHOSTS:  Optional argument to PSEUDOHOST creates a specialized cd command

rmkaplan:
For example, (PSEUDOHOST 'NOTECARDS "[notecards path]" 'n) creates the command cdn, and
   cdn  connects to {NOTECARDS}
   cdn system connects to {NOTECARDS}/system/


rmkaplan:
I looked more carefully at what happens when PSEUDOHOST can't resolve a prefix to a real host, either error or return NIL if new argument NOERROR is set.

masinter:
I'm a little worried about the proliferation of short commands 'cdn' that are not of widespread utility.
I know each is relatively harmless, but .... who would use this feature?

Part of my concern is the ambiguity of typing a variable name to EVAL it vs a short command with the same name. It's only ambiguous when there's a command that makes sense when the command name by itself with no other items on the line. 

Also I think we want to move the extra repos (test, loops, notecards) to be subtree or submodule of medley.

rmkaplan:
Since I rotate between a bunch of pseudohosts (clone, working medley, release, lfg, pargram english/german, idl, old tedit, tedit tests…) I use these all the time.

rmkaplan:
And by the way, it’s optional:  if you don’t pass the argument, you don’t get the command.

nbriggs:
Isn't this the sort of thing that you could have outside of the PSEUDOHOSTS code?  "cd {NOTECARDS}" will do as expected so adding command aliases in your init file for each of the "cd<foo>" that you want should be trivial without having it in the PSEUDOHOST arglist, shouldn't it?

pamoroso:
I did some light testing of this and it works as expected.

masinter:
3/3/25  meeting ... will wait on doing this.

PR #2002, opened 2025-02-01
FONTSAMPLE add option for which charset to display

MattHeffron:


library/FONTSAMPLE
```diff
@@ -1,30 +1,38 @@
...
-                           (FNS FNT.MAKEBOOK FNT.LESSP FNT.SORTP FNT.DISPLOOK FNT.DISPTBLE 
-                                FNT.DISPDSCR FNT.NARRDSCR FNT.DISPINIT FNT.FACEMAP FNT.SIZEMAP 
-                                FNT.MAKENAME FNT.MAKEWIND FNT.FILEMAP FNT.FINDALL FNT.FLST)))
+(RPAQQ FONTSAMPLECOMS
+       ((MACROS IDIVUP)
+        (VARS FNT.PANEL FNT.FNAME FNT.INFOFONT FNT.OUTFTEXT)
+        (FNS FNT.MAKEBOOK FNT.LESSP FNT.SORTP FNT.DISPLOOK FNT.DISPTBLE FNT.DISPDSCR FNT.NARRDSCR 
+             FNT.DISPINIT FNT.FACEMAP FNT.SIZEMAP FNT.MAKENAME FNT.MAKEWIND FNT.FILEMAP FNT.FINDALL 
+             FNT.FLST)
+        (DECLARE%: EVAL@COMPILE DONTCOPY (FILES (FROM LOADUPS)
```
nbriggs, line 24: It makes me uncomfortable to introduce a dependency on "LOADUPSDIRECTORIES" - if it really needs EXPORTS.ALL (does it?) then just (FILES EXPORTS.ALL) and it should use DIRECTORIES which should be set up in the site init file.

nbriggs (commented):
See comment attached.

Do you know about lispusers>fontsampler (which already had character sets as an option)?

MattHeffron:
> Do you know about lispusers>fontsampler (which already had character sets as an option)?

I did not. I used FONTSAMPLE to check the BDF font conversion because you had recommended it to me back in December.
This can be closed without merging if that already has this functionality.

library/FONTSAMPLE
MattHeffron: So, **_in general_**, how should dependency on system level definitions (e.g., `CHARSETINFO`, `FONTDESCRIPTOR`, ...) be set in the `COMS` for modules in LIBRARY or LISPUSERS? The pattern:
```common-lisp
(DECLARE%: EVAL@COMPILE DONTCOPY (FILES (FROM LOADUPS)
                                                EXPORTS.ALL))
``` 
I found in `UNICODECOMS`. (@rmkaplan)

masinter:
i'm confused about library/FONTSAMPLE vs. lispusers/FONTSAMPLER as they seem to be very different.
Also, neither one seems to print out an (enlarged) bitmap from the DISPLAYFONT -- I can't quite tell ...

Anyway, it would be good to at least have a description of what's going on.

library/FONTSAMPLE
masinter: I think the convention in the past was that those working on the "low level" system or calls to it would load the file SYSEDIT; SYSEDIT sets some variables as well as loading EXPORTS.ALL if needed. (There was also a file called "ABC" which looked like SYSEDIT and I tried to merge the two.

nbriggs:
Yes, they're different.  I wrote FONTSAMPLER in 1985 because FONTSAMPLE didn't do what I wanted.
Documentation on lispusers/FONTSAMPLER.TEDIT

(FontSample (FONTCREATE 'CLASSIC 12) 239 NIL 'DISPLAY)

will prompt you to sweep out a window in which it will display character set 239 of CLASSIC-12.
You can give it a printer name and type if you want to create a Postscript or Interpress stream.

FONTSAMPLE was expressly intended to create books of font samples on Interpress printers for the fonts known to Lisp and the printers.
Documentation on library/FONTSAMPLE.TEDIT

library/FONTSAMPLE
MattHeffron: See also related discussion in [PR 2003](https://github.com/Interlisp/medley/pull/2003#issuecomment-2629637116)

MattHeffron:
> Yes, they're different. I wrote FONTSAMPLER in 1985 because FONTSAMPLE didn't do what I wanted.

Alas, it cannot create a PDF file showing the DISPLAYFONT glyphs (bitmaps), which would be really handy to check out my BDF to DISPLAYFONT conversion. So I added that ability: PR #2007 😉

library/FONTSAMPLE
MattHeffron: @masinter 
> I think the convention in the past was that those working on the "low level" system or calls to it would load the file SYSEDIT

But ... SYSEDIT does:
```common-lisp
(RESETVARS ((CROSSCOMPILING T))
           (FILESLOAD (SOURCE FROM LOADUPS)
                  EXPORTS.ALL))
``` 
So why wouldn't it be acceptable for individual modules (e.g., on **lispusers**) to do a similar `FILESLOAD` under `(DECLARE%: EVAL@COMPILE DONTCOPY`. (Without the `(RESETVARS ((CROSSCOMPILING T))` which is SYSEDIT-specific.) This also wouldn't change other system-level `VARS` that a particular user might not want changed.

More discussion? @rmkaplan @nbriggs

nbriggs: @MattHeffron it only does it exactly like that because Larry changed it to do the FILESLOAD "FROM LOADUPS".  It's the "FROM LOADUPS" that I'm objecting to, rather than having the right things on DIRECTORIES and not the fact of loading exports.all.

MattHeffron: But ... where did `LOADUPSDIRECTORIES` come from? 
I take it that you're proposing that it be removed from the standard loadup sysouts?

nbriggs: It (`LOADUPDIRECTORIES`) is set in sources/MEDLEYDIR -- and while I'm happy to have the *mechanism* there (though I think MEDLEYDIR should be in library not sources), I don't believe that that is the right place to set up a lot of values that are not actually a fundamental part of the system - that's the role of the site init (site greet) file.

MattHeffron: So, instead of a `LOADUPDIRECTORIES` variable, would it be _better_ (_acceptable_) for individual modules (e.g., on **lispusers**) to have:
```common-lisp
           (FILES (SOURCE FROM VALUEOF (MEDLEYDIR "loadups>")) EXPORTS.ALL))
``` 
Should the location of **EXPORTS.ALL** be a **_known_** place? 
The loadup sysout files are in the **loadups** directory because the **_MEDLEYDIR_/scripts/medley/medley.command** expects to find them in **loadups**. 
Why not have the same _a priori_ condition that **EXPORTS.ALL** is in the same place?

Or, should modules like these have a **_MODULENAME_-SETUP-TO-COMPILE** function that _does_ the setup, or _prints_ the setup instructions? 

And, is it generally better for modules to load **EXPORTS.ALL** (somehow), or should they use `LOADCOMP` of the specific system module(s) needed? E.g., TEDIT has:
```common-lisp
(DECLARE%: EVAL@COMPILE DONTCOPY (FILES (LOADCOMP) UNICODE))
``` 
I'm _not_ trying to be difficult!

Should this discussion be extracted into a **GitHub Issue _or_ Discussion**?

nbriggs: I believe it should just have `(DECLARE: EVAL@COMPILE DONTCOPY (FILES EXPORTS.ALL))` - MEDLEYDIR and anything like that has anything to do with running on the emulator, unless the module is specifically targeted to something Unixy, should not appear in the module.  The locations for everything should be set up in the site greet file and appear on DIRECTORIES, which FILESLOAD and friends already know about.  I think that an EXPORTS.ALL that corresponds to the files that are in `sources` and `library` should be checked in to `library`, or at least it should appear there in the assets of the release.  I don't know about TEDIT and UNICODE.

PR #2003, opened 2025-02-01
Fix unbound vars errors in WRITESTRIKEFONTFILE from earlier edit.

MattHeffron:


rmkaplan:
DUMMYINDEX etc should probably be a constant.  And in fact it is, in EDITFONT.  That declaraton probably should be moved to FONT.

MattHeffron:
Moved those two constants from EDITFONT to FONT, per @rmkaplan.

rmkaplan:
I think the constants BITSPERWORD and BYTESPERWORD should also be removed from EDITFONT.  Those are in MODARITH and EXPORTS.ALL

MattHeffron:
> I think the constants BITSPERWORD and BYTESPERWORD should also be removed from EDITFONT. Those are in MODARITH and EXPORTS.ALL

So, if removed, should **EDITFONT** add **MODARITH** to the `LOADCOMP`? I.e., do 
```common-lisp
           (FILES (LOADCOMP) FONT MODARITH))
```

At this point, this is no longer part of fixing `WRITESTRIKEFONTFILE` in file **FONT**, but is more general _clean up_.

rmkaplan:
No, EDITFONT should load EXPORTS.ALL for compiling, that brings in the constants from MODARITH.

MattHeffron:
> No, EDITFONT should load EXPORTS.ALL for compiling, that brings in the constants from MODARITH.

See the discussion on [PR 2002](https://github.com/Interlisp/medley/pull/2002#discussion_r1938358921) that _strongly implies_ **not** to automatically load EXPORTS.ALL, even under `(DECLARE%: EVAL@COMPILE DONTCOPY` .
Loading EXPORTS.ALL _here_ makes the `LOADCOMP` of FONT superfluous.

How do we want these kind of dependencies to be "taken care of"? 
_Precisely_, with `LOADCOMP`, like EDITFONT is currently? 
Or _wholesale_, by loading EXPORTS.ALL?
Or, _manually as required_, by loading SYSEDIT?

nbriggs (approved):
Approved per Monday Feb 3 2025 meeting.

Issue #1942, opened 2024-12-29
Fontclasses in Tedit and in Charlooks menu


rmkaplan:
Tedit allows fontclasses (DEFAULTFONT, BOLDFONT...) to be installed as character looks, and without further specification (e.g. a FONT entry on TEDIT.DEFAULT.PROPS, the user's DEFAULTFONT fontclass is the font for new characters when an empty Tedit window is opened.   A fontclass can also be asserted programmatically via the TEDIT.LOOKS function, and retrieved by the TEDIT.GET.LOOKS.

But character looks are also and more typically shown and asserted via the various buttons and fields of the charlooks menu.

A fontclass differs from an ordinary font in that it can have different particular fonts for display vs. the various hardcopy devices (pdf, postscript, interpress).  For example, a class with size 12 for display readability might have size 8 for hardcopy.  That raises the question whether/how the variant fontclass information should be made visible and settable by the user.

I initially thought that showing the fontclass name (like DEFAULTFONT or BOLDFONT) in the list of font-families would address that, since that would signal the existence of a class that could then be inspected to get more information.  And by interpreting an atom like BOLDFONT in the Other font field, the user could use the menu to install a fontclass.  That's how I set it up in Fifth Round #1857, with the additional convention that the other properties (bold/italic/size) shown in the menu would be taken from the display component of the class, even if those were different from the properties of all the hardcopy variants.

But this is not a good solution.  Suppose you show some characters in, say, the DEFAULTFONT class, you use the menu to specify a new size, and apply it.  Which if any of the device-fonts should be changed, especially if they are not all the same to begin with?  Or, suppose you put a file with some characters with your DEFAULTFONT to a file, and someone else with a different DEFAULTFONT gets that file.  Should the menu still show DEFAULTFONT?  If it does, then inspecting the current DEFAULTFONT wil only give you misleading information about what is actually controlling the formatting (both display and hardcopy).

There may be a different menu design that will make clear all of the relations and properties. Perhaps putting fontclasses as a separate type of button that pops up a submenu with just the properties of the selected class, and preventing the family/face/size buttons from doing anything to fontclass characters.  Suggestions are welcome.

[...]

PR #2004, opened 2025-02-02
Tedit: eliminate some CHARLOOKS fields, change Charlooks menu, add pagenum object, a few cleanups

rmkaplan:
This update was initially prompted by the discussion in #1942 , the question of how Tedit deals with fontclasses (DEFAULTFONT) and more generally, how the user can specify that a given "look" can be realized by different fonts for different hardcopy display devices (Interpress, postscript, pdf).

I discovered that the CHARLOOKS data type had separate fields for bold, italic, size, and family, that there was a certain amount of effort to keep those consistent with the display font...and that those fields were actually never used.

So this update removes those fields and the code that was maintaining them.

With respect to specifying different fonts for different devices, this update puts a new "device" selector in the charlooks menu, so that the user can specify that a font change is to apply to all devices, or one particular device (I did it as an N-way, maybe the devices should be separate toggles--TBD).  The menu is also rearranged to group the font stuff away from some other toggles.

This update also implements a page-number image object in anticipation of a more general way of specifying page numbers, essentially by putting the page image object in a heading paragraph.  That would enable all of the formatting tools to be applied to page numbers, and eliminate all of the special "folio" code.  And simplify the page-layout menu by eliminating the special page-number font stuff.  But there are still interface/menu design issues to be resolved.

Finally, this includes a few bug fixes and a bit of cleanup.


masinter (approved):
2/3/25 meeting decide to accept
includes definition of page number image object

PR #2009, opened 2025-02-03
Move lispusers fontsampler to obsolete -- only good for Interpress printer fonts. library fontsample (no R) instead.

masinter:


MattHeffron:
**NO** this is moving the **wrong** thing to **obsolete**.
The "no R" (`FONTSAMPLE`) is the obsolete one. 
`FONTSAMPLER` (with R) is the one to "keep".

PR #2012, opened 2025-02-04
saved fix ethernet

masinter:
- **apply ethernet related \MAIKO.MOVDs and rename \MAIKO fns and vars**
- **leave functions for setting time**
- **Remember old (non-working) NS.SETTIME in case we ever want to resurrect it**


masinter:
I'm going to take another tack -- closing this PR.

PR #2013, opened 2025-02-04
FONTSAMPLER can show DISPLAYFONT on non display stream (2nd retry)

MattHeffron:
Enable **FONTSAMPLER** to display glyphs from **.DISPLAYFONT** (bitmap font) on non-`DISPLAY` stream (e.g., **PDF**)
Change to match 2025-02-03 discussion on Ascent/Descent per-`CHARSETINFO` vs. whole `FONTDESCRIPTOR`. 
(Use whole `FONTDESCRIPTOR`.)

Third try at a clean PR.
I checked out **master** and did a Pull.
I then created this new branch, copied the 2 files into **lispusers** (replacing the ones there), committed the changes, and created this PR.

PR #2008, opened 2025-02-03
FONTSAMPLER display glyphs from bitmap font to PDF (retry)

MattHeffron:
Enable FONTSAMPLER to display glyphs from DISPLAYFONT (bitmap font) on non-DISPLAY stream (e.g., PDF)
Added .LCOM to repository.

Corrected PR.

PR #2014, opened 2025-02-05
Yet another attempt to make a clean PR (compared with broken PRs #2007 & #2008 & #2013)

MattHeffron:
Enable **FONTSAMPLER** to display glyphs from **.DISPLAYFONT** (bitmap font) on non-`DISPLAY` stream (e.g., **PDF**)
Change to match 2025-02-03 discussion on Ascent/Descent per-`CHARSETINFO` _vs._ whole `FONTDESCRIPTOR`.
(Use whole `FONTDESCRIPTOR`.)


nbriggs (approved):
This looks good.  It has no unexpected commits, just a single commit with the FONTSAMPLER files that are expected to change.

PR #2018, opened 2025-02-06
FONTSAMPLER - Enable option to use HEX vs OCTAL on page info

MattHeffron:
Enable option to use HEX vs OCTAL for charset number (in page title) and grid row/column titles.
Updated doc file.


pamoroso (approved):
I tested this PR and it works as described. In the sentence "If Hexadecimal is non-NIL, [...]" of the TEdit documentation the function parameter Hexadecimal is not in italics, whereas other parameters in the same document are in italics.

MattHeffron:
> In the sentence "If Hexadecimal is non-NIL, [...]" of the TEdit documentation the function parameter Hexadecimal is not in italics, whereas other parameters in the same document are in italics.

Addressed in latest commit: 427fe60. No code changes.

pamoroso (approved):
It still works fine.

PR #2023, opened 2025-02-07
lispusers/DOCUMENT cleaned up and working with current Tedit

rmkaplan:
The utility of this depends on how careful and thoughtful were the comments in the target source file, but it could be a start for editing up something useful.

I replaced ARGLIST with SMARTARGLIST, and I also made it skip over various formats of edit dates (although I didn't pick out a comment that may be inside an editdate).

I simplified the code quite a bit, eliminating most of the TEDIT.INSERTS by straightforward printout commands.  This revealed a missing functionality in Tedit, that there is currently no way to specify the paragraph looks to be used when you print at the end-of-file.  TEDIT-PARALOOKS only operates on existing paragraphs.  TBD.


pamoroso (approved):
I tested the patch with both Stringscope and DOCUMENT itself and everything seems to be working fine.

pamoroso:
Is DOCUMENT supposed to work with Common Lisp and CLOS? When I evaluate `(Document.Create 'NEW-CLOS-BROWSER)` to process medley/clos/NEW-CLOS-BROWSER I get an error like:

```
TYPE-MISMATCH
In CDR:
#<Compiled Function CLOS-BROWSER:BROWSE-CLASS/131.4464>
```

rmkaplan:
I don’t think so.  I think it is driven off the FILECOMS of the file, so should only work with managed files.

More work would have to be done to gather the functions/variables/etc and top-levelr comments from an unmanaged file.  I think this was a pre-commonlisp hack, I just made it work again.

pamoroso:
You're right, I had forgotten NEW-CLOS-BROWSER is not a managed file. DOCUMENT does work with managed Common Lisp code, for example INSPHEX.

PR #2027, opened 2025-02-10
Restore REM.CM to be separate file from greet file; Enable chaining of medley runs using REM.CM file

fghalasz:
This PR does the following:

1.  Fixed INTERPRET.REM.CM so that it no longer tries to load the file pointed to by LDEINIT and instead loads the file pointed to by LDEREMCM.  LDEINIT remains the file used by greet.    REM.CM file format remains as before - must start and end with double-quote character so file can be read as a string.  Note that the REM.CM file is eval'd AFTER greet processing.

2.  Adjusted the medley script to have a new argument -cm (or --rem.cm) which takes the name of a "REM.CM" file and sets LDEREMCM as appropriate before launching lde.    If  medley is called without -cm or -cm has value "-", no "REM.CM" file is used.

4. Updated the loadup scripts as required to use this new -cm argument when invoking medley.  

5. Repeat feature added to medley script:  Added a new feature to the medley script.  The -cc (or --repeat) argument specifies a REM.CM file.  The -cc  flag signals that when the initial medley run finishes, it should check for the -cc file.   If that file exists (and is not zero-length), medley is run again (i.e., repeated) with LDEREMCM set to the  -cc file.  This repeats until the -cc file no longer exists or is zero-length.  Inside Medley, the -cc file can be found using (UNIX-GETENV  'LDEREPEATCM) so that each invocation of medley can modify (or delete) this file so as to change the subsequent run of medley or terminate the chain.

Except for the -cm changes, loadup scripts remain unchanged in this PR.  A subsequent PR will include changes to the loadup scripts that we've been working on.


masinter (approved):
discussed 2025 92 24 neet.

PR #2030, opened 2025-02-12
UNPACKFILENAME.STRING deals with TOPS20 versions

rmkaplan:
This replicates the behavior in the Venue sysout, as best as I can determine, if OSTYPE is TOPS20.

If there is one dot in the name, it sets off the extension.

Two dots: the extension and version

More dots:  the last 2 are the extension and version, the others go into the name.

If also a ;T, return includes TEMPORARY T.

Not sure what the right behavior is for other   ;FOO.  Venue returns a second version (which will presumably never be seen by a LISTGET), this throws it away.


MattHeffron:
> Not sure what the right behavior is for other ;FOO.

Ignore any irrelevant or non-standard attributes; especially since we don't need/use any information

nbriggs:
I agree -- ignore other attributes.  It's fine to return the (TEMPORARY T) for ;T but Unix systems don't have a built-in concept of files that get flushed at logout, so it's not terribly useful.

rmkaplan:
Simpler code to just flush it altogether, if nobody cares:  For TOPS20, ignore ; and everything after it.

masinter:
I think this request might have come through the attempts to run Eurisco / AM from Doug Lenat without making changes, but i'm nto sure. Without  a clear example of the problem it seems likely that we'll get it wrong.

It's not clear to me that the "tenex/tops-20" file system will ever be used for any files we try to access.

PR #2034, opened 2025-02-15
Update Tedit field names in Lafite

rmkaplan:
Lafite is not in the fuller ms database, so I missed these references in TEDIT-INDENT.

But note also:  there are 2 definitions of the function \GV.PARSERECIPIENTS1, one on lafite/TEDIT-PRIVATEDL and one on internal/MAILCLIENT.

The one on TEDIT-PRIVATEDL references the variable ADDRESSPARSERRDTBL, which is not initialized.


nbriggs:
You mean LAFITE-PRIVATEDL.  At the time I created LAFITE-PRIVATEDL I think that ADDRESSPARSERRDTBL was defined somewhere in LAFITE, and my code modified the existing value.  Subsequently, Bill incorporated the private DL syntax into the regular Lafite mail client, and changed the way he did the parsing so it no longer used that read table.

rmkaplan:
Do you think it still uses that file?  Maybe it show to obsolete.

rmkaplan:
I removed a few more dependencies on Tedit compile-time declarations.  I think what remains is reference to LINEDESCRIPTOR fields in LAFITE-INDENT (a separate utility contributed by Stan Lanning).  But looking at this, I don't think it would work as seems to be intended.  It is confusing the characters of a selection with the lines of the subset of those characters that are visible in any particular window pane.  It will result in an error if the first or last characters are not visible in the first pane, e.g. because of scrolling.

masinter (approved):
should move LAFITE-INDENT to obsolete? In the meanwhile, open an issue to fix up the features so they deal with wrapped text which has linefeeds.

PR #2045, opened 2025-02-23
PDF to {NULL} only does postscript part

rmkaplan:
TMAX does hardcopy to {NULL} in order to determine page numbers for table of contents and index.  The postscript part works fine, but there is then nothing to be converted from postscript to pdf


nbriggs (commented):
That's better - but what if you make a pseudohost with prefix `{null}`?  Under what circumstances should code be making decisions based on the host name without using TRUENAME ?

rmkaplan:
> That's better - but what if you make a pseudohost with prefix `{null}`? Under what circumstances should code be making decisions based on the host name without using TRUENAME ?

Good point, I'll fix it

PR #2047, opened 2025-02-26
update LispUsers  xerox-to-xbm to at least be cl:compiled  (DFASL not LCOM) and TXT -> TEDIT. 

masinter:


pamoroso:
Now the function `xerox-to-xbm` gets defined and can be called.

pamoroso (approved):
I tested it and it works.

PR #2049, opened 2025-02-26
TEDIT: Fix image object glitch plus minor extensions

rmkaplan:
This fixes a misplaced argument in TEDIT.INSERT.OBJECT.

Also extends the COLLECT? argument of TEDIT.MAP.OBJECTS so the caller can specify what the return value should be (OBJECT, CH#, VALUE, FIRST), any other non-NIL gives old behavior (CH# OBJECT VALUE).

The HOSTSTREAM of an image object is passed to the DISPLAYFN but it was not being passed to the IMAGEBOXFN.  It is now also passed to the IMAGEBOXFN as an additional argument.

Unrelated to image objects, I added an additional argument QUIET to TEDIT.FORMAT.HARDCOPY.  If T, the promptwindow isn't updated with #pages and destination file--the caller then has separate control of the message.  (Motivated by the dummy hardcopy needed for TMAX table-of-contents and index files.)


pamoroso:
I'm testing this also with applications that use image objects extensively such as NoteCards and the EQUATIONS LispUsers module. So far everything is working fine.

pamoroso:
I tested up to 8d0cf13 and used SEdit quite a bit. Nothing unusual to report.

rmkaplan:
The Sedit issues are longstanding, one of them just happen to bite when I saved and then committed one of the files.

I think the general problem is that code unknown to or outside of Sedit can make changes to the structure represented on screen that Sedit doesn't know about and thus those changes aren't reflected on the screen.  for example, if an open function is executed, DWIM might alter the structure and Sedit doesn[t notice.  Changes to bindings in iterative operators may look good on the screen but the structure is wrong.

The example that got me here I think is something in EDITDATE.  You exit SEdit with ctrl-X, but then edit date goes in and puts the new date on top, modifying the structure.  But the window doesn't show the new date.    In some circumtances that I have not been able to pine down, the whole structure disappears, but the window looks good.  You have to close th Sedit window and re-edit the function in order to see that this is the case.  That's what got me here, in one of my updates.  The file was made with an empty function definition, just the edit dates.

pamoroso:
This bit me too.

masinter:
3/3/25 @pamoroso reports @rmkaplan work in progress -- converted to draft.

rmkaplan:
This isn't particularly a draft.  It is "work in progress" only in the sense that this is currently where I am pushing odds-and-ends little fixes to things that I notice as I work on other problems.  These changes are mostly independent but may overlap in the same file, and it isn't worth trying to track the dependencies.   As long as these are not merged, I probably will continue to commit new things that crop up--and when it is merged there will probably be another catch-all PR that will accumulate in the same way.

pamoroso:
I updated to e41009b, nothing unusual.

masinter (approved):
as per discusssion 2/5/25

masinter:
please update issue title & description to mention any behavioral changes, before merging

rmkaplan:
Apart from the change-description in the first comment, this also extends the TEDIT.SELPROP function to make the SET property accessible and to make it possible to change the SHADEHEIGHT and SHADE properties

Issue #2050, opened 2025-02-27
scripts/loadup-full.sh fails if not connected to the clone directory


rmkaplan:
On my mac running sequoia, but I am connected to my Medley working directory, not the clone master.  I have copied all of the new scripts from the clone master.  It seems that it really wants to be in a clone.

Here is the shell transcript until it breaks:

kaplan@Ron-Macbook-Air git-medley % cdwm
kaplan@Ron-Macbook-Air working-medley % scripts/loadup-full.sh
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
>>>>> START loadup-init
"/Users/kaplan/Local/medley3.5/working-medley/../maiko/darwin.aarch64/lde" "/Users/kaplan/Local/medley3.5/working-medley/internal/loadups/starter.sysout" -id "loadup_1" -title "Medley Interlisp :: loadup_1" -g 1024x768 -sc 1024x768 -noscroll
MEDLEYDIR: "/Users/kaplan/Local/medley3.5/working-medley"
LOGINDIR: "/tmp/loadups-78991/logindir"
GREET FILE: "/tmp/loadups-78991/loadup-init.init"
VMEM FILE: "/tmp/loadups-78991/logindir/vmem/lisp_loadup_1.virtualmem"
+++++ SUCCESS +++++
..... files created .....
-rw-r--r--  1 kaplan  wheel  4316160 Feb 27 10:21 /tmp/loadups-78991/init.dlinit
-rw-r--r--  1 kaplan  wheel  119329 Feb 27 10:21 /tmp/loadups-78991/init.dribble
-rw-r--r--  1 kaplan  wheel  4169728 Feb 27 10:21 /tmp/loadups-78991/init.sysout
-rw-r--r--  1 kaplan  wheel  55805 Feb 27 10:21 /tmp/loadups-78991/RDSYS
-rw-r--r--  1 kaplan  wheel  42969 Feb 27 10:21 /tmp/loadups-78991/RDSYS.LCOM
-rw-r--r--  1 kaplan  wheel  89162 Feb 27 10:21 /tmp/loadups-78991/I-NEW
-rw-r--r--  1 kaplan  wheel  70633 Feb 27 10:21 /tmp/loadups-78991/I-NEW.LCOM
<<<<< END loadup-init

fatal: not a git repository (or any of the parent directories): .git
>>>>> START loadup-mid-from-init
[...]

PR #2052, opened 2025-03-01
Fix Issue #2050 - loadup script failures if Medley is not a .git directory

fghalasz:
Fix Issue #2050 - loadup script failures if Medley is not a .git directory.  Change is in loadup-setup.sh, where it now checks if the .git subdir exists before trying to compute the COMMIT_ID.

### DIFF
--- a/scripts/loadup-setup.sh
+++ b/scripts/loadup-setup.sh
@@ -50,14 +50,27 @@ then
   fi
 fi
 
-HAS_GIT= [ -f $(which git) ] && [ -x $(which git) ]
+HAS_GIT= [ -f $(command -v git) ] && [ -x $(command -v git) ]
 export HAS_GIT
 
-git_commit_ID () {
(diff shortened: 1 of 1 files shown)


nbriggs:
This is what I use in maiko --
```
#!/bin/sh
if command -v "git" >/dev/null 2>&1; then
    MAIKO_REV="$(git status --porcelain)"
    if [ $? == 0 ]; then
        if [ ! -z "$(git status --porcelain)" ]; then
            MAIKO_REV="$(git rev-parse --short HEAD)-dirty"
        else
            MAIKO_REV="$(git rev-parse --short HEAD)"
        fi
    else
        MAIKO_REV="none"
    fi
else
    MAIKO_REV = "none"
fi
```
(where "none" means you're not in a git controlled directory)

rmkaplan (approved):
I didn't look at the code, but I tried it out.  My loadup now goes through.

PR #2056, opened 2025-03-06
move files from sources that are not used

masinter:
the files moved are not loaded in any loadup or EVAL@COMPILE or otherwise referenced.


nbriggs:
Uh, "not used" in a loadup doesn't mean nobody ever needs to load them.   Nobody will ever need to load SEDIT-DEBUG if they're working on SEDIT ??

masinter:
SEDIT-DEBUG is interesting -- a set of utilities, e.g., changing FNS to FUNCTIONS.
There are no mentions of SEDIT-DEBUG anywhere (`grep -lri sedit-debug `) in any documentation or other files.  

I could see moving the file to the "internal" directory instead of "obsolete/sources/" but it is unlikely to be useful.

rmkaplan:
I put my TEDIT-DEBUG in internal/, partly because it is mostly undocumented.  But perhaps it would be more likely to be seen and used if it was in the TEDIT/ subdirectory.

I hate to mention it, but maybe SEDIT should have its own subdirectory under sources/ ?

masinter:
should triage and distinguish between 'unused' and 'obsolete'.

masinter:
the 'obsolete' directory has been removed from master (copied to Interlisp/history repo).
So any files that here are "move to obsolete" here should be turned into an issue "delete if unused" and a bigger net to catch files that are actually used or mentioned anywhere.

In the meanwhile, closing this PR.

Issue #2055, opened 2025-03-06
Tedit: line positioning, find and substitute "didn't", and break while saving


MattHeffron:
**Describe the bug**
Arrow key positioning seems to be off by one sometimes. 
Find and Substitute commands didn't search the file.
Also, trying to save a file gave a BREAK.

**To Reproduce**
1. On GitHub I had merged master (b125764) into branch mth35 (PR 2015)
2. I switched to that branch locally, and did a git pull.
3. I emptied my `loadups` directory and built the `loadups` locally
4. I copied the `lispusers>000-docs>LispUsers-Template.TEdit` to `lispusers>READ-BDF.TEDIT`
5. I started medley with the just built FULL sysout.
6. I opened Tedit on the `READ-BDF.TEDIT` file, and opened the attached "Expanded Menu"
7. **With the mouse**, I positioned the type-in cursor at the end of the heading line (after the `>>MODULE NAME<<`)
8. I typed `READ-BDF` but it was inserted **between** the 2 `<` characters.
9. In the Expanded Menu, I typed `>>MODULE NAME<<` in the Find field. Clicking on Find did nothing.
10. Similarly, the Substitute in the Expanded Menu did nothing.
11. Using the Find directly from mouse clicking on the Tedit title bar, found the text, but it seemed to be off by 1 position at each end of the highlighting 

![Image](https://github.com/user-attachments/assets/4a666b08-69f9-42d6-913f-dd558afa615a)
Note the underline starts 1 character right of correct, and the cursor position is **not** at the end of the found text, but is one position to the right of the end of the line.

12. I searched for `>>DATE<<`, and the highlighting was incorrect to the left! 

![Image](https://github.com/user-attachments/assets/8010939f-ba59-4828-8e18-cdf7dd001d4f)

13. I used the Put from the Title bar popup menu, pressed enter to accept the file name and it entered a break. 

![Image](https://github.com/user-attachments/assets/685039c3-9e12-4634-849f-257bb87f10ca)
![Image](https://github.com/user-attachments/assets/ebf8cc73-62e6-411f-835c-4d9c9ee585a8)

^ from the BREAK resulted in no file written, but the Tedit Window still showed version 2.

**Context (please complete the following information):**
 - Are you using online.interlisp.org? [yes / no]
 - OS: WSL on Win 11
 - OS Version: Ubuntu 22.04
 - Display/window system: X11
 - Host arch: x86_64

MattHeffron:
[...]

PR #2058, opened 2025-03-07
rmk72:  Address Tedit Find and Substitute issues reported in #2055

rmkaplan:
Selection highlighting should be correct if line contains a preceding diacritic.

Find from expanded menu should work.

Screen and selection should not be garbled after substitute.

I have not yet replicated the break-on-put also reported in #2055


MattHeffron (approved):
Approved per 03/10/2025 meeting.

PR #2060, opened 2025-03-08
LLREAD:  Add character names One, Two... for digits 0, 1 ...

rmkaplan:
Currently, (CHARCODE 1) produces 49, just like (CHARCODE a) produces 97.  That's because CHARCODE.DECODE interprets a single character (that is not also defined as a cHARACTERNAME) input as exactly that character, and gets its character code.

But a character name in the charset,charnum format is treated in a different way.  If either part of the name is a number, it is interpreted as an octal number, to make it easier to match against the XCCS specifications.  So (CHARCODE Meta,1) does not map to the position of digit 1 in the Meta (currently 1) character set (= 256+49=305), it maps to 256+1=257.  In order to get 305, you have to specify Meta,61 (where 61 is the octal equivalent of 49).  

That's rather opaque, and I found that TEDITKEY had a workaround based on the obscure fact that CHARCODE.DECODE adds 128 for each # it sees in the beginning of a character name.  So ##1 adds 2*128 to 49 to get 305, corresponding to digit 1 in the Meta character set.  But that depends on the fact that Meta is currently defined as charset 1, it won't work when we move Meta up to the Unicode/XCCS undefined region of the code space.

This PR adds (Zero 48) (One 49)...(Nine 57) to the initial value of CHARACTERNAMES in LLREAD, so that 305 for example can be specified as Meta,One.  This is more obvious than Meta,61 and will remain valid when Meta is moved, unlike the ## hack..


pamoroso:
I confirm that with this change `(CHARCODE Meta,One)` returns `305`.

nbriggs:
It's probably a good idea to look at http://www.bitsavers.org/pdf/xerox/interlisp-d/198510_Koto/198512_3101312_Koto_Release_Notes.pdf on character codes (and keyboards etc.) and consider the other places that naming "One" etc. will have an effect, and whether the straightforward interpretation is what you expect/want.  For example, is the value of `(CHARCODE Meta,^One)` expected?  `(CHARCODE ^Infinity)` is 8455, `(CHARCODE Meta,^Infinity)` is a "bad character specification" (though `(CHARCODE META,^LF)` is fine).

rmkaplan:
The documentation isn't exactly right, in at least one respect.  It says that # causes the 8th bit to be set, but in fact, # causes 128 to be added to whatever is there.  Which is why the hack of ##x getting to x in the (current) Meta character set.

But there are also corner cases where my expectations are at best flaky.  One of the examples is (CHARCODE ^Greek,A), which turns out to be equivalent to (CHARCODE Greek,^A).  I think my expectation would be an error with the ^ in front (What is the control-Greek character set?).

Similarly I am not surprised that you get an error with charset,charnum when charnum is the octal-code or equivalent name of a character that is outside of the [0,377Q] range.  What else should you get?

In any event, I think that adding names to clarify that you intend to map to the digit slots in various character sets is orthogonal to any other odd behavior.

MattHeffron:
> But there are also corner cases where my expectations are at best flaky. One of the examples is (CHARCODE ^Greek,A), which turns out to be equivalent to (CHARCODE Greek,^A). I think my expectation would be an error with the ^ in front (What is the control-Greek character set?).

This is because the CHARCODE.DECODE handles the # and ^ prefixes by calling itself recursively.

MattHeffron (approved):
Looks fine.

PR #2066, opened 2025-03-14
Delete DLIONFNKEYS

rmkaplan:
DLIONFNKEYS created button-images on the screen corresponding to some of the Dandelion edit buttons for Tedit.  EDITKEYS is a newer and better version of the same functionality


masinter (approved):
LGYM

PR #2069, opened 2025-03-17
WHEELSCROLL:  Put more of the branching logic inside the WHEELSCROLL function, add character names

rmkaplan:
Changes the function WHEELSCROLL so that the caller can pass a flag to specify the scrolling direction (positive or negative), and WHEELSCROLL decodes the parameters internally (also testing whether or not wheelscroll is enabled).

This is to allow an application(like Tedit) that runs with its own interrupts a simpler way of specifying the intended behavior.

I also introduced cHARACTERNAMES for the characters that wheelscroll uses to signal what should be done.  This is to reduce dependency on the absolute values, so there is one less thing to worry about when we eventually move to Unicode.

I notice also that WHEELSCROLL is not incuded in the full.sysout.  Is it time to include it by default?


MattHeffron:
> I notice also that WHEELSCROLL is not incuded in the full.sysout. Is it time to include it by default?

It appears the it **is** already in `full.sysout`. I started `full.sysout` with no greet file, opened an Interlisp Exec, and entering `PL WHEELSCROLL` displayed the `FILEDATES` property, and the scrolling works fine.

rmkaplan:
Sorry, yes, it's there.  I was looking in the XCL exec.

masinter (approved):
as per discussion in impl meeting

PR #2070, opened 2025-03-17
TEDIT:  New architecture for key bindings, plus better suggestions for initial window regions

rmkaplan:
As usual, lots of files.  Mostly targeted at getting better, more declarative control over keybindings for different Tedit operations (plus random odd corrections along the way).  I will document this in the release notes that will be added in the next day or so.  (This replaces the functionality of EDITKEYS, TKDORADO, and TEDITKEY, which should go obsolete on the next round.)

A relatively minor user-interface change that got swept up in this is a new heuristic for the initial Tedit window region.  When Tedit "gets" an existing file, it used to ask for the window region before it looked at the file.  With this change it gets the file before it opens the window.  It tries to use the file's formatting information, specifically the margins, to calculate the proper width for that file (text area plus line and splitting regions), and it calculates the height that it would take to display the first 20 lines.  The user is asked just to locate the ghost region of that particular size, no sweeping out.

For TDRIBBLE, it uses the width and height of the current Exec.  It uses a different heuristic when viewing Lisp source files, etc.

This may require some fine tuning, but I hope it makes for a smoother experience.


pamoroso:
Is there in any of the sources of this change some table or list of the new TEdit keybindings? Are they the same as TEDITKEY?

pamoroso:
Since I'm testing this PR, for the time being I just need a list of keybindings to exercise and `TEDIT.KEYBINDINGS` in TEDIT-FNKEYS will do. However, in `TEDIT.KEYBINDINGS` there are some keybindings I'm not sure what keys they involve such as `(ITALIC.ON "Function,103")`, so I can't test them yet.

If I understand correctly, `Meta-Ctrl-B` is supposed to turn on bold but instead  inserts a newline and a black box.

When launching TEdit with no document argument, such as from the background menu, the initial window region is wide and thin with a limited editing area:

tedit-initial-region

This is not a problem though as I can shape it to show a larger area. Passing a document argument to TEdit works well with the initial region and the margins. But, when opening or viewing a document rather than writing it from scratch, I usually make the TEdit window taller to fit more text.

pamoroso:
`B` and `N`, as well as `I` and `O`, are adjacent on my Italian keyboard.

I actually don't have any specific preferences for bindings, as long as they work on my keyboard. My problem is I don't know what bindings do what.

rmkaplan:
I can add another little interface function that provides a list of all the bindings in a given readtable.  Maybe in both directions:  for every key with a mapping in the table, what action it maps to; for every action with a mapping in the table, which keys map to it.

The functions there now let you ask about one key or action at a time.

pamoroso:
`TEDIT.GET.ALL.CHARACTIONS`, `TEDIT.GET.ALL.CHARBINDINGS`, and `CHARNAME` are really helpful thanks.

Concerning opening TEdit with an empty document, would it be possible to have a behavior similar to what TEdit did previously, i.e. show an initial outline like the File Browser? Now the mouse pointer changes shape and a message in the prompt window advises to swipe an area. While this is okay with me, having some kind of outline is a clue that may help users less ecperienced with Medley as it's clear what needs to be done next.

pamoroso:
I confirm d4f921a fixes #2075.

pamoroso:
Here are some TEDITKEY actions that would be great to have bindings for in TEdit:

* `\TK.FORWARD.WORD`
* `\TK.BACK.WORD`
* `\TK.GOTOLINEBEGIN`
* `\TK.GOTOLINEEND`
* `\TK.GOTODOCBEGIN`
* `\TK.GOTODOCEND`

nbriggs:
@pamoroso It would be good to look at the list of possible TEdit actions and see if the documentation makes it clear how you should add the key bindings for the actions you want to your personal init file.  If that's not clear then there should be an issue for that against the documentation.  The Interlisp philosophy is very much "we provide the mechanism, you use/configure it to do what you want."

rmkaplan:
> Here are some TEDITKEY actions that would be great to have bindings for in TEdit:
> 
> * `\TK.FORWARD.WORD`
> * `\TK.BACK.WORD`
> * `\TK.GOTOLINEBEGIN`
> * `\TK.GOTOLINEEND`
> * `\TK.GOTODOCBEGIN`
> * `\TK.GOTODOCEND`

It;s now easy to install those actions...the problem is how to figure out the best assignments for the limited number of characters that make sense, are easy to type, and don't conflict with e.g what TTYIN also does.

MattHeffron:
Microsoft Word has a pretty nice interface to define/redefine keys-to-commands. This could _inspire/inform_ how it could be done for TEdit. Notice, that this shows any current key (character) assignments, and shows if the _new_ key is already assigned to a command. This could build the form to assert the set of assignments, or a form to add to `INITCOMS` (or another preferred file).
image

rmkaplan:
I think the interface functions are now there. You can clear out old assignments, add new assignments in your init (although maybe the TEDIT.INSTALL.CHARBINDINGS should take a CLEAR flag so it can all be wrapped up in one call.

The problem is to decide how to allocate the characters to the actions.  E.g. I'd like Meta-S to mean "Save", as in the operating system.  But then what about Substitute?  Possibly Meta,R for Replace.  But then what about Redo?  

We need a better theory of the default assignments.

masinter:
Some ways of evaluating key bindings (are there others? How would you rank these?)
* 
* Consistency between TEdit, SEdit, TTYIN
* Ability to type in without conflict to online 
   (don't assign keystrokes that are eaten by Chrome or popular browsers)
* Consistency (no overlap) with User Interface Guidelines for Mac
*   ditto for PC Windows 11
* Consistency with emacs / SLIME type-in 
*   ditto for typein of Lisp to Visual Studio Code

MattHeffron:
> E.g. I'd like Meta-S to mean "Save", as in the operating system.

But on Windows, Ctrl-S is typically Save. Ctrl-Z is Undo, Ctrl-Y is redo, Ctrl-H is Replace.
It might be a _reasonable_ idea to have _OS-specific_ initial sets, and select based on the underlying OS, then have personal preferences applied on top of that. Or, at least, have the OS-specific sets available, let users modify them, then let them save the **whole set** and apply it _en masse_ in `INIT`, for example.
Some people may prefer to assign the keyboard _Fn_ keys (if we can figure out how to reliably detect them) as the _old_ D-machine keyboard `BOLD`, etc. keys, and then use those in Tedit...

nbriggs:
@MattHeffron if you made a sysout on one machine and then moved it to another you might (a) want the original settings that were in place when you made the sysout because it's *you* using it, or (b) not want the original settings because it's for someone else.  So I suppose any defaulting mechanism would have to observe the difference between `(MAKESYS ...)` and `(SYSOUT ...)`.  On top of that, for the non-OS-specific mappings, you might want to match the software system specific bindings (e.g.,  accelerators for LFG operations) to use the same modifiers as the OS-specific  set of the system it's running on.  Pretty messy.

rmkaplan:
I pushed a little update that extends the way that abbreviation keys are recognized for expansion when ^X is typed.  It tries not just the character next to the caret and the characters in the selection, but also tries extending the selection out to its word boundaries (and it optionally considers a preceding \ as part of the word, so Latex names like \cup can be recognized and expanded).  And, if the substring looks like an MCCS character name (e.g. Greek,Alpha), it tries to decode that too.

pamoroso:
After 0954b84, with this text and caret position in TEdit:

```
\cup^
```

when I press `Ctrl-X` I get this where `#` is a black box:

```
\cu#
```

Selecting `\cup` and pressing `Ctrl-X` yields the warning `No abbreviation to expand`.

pamoroso:
Never mind, when the font is not Gacha the black box renders as a paragraph symbol when pressing `Ctrl-X` next to `p`.

rmkaplan:
> After [0954b84](https://github.com/Interlisp/medley/commit/0954b8494f65f5925ee266666ca024344c41cff8), with this text and caret position in TEdit:
> 
> ```
> \cup^
> ```
> 
> when I press `Ctrl-X` I get this where `#` is a black box:
> 
> ```
> \cu#
> ```
> 
> Selecting `\cup` and pressing `Ctrl-X` yields the warning `No abbreviation to expand`.

Yes, someone would have to create a little file that adds the Latex character name/expansions to TEDIT.ABBREVS.  E.g (\cup . "357,127").  The new code just "does the right thing" with respect to the backslash marker.

rmkaplan:
> * Consistency between TEdit, SEdit, TTYIN

I think a good first step would be to decompose the keyboard mappings in Sedit and TTYIN along the lines of Tedit, so that the character bindings pivot around a core set of common action names, with implementations obviously defined separately.

Sedit would be a good starting point, since the current set up seems to isolate the mappings in the SEDIT-COMMANDS file.  It would be a more abstract interface to the mechanisms and data structure that it current uses to define the mappings and set up the menu.  TTYIN may be harder to get under control.

This also suggests an extension to what I have already done, provide another mapping from action names to short and long form strings that can be used to make a menu.  The Sedit menu strings would be a good place to start, for common actions.  And presumably then it should be common code to create menus that reflect the differently installed charbindings.

rmkaplan:
I fixed it so that smaller/larger go to the next available size, giving a not available message when they go out of range.  Inherited code always bumped by 2 points, so could never go, say, from 14 to 18 if 16 didn't exist.

MattHeffron:
> I think a good first step would be to decompose the keyboard mappings in Sedit and TTYIN along the lines of Tedit, so that the character bindings pivot around a core set of common action names, with implementations obviously defined separately.

Would it make sense to have the keyboard character _operations_ that are common across SEdit, TTYIN, and TEdit (and ...) in a **single _global_ list/table**, with each application able to override the global list at the user's preference?
So, for example, `move 1 character left`, would all be the same by default, but a user _could_ change them. (_**Why would they?**_) 

Are the system _interrupt_ characters "protected", or can the user rebind those as well?
In any case, any functions that map characters to operations (e.g., `TEDIT.INSTALL.CHARBINDINGS`) must check for attempting to reassign to other functions those characters currently bound to the interrupts and block that. 
(If the interrupt character bindings _can_ be reassigned, we should require that the system operations be moved **to** new characters before other operations may be mapped to those characters. I.e., there **must _always_** be bindings to the system interrupt actions.)

nbriggs:
@MattHeffron - say more, what would be the data and control flow for using that?  What would an application that had actions bound to keystrokes that were not common have to do to make use of that, and how would the system dispatch for shared/private keystrokes?

pamoroso:
I wanted to test 955efc9 but I haven't figured out what  `Function` is mapped to in the bindings:

```
("Function,H" SMALLER)
("Function,h" LARGER)
```

rmkaplan:
> I wanted to test [955efc9](https://github.com/Interlisp/medley/commit/955efc98da17f142db27d5b74b26e6a4539cd217) but I haven't figured out what `Function` is mapped to in the bindings:
> ("Function,H" SMALLER)
> ("Function,h" LARGER)

It's a little tricky to test the binding of characters if you can't type them in.  You can click the Buttons menu, for the ones that are represented there.  Otherwise, you have to use BKSYSBUF to insert the character into the system buffer, but that works only if the Tedit window currently has the TTY.

So something like
(DEFINEQ (TESTACTION (LAMBDA (CHAR)
       (DISMISS 5000)
       (BKSYSBUF (CHARACTER (CL:IF (CHARCODEP CHAR) CHAR  (CHARCODE.DECODE CHAR)))))

In the Exec, do (TESTACTION "Function,H"), then you have 5 seconds to click in the Tedit window to give it the TTY.   Probably could set it up so that you pass the tedit stream or window, and switch the tty back and forth.

rmkaplan:
Here's a version that takes the stream/window as its argument.  Don't know why it needs the delay, probably waiting for the TTY.

(DEFINEQ (TESTACTION (LAMBDA (CHAR TSTREAM)
       (RESETFORM (TTY.PROCESS (WINDOWPROP (OR (WINDOWP TSTREAM) (WFROMDS TSTREAM)) 'PROCESS))
              (DISMISS 250)
              (BKSYSBUF (CHARACTER (CL:IF (CHARCODEP CHAR) CHAR (CHARCODE.DECODE CHAR))))))))

pamoroso:
I have tested both versions of `TESTACTION` and the smaller/larger change works.

rmkaplan:
I decided to add a variant of TESTACTION above to the TEDIT-DEBUG file.  So if you load TEDIT-DEBUG, you can try
(TESTACTION "Function,h" [tstream]) where it uses GTS to resolve the stream reference--NIL is the one currently in focus, or the one under the mouse, etc.

Also, you can give an actionname instead of a character:  (TESTACTION 'LARGER).

In doing this, I split out as a subfunction the piece of the TTY command loop that interprets the function calls, so that it can be called directly without going through the keyboard.  That cleaned up the button whenselected function as well--no more BKSYSBUF.

This commit also has some other little fixes.

rmkaplan:
Now also, if you put + or - in the size field of the charlooks menu, then for each selected characer it should go to the next larger/smaller font for that character.  E.g. Select All and raise all the fonts in the whole document (presuming that fonts exist).

(Note:  I'm not sure that FONTSAVAILABLE is returning all the existing directory sizes)

pamoroso:
I tested up to 272bfdd on Linux Mint 22 Cinnamon. `(TESTACTION 'LARGER)` and `(TESTACTION 'LARGER)` work fine.

But, after loading TEDIT-DEBUG, `(TESTACTION "Function,H" (WHICHW))` or `(TESTACTION "Function,H" (WHICHW))` with the pointer over a TEdit window yields the error `Function,H is not a keybinding`. The same with `"Function,h"`.

There's another issue. In a freshly open TEdit window with some text selected and the charlooks menu open, selecting TimesRoman and clicking `APPLY` yields the error `Can't find font TIMESROMAN   (MEDIUM REGULAR REGULAR)--aborted` in the prompt area. The same with other fonts such as Modern or Helvetica.

pamoroso:
`+` and `-` in the size field of the charlooks menu seem to work fine.

rmkaplan:
"Can't find font" error is fixed

pamoroso:
The commits up to 29a2fce fix for me the [can't find font error](https://github.com/Interlisp/medley/pull/2070#issuecomment-2747211916). In addition, opening TEdit with an empty document now supplies a suitable initial region.

rmkaplan:
I cleaned up the CHARNAME function in TEDITFNKEYS.  @MattHeffron , can you compare with your CHARCODE.ENCODE so we can decide which should be moved to CMLCHARACTER?

MattHeffron:
> I cleaned up the CHARNAME function in TEDITFNKEYS. @MattHeffron , can you compare with your CHARCODE.ENCODE so we can decide which should be moved to CMLCHARACTER?

I pushed a new commit of #2078 to:
Handle recursion in the `CHARACTERNAMES` alist.
Allow `(CL:CHARACTERP x)` as well as `SMALLP` arguments.
Cleanup handling of char=255 in _any_ character set.

rmkaplan:
There is also a recursion through lists.

WRT 255, why is that special in the code?  My CHARNAME returns \NORUNCODE for that, because that's what's in the CHARACTERNAMES.  If we always want that to return octal "377" instead (to be less committed to XCCS), I think we can get that by adding ("337" 255) to CHARACTERNAMES.

MattHeffron:
> There is also a recursion through lists. 

Is the "recursion through lists" the 
```common-lisp
((LISTP C)
 (CONS (CHARCODE.DECODE (CAR C) NOERROR)
       (CHARCODE.DECODE (CDR C) NOERROR)))
``` 
 I don't see that in this _reverse_ operation as relevant. It's easy enough to add, but _why_?

> WRT 255, why is that special in the code? 

Because it is not allowed as a character in XCCS, and that's what we currently have. Without my current special handling, or putting a special entry for it in `CHARACTERNAMES` my `CHARCODE.DECODE` returned `"#Rubout"`.

> My CHARNAME returns \NORUNCODE for that, because that's what's in the CHARACTERNAMES. If we always want that to return octal "377" instead (to be less committed to XCCS), I think we can get that by adding ("337" 255) to CHARACTERNAMES.

In the latest commit of rmk76, `\NORUNCODE` is **not** in `CHARACTERNAMES`.

Question: Right now `CHARCODE.DECODE` returns `NIL` for anything that's not `SMALLP` or `CL:CHARACTERP`. Would it be better just to return whatever the _invalid_ character-code is?

MattHeffron:
> In the latest commit of rmk76, `\NORUNCODE` is **not** in `CHARACTERNAMES`.

It _would_ be added when XCCS.LCOM is loaded, but it is under a `(DECLARE%: EVAL@COMPILE DONTCOPY ...` so it isn't.

pamoroso:
I tested up to 805ff73 on Linux Mint 22 Cinnamon. `CHARNAME` and everything else seems to be working fine and I haven't found any obvious issues.

pamoroso:
I tested up to commit 0aaf143 on Linux Mint 22 Cinnamon and didn't find unexpected behavior.

MattHeffron (approved):
Discussed at 3/31 meeting

PR #2071, opened 2025-03-17
Replace Envos with Medley in the lispusers documentation template

rmkaplan:
Probably the only one-file PR I have ever pushed


MattHeffron (approved):
Looks fine in Tedit.

PR #2073, opened 2025-03-17
Rmk78  IMINDEXto remove obsolete Tedit field dependency

rmkaplan:
Removes dependency on \WINDOW field.  Assigned to Frank, for when/if he gets back to working on the IRM.


fghalasz (approved):
LGTM.

Issue #2072, opened 2025-03-17
HCFILES goes into a break during Github Actions (doHCFILES.yml)


fghalasz:
Possible cause:  scripts/do_hcfiles.sh needs to be updated to new cm vs init file setup.

PR #2074, opened 2025-03-17
Fix issue #2072 (doHCFILES.yml fails).  Also fix unrelated bug in buildReleaseInclDocker.yml workflow that was causing build failure.

fghalasz:
This PR fixes Issue #2072   By updating do_hcfiles.sh to use new split between init file and cm files when calling the medley script.

This PR also fixes an unrelated problem in buildReleaseInclDocker.yml workflow by updating gh command that refers to the interlisp/online repo to use "main" instead of "master" as reference branch.  Online repo recently renamed "master" to "main"


masinter (approved):
lgtm

PR #2082, opened 2025-03-26
Remove charset-shifting characternames from XCCS, leaving constants

rmkaplan:
The character names were a mistake


MattHeffron (approved):
LGTM

PR #2088, opened 2025-03-31
TEDITKEY TKDORADO and EDITKEYS are obsoleted by new Tedit keybinding architecture

rmkaplan:


rmkaplan:
I added another commit to also move lispusers/TEDITDORADOKEYS to obsolete/lispusers.

The bindings here that don't conflict with newer uses of those characters are installed in the next Tedit.  (The conflicts are the new uses of Meta,C and Meta,X (clipboard) and others that now map to LINE.UP and LINE.DOWN)

But the ones that have been carried over are useful:  Meta,B/b for bold, Meta,I/i for italic, and a new kind of action for wrapping the current selection.  So Meta,( will wrap in parens, Meta," will wrap in parens.

But no point in not getting this file out of the way now, with these other obsoletes.

PR #2089, opened 2025-03-31
Initial TEDIT window based on Lisp-source width estimate TEDIT.SOURCE.LINELENGTH

rmkaplan:
A reasonable initial width: not too large, but large enough to suppress most unwanted line breaks


pamoroso:
On Linux Mint 22 Cinnamon I tested this change with both TEdit documents and Lisp files and it works well, especially for the initial width. However, with Lisp files the initial height is shorter (about 20 lines) than I'd want resulting in a wide and short window like this for INSPHEX:

tedit-insphex

rmkaplan:
Maybe there should be a difference in the heuristic for number of initial lines to show, depending on whether you are looking at a source file, a lisp function (via the tf command), or an actual Tedit document.

I agree that a source file probably wants a taller window, since there is always junk at the top before you see any of the real content.  Maybe 25 or 30 times the height of the default font, instead of 20?

masinter:
needs release note update ... only covers TEDIT-pF-SEE, COMPARE etc

masinter (approved):
discussed 4/7/25 meeting

PR #2090, opened 2025-04-01
Sets EOL convention ANY in case SYSTEM-EXTERNALFORMAT defaults to :THROUGH

rmkaplan:
This is a follow on to the issue I brought up in the technical meeting 3/31/2025.

The problem is that information returned by a GIT-COMMAND is read from a file with the value returned by SYSTEM-EXTERNALFORMAT.  For a separate unknown reason, sometimes on my Mac UNIX-GETENV returns NIL for all of the possible ways that it tries to detect the system's default encoding, and in that case it returns the format :THROUGH, which just does byte-for-byte with no interpretation.

And if the command result-string has entries separated by LF, those don't get broken up when they are read by GIT-RESULT-TO-LINES.  For the GIT-BRANCHES command, you get one big string with all the branch names that it knows about, separated by LF, and that one big string can't be converted to an atom.  So, for example, it can't even decide that origin/master is one of the branches--it fails.

We talked about changing the default system format to something else (:THROUGH-ANY or :ISO8859/1).  In this update I took the easy way out:  I explicit assert that the EOL convention is ANY, no matter the selected format.

At least this error goes away.  But presumably the return really is UTF-8, so if it had non-ASCII characters it would probably not do the right thing.  We should figure out why e.g. (UNIX-GETENV "LANG") is sometimes defined and sometimes not (on Mac, maybe others).  If we can't guarantee  it automatically, then perhaps there should be a Lisp variable DEFAULT-SYSTEM-EXTERNALFORMAT that the function uses instead of :THROUGH when it gets no other information.  At least the user has easier control.

But what I did here is a first improvement.


pamoroso:
I tested this change on Linux Mint 22 Cinnamon and didn't notice anything unusual with GITFNS.

masinter (approved):
discussed 4/7/25 meeting

PR #2091, opened 2025-04-01
Rmk85  Tedit  distinguish public TEDIT.NTHCHARCODE from private \TEDIT.NTH..., fix screen update bug

rmkaplan:
Pretty much just an internal cleanup to make the distinction between public and private versions of the NTHCHAR functions.  The public versions do more argument coercion, the private versions do the work, and they are gradually being used more often in internal code.

This also fixes a bug in the screen update when the caret is positioned to the left of the first character of the top line of a window, and that character is deleted.  That presumably deletes the space at the end of the unseen line above, but the last unseen word was not being moved down to the beginning of the top-line.


pamoroso:
I tested this change on Linux Mint 22 Cinnamon and couldn't reproduce the bug.

pamoroso:
I can't reproduce the bug on master either, so I'm likely testing it wrong. Here's what I do:

1. open TEdit
2. type `a`
3. press the left arrow key to move the caret at the beginning of the line with `a`
4. press `Ctrl` and middle-select `a` to delete it

At this point I don't notice any rendering issues on main.

rmkaplan:
Here is the bad situation:

The top line in the window is a line in the middle of a paragraph (say it begins with a word "thus") that comes after a preceding line in the paragraph that was broken at a space (which would be the last character on the line if that preceding line were to be displayed).  Suppose the last word on that preceding line was the word "and".

Position the caret at the beginning of the first visible line. Type backspace to delete the space.

You should see the top line beginning with "andthus".  The bug is that you still see only the original "thus"--the top line doesn't get updated.

pamoroso:
Here is a quick test I did on master. I typed a short paragraph in TEdit:

tedit-bug1

Next, after scrolling the window to have a top line in the window and a hidden preceeding line, I placed the caret at the beginning of the top line:

tedit-bug2

Finally, I typed `Backspace`:

tedit-bug3

Does this exhibit the issue?

masinter (approved):
discussed 4/7/25 meeting

PR #2092, opened 2025-04-02
Remove library/keyboardeditor.tedit, it now resides in library/virtualkeyboards/

rmkaplan:
This should have been deleted when the files relevant to virtualkeyboards were grouped together and moved down a level.


MattHeffron (approved):
LGTM.

Issue #2093, opened 2025-04-05
Ensure consistency between the Meta character set and keyboard interpretation of the Meta shift key


rmkaplan:
This was discussed in the technical meeting on 3/31/2025.

There has apparently been some confusion about how a typed character code is modified when the meta-shift is down.  At one point the effect of the meta-shift was to turn on the 8th (128) bit of the code, perhaps following along what "meta" meant in other systems (MIT-derived Lisp machines?).  But in the Medley that we inherited, the function \DECODETRANSITION turns on the 9th bit (256).  This is consistent with the fact that CHARACTERSETNAMES maps Meta to character set 1, so that (CHARCODE Meta,a) returns the code that the keyboard would deliver for the meta-a combination.

Eventually we will want to move the Meta character set to a region of the code space that is undefined for both XCCS and Unicode, as a precursor to moving to Unicode internally.  This would separate meta codes from valid Unicode characters in character set 1.

A first step is to fix \DECODETRANSITION in LLKEY to ogor the base ascii code with (CHARCODE Meta,0) instead of with the constant 256.

PR #2094, opened 2025-04-05
Use (CHARCODE Meta,0) instead of \KEYBOARD.META in LLKEY

rmkaplan:
Eliminate the constant \METABIT (=128) and replace the variable  \KEYBOARD.META with (CONSTANT Meta,0).

\METABIT is not used anywhere, \KEYBOARD.META only in \DECODETRANSITIONS.

See comments in #2093


sources/LLKEY
```diff
@@ -964,8 +965,9 @@
               (COND
                  ((AND (OR (fetch (KEYBOARDEVENT META) of KEYSTATE)
                            (fetch (KEYBOARDEVENT FONT) of KEYSTATE))
-                       (ILESSP ASCIICODE \KEYBOARD.META))
-                  (SETQ ASCIICODE (LOGOR ASCIICODE \KEYBOARD.META]
+                       (ILEQ ASCIICODE 255))
+                  (SETQ ASCIICODE (LOGOR (CHARCODE Meta,0)
```
MattHeffron, line 969: I was going to suggest wrapping the `(CHARCODE Meta,0)` in `(CONSTANT ...)`, but checked and `CHARCODE` has an optimizer, so unnecessary.

MattHeffron (approved):
LGTM. (untested)

PR #2096, opened 2025-04-05
Masterscope OUTPUT TEDIT uses atom word-bound table

rmkaplan:
So when you middle-click on an string in the output, it extends to the sepr and break characters.  Previously, middle-clicking on an X in "XXX.YYY"  would select XXX, with this change it selects XXX.YYY.  So, for example, copy-select and meta,O tend to do the right thing.


pamoroso:
I tested this change on Linux Mint 22 Cinnamon up to 6bedae3. `OUTPUT TEDIT` works as expected and I haven't found anything unusual.

masinter (approved):
discussed 4/7/25 meeting

Issue #2107, opened 2025-04-17
READ-BDF/BDF-TO-FONTDESCRIPTOR fails on FontForge generated BDF of macOS Monaco font


nbriggs:
**Describe the bug**
After reading the BDF for Monaco, calling `BDF::BDF-TO-FONTDESCRIPTOR` fails with a type-mismatch error, "(-1) is not a number" when (-1) was passed to IL:UNICODE-EXTEND-TRANSLATION, called from UNICODE-UNMAPPED, called from IL:UTOXCODE?

**To Reproduce**

Used FontForge to load macOS /System/Library/Fonts/Monaco.ttf and write it out, with default settings, creating amongst other sizes, Monaco-10.bdf

```sh
(SETQ MONACO10 (BDF:READ-BDF "{dsk}<~>Lisp>Monaco-10.bdf"))
(BDF:WRITE-BDF-TO-DISPLAYFONT-FILES MONACO10 "{dsk}<~>Lisp>" NIL NIL 'MONACO 10 'MRR 0 'DISPLAY)
```

**Expected behavior**
I expected it to produce a Monaco 10 display font file.

nbriggs:
[Monaco-10.bdf.gz](https://github.com/user-attachments/files/19786212/Monaco-10.bdf.gz)

PR #2108, opened 2025-04-17
Better handling of a glyph with ENCODING of -1.

MattHeffron:
I treat it as the _slug_ glyph, instead of the _default_ of a solid block.
This addresses issue: #2107 

None of my test cases had that any glyph with that ENCODING.


pamoroso:
On Linux Mint 22.1 Cinnamon I tested the changes up to edcc0e3 and now `BDF:WRITE-BDF-TO-DISPLAYFONT-FILES` processes the [Commodore 64 font](https://beej.us/c64bdf) with no errors.

But how can I determine the name of the converted font to feed to the TEdit charlooks menu? I added the directory with the new font to `DISPLAYFONTDIRECTORIES` but entering `c64`, `c64d`, `c6416`, `c6416-mrr`, or anything else in the `Other` field of TEdit's charlooks menu yields `Can't find OTHER font c6416-mrr`.

MattHeffron:
> But how can I determine the name of the converted font to feed to the TEdit charlooks menu? I added the directory with the new font to `DISPLAYFONTDIRECTORIES` but entering `c64`, `c64d`, `c6416`, `c6416-mrr`, or anything else in the `Other` field of TEdit's charlooks menu yields `Can't find OTHER font c6416-mrr`.

This appears to be due to the way that font files are named. 
Since there is no punctuation between the `FAMILY` and the `SIZE` in the file name (e.g. `C6416` when `FAMILY` = `C64` and `SIZE` = `16`), calls to `FONTSAVAILABLE` will fail to report the font. (Internally, it does everything in UPPERCASE.)
It finds the file, then takes apart that filename and checks that the pieces match the original `FAMILY`, `SIZE`, etc. original arguments. It takes the first character that is a digit as the beginning of the size. So, for `c6416` it splits it to `FAMILY` = `C` and `SIZE` = `6416`. The `FAMILY` doesn't match the `C64` it was expecting, so it rejects the match.

Interestingly, calling `(FONTCREATE 'C64 16)` works correctly.
And, after creating that, `FONTSAVAILABLE` **does** succeed, as does using "Other" in TEDIT.

So, what should `BDF:WRITE-BDF-TO-DISPLAYFONT-FILES` do if the `FAMILY` argument provided, or internally determined `FAMILY`, contain any digits?

nbriggs:
I don’t think it should do anything different, but the documentation should say “don’t name fonts with names that end in digits”

MattHeffron:
> “don’t name fonts with names that end in digits”

Actually the font `FAMILY` should not **contain** any digits, it's not just at the end.
Documented in latest commit, which includes some other ease of use changes.

pamoroso:
I actually pass to `BDF:READ-BDF` `c64d.bdf` which doesn't end in digits but the function picks the name in the font metadata which does:

```
STARTFONT 2.1
FONT c64
SIZE 16 75 75
FONTBOUNDINGBOX 16 16 0 -4
STARTPROPERTIES 3
FONT_DESCENT 4
FONT_ASCENT 12 
DEFAULT_CHAR 0 
ENDPROPERTIES
CHARS 95
STARTCHAR C0032
ENCODING 32
SWIDTH 666 0
DWIDTH 16 0
BBX 16 16 0 -4
BITMAP
0000
...
```

Anyway I'd say READ-BDF works pretty well:

tedit-c64

MattHeffron:
> I actually pass to `BDF:READ-BDF` `c64d.bdf` which doesn't end in digits but the function picks the name in the font metadata which does

It isn't the name of the file in the call to `BDF:READ-BDF` that matters, it is in `BDF:WRITE-BDF-TO-DISPLAYFONT-FILES` that the `FAMILY` argument provided, or internally determined from the metadata, that shouldn't **contain** digits (not just _end_ with digits).

So, if you had entered 
`(PROG1 T (SETQ C64D (BDF:READ-BDF "c64d.bdf" T)))   ; Using the VERBOSE parameter` 
it would have displayed:
```
Name: c64
Family: c64
Size: 16
Weight: MEDIUM
Slant: REGULAR
Expansion: REGULAR
``` 
_(Wrapping with the_ `(PROG1 T ...)` _prevents the printing of the_ `BDF::BDF-FONT` _structure returned and enables seeing that display.)_

If then you called
`(BDF:WRITE-BDF-TO-DISPLAYFONT-FILES C64D "{DSK}my-font-path>" :FAMILY "Commodore")`
it would have returned five values:
```common-lisp
{COMMODORE16-MRR/743,121356}
(0 33 34 35 225 226 227 236 238 239 253)
NIL
NIL
NIL
``` 
_(The `743,121356` is the address of the `IL:FONTDESCRIPTOR` object, and would be different every time.)_
This would have written files named like: `{DSK}my-font-path>c0>Commodore16-MRR-C0.DISPLAYFONT` and it would have been _findable_ in the TEdit Char Looks menu **Other** with the name "Commodore".

pamoroso:
I updated to b3796eb and regenerated the font as [you explained](https://github.com/Interlisp/medley/pull/2108#issuecomment-2830963893). Now TEdit detects the font with no issues.

pamoroso (approved):
Looks good to me.

PR #2111, opened 2025-04-20
Extensive loadup scripts revamp

fghalasz:
This PR supersedes PR#1958 as well as PR#1948, which have been closed.

This PR is an extensive revamp of the loadup scripts including the following changes:

1.  New omnibus script _scripts/loadup_ which is meant to be the single interface to the loadup system.  The man page for this script can be found here: [https://online.interlisp.org/downloads/man_loadup.html](https://online.interlisp.org/downloads/man_loadup.html)

2.  The new loadup script allows you to restart the loadup process from any particular stage - init, mid, lisp, full, etc. For example, you can start the loadup from an existing init-mid.sysout and have it run thru creating the full.sysout.  The call for this would be: ```./scripts/loadup --target full --start mid```.   (See man page for all options to the loadup script as well as examples of their use.)

     In order to facilitate this target/start feature, the loadup workdir has been moved to a single location per MEDLEYDIR, specifically to MEDLEDIR/loadups/build.   (Previously, it was different for every invocation in /tmp/loadups-$$)

      When restarting the loadup from, say, lisp.sysout, the script will look for the lisp.sysout to start from first in MEDLEYDIR/loadups and then in MEDLEYDIR/loadups/build.  If the starting sysout is found in loadups, it will be copied (non-versioned) into the workdir before doing the loadup, overwriting whatever is already there.

     There is now a lock (MEDLEYDIR/loadups/build/lock) that prevents concurrent loadups from running (and very occassionally needs to be manually removed).

3.   At successful completion of a loadup run, the created .sysouts and other files are moved into the loadups directory as before (unless the --nocopy option is specified).  BUT the build files - including the dribble files - are left in the working directory (i.e., in loadups/build) 

4.  The loadup script now supersedes the former loadup-all.sh, loadup-full.sh and loadup-db.sh scripts.  But I have left in their place scripts that call the new loadup script with the right options.  loadup-all.sh is now just a link to the single loadup script, who's options are (hopefully) a superset of loadup-all.sh.

5.  All of the worker scripts (e.g., loadup-full-from-lisp.sh) have been updated to use better mechanisms to catch and report errors, including the new (LOGOUT T EXITCODE) feature.  The main script has been updated to better catch errors in these worker scripts when they do happen.

6.  You can now specify a MAIKODIR for the loadup, either using the MAIKODIR env variable or thru the ```--maikodir``` command line option.

7.  All of the loadup scripts have been moved down one level into scripts/loadups.  But there are links from the scripts/ directory into the scripts/loadups directory for all of the top-level legacy scripts - loadup-all.sh, loadup-db.sh, loadup-full.sh so that you do not have to change your own scripts unless you need to use some of the new features.  More importantly, none of the github workflows need to change right now.

8.  The SYSOUTCOMMITS mechanism had to be changed since there is no guarantee that e.g., lisp.sysout and full.sysout are built on the same commit.  So (ASSOC 'MEDLEY SYSOUTCOMMITS) now returns an assoc list of sysouts, e.g., ((INIT "aaa")(LISP "bbb")(FULL "ccc")(APPS "ddd")), showing the commits for the various layers of the loadup. 

9.  The new loadup scheme allows different sysouts to be created from different commits, so there is an issue with RDSYS (and RDSYS.LCOM) being out of sync with one or more of the sysout in loadups.  RDSYS(.LCOM) are copied into library when ever a loadup of the Init stage completes successfully (unless the -nocopy option is specified).   The only way to solve this issue when it arises is to do a complete loadup from starter.sysout to full.sysout (or apps.sysout) to ensure evrything is built on the same commit.


pamoroso:
On Linux Mint 22.1 Cinnamon I ran my usual legacy loadup process under this change, i.e `loadup-all.sh -apps`. It completed successfully and Medley runs fine.

The only thing of notice is on my 5 GHz 13th gen Intel Core i7 PC there's a short delay each time Medley is launched between the various steps of the process. It's completely harmless and no more than a fraction of a second, but enough to be perceived.

masinter (approved):
much clearer than before. Did a little testing, no fails so far.

PR #2112, opened 2025-04-21
(WHEREIS xx NIL) means (WHEREIS xx '(FNS FUNCTIONS))

rmkaplan:
That's the convention for other filepkg entries (GETDEF etc.)


pamoroso:
I tested this change on Linux Mint 22.1 Cinnamon and it works as expected with no issues.

masinter:
The convention for some file package functions is to enumerate all defined file-package types, and if there is more than one, to ASKUSER which one they want.

I think this is the case for ED and PF*

rmkaplan:
PF just does FNS/FUNCTIONS explicitly.

ED is a user interface function, so user interaction is reasonable. WHEREIS is also called internally.

MattHeffron (approved):
Per discussion 4/21 meeting. 
Additional functionality mentioned could be additional PR.

PR #2113, opened 2025-04-21
Rmk87  minor updates  EXAMINEDEFS, REGIONMANAGER, and  TEDIT-PF-SEE

rmkaplan:
See commits


masinter (approved):
4 separate PRs would have been better?
Will merge without squashing

PR #2117, opened 2025-04-23
LOADUPFULLFONTS: don't choke on extra index.html files in postscript font directory

masinter:


fghalasz (approved):
Works exactly as expected.  Approved.

PR #2123, opened 2025-04-25
Remove DIRECTORYNAMEP check in EDITCALLERS

rmkaplan:
So that (EDITCALLERS 'xxx 'UNICODE) works.

UNICODE is both a directory and a file.  The DIRECTORYNAMEP test in EDITCALLERS means that nothing happens if called as above.  (It would work 'UNCODE. instead of 'UNICODE).

As near as I can tell, nothing breaks if that test is removed--the file is found and the directory doesn't seem to show up.


masinter:
I don't understand; is it legal to have a file and a directory with the same name?
in Unix? In Meley {DSK}?

rmkaplan:
Not at the same level.  But the target file can be a level down.

masinter (approved):
Nut sure why the DIRECTORYNAMEP was added but it doesn't seem to be needed.

The example for this was UNICODE: EDITCALLERS(FOO UNICODE) at MedleyDIR returns without searching while EDITCALLERS(FOO UNICODE.) searches library/UNICODE.

PR #2126, opened 2025-04-26
lispusers/COMMENTHACKS had a bogus 255

rmkaplan:
I don't know what this does, but it had an invalid MCCS/XCCS 255 byte after the STOP that was screwing up reading it as text.  I deleted that byte.


MattHeffron (approved):
Looks correct. (How did this happen?)

Issue #1848, opened 2024-10-05
medley loadup scripts should not error if an lde/ldeinit is on the path


nbriggs:
I thought we'd fixed this problem a while back --

```
>>>>> START loadup-init
ESC[41mESC[37mERROR: Cannot find the Maiko executable (darwin.x86_64/lde) in eitherESC(BESC[m
ESC[41mESC[37m"/Users/briggs/Projects/medley/maiko" or "/Users/briggs/Projects/medley/../maiko".ESC(BESC[m
ESC[41mESC[37mExiting.ESC(BESC[m
----- FAILURE -----
..... files created .....
<<<<< END loadup-init

----- loadup-all.sh: FAILURE -----
briggs@macrobiotic medley % which lde
/Users/briggs/Projects/maiko-test/darwin.x86_64/lde
briggs@macrobiotic medley % 
```
If lde (and ldeinit) are accessible from the PATH then it should *NOT* be trying to find them in locations constructed from the current working directory.

nbriggs:
Not fixed -- I have an lde, ldex, and ldeinit on my PATH, but no maiko directory, and yet it reports:

```
briggs@macrobiotic medley % ./loadup --lisp
>>>>> START loadup-init
ERROR: Cannot find the Maiko directory at either
"/Users/briggs/Projects/medley/maiko" or "/Users/briggs/Projects/medley/../maiko".
You can use the --maikodir argument or the MAIKODIR env variable
to specify the Maiko directory.
Exiting.
----- FAILURE loadup-init-----
..... files created .....
<<<<< END loadup-init

----- loadup: FAILURE -----

briggs@macrobiotic medley % which lde
/var/tmp/lde
briggs@macrobiotic medley % which ldeinit
/var/tmp/ldeinit
briggs@macrobiotic medley % which ldex
/var/tmp/ldex
briggs@macrobiotic medley % git log
commit a4b9099b80ec743239816514cb3b01658a5fdf75 (HEAD -> master, origin/master, origin/HEAD)
```

pamoroso:
Did commit f53da75 fix this issue?

nbriggs:
This was fixed by Frank's subsequent rewrite of the loadup scripts.

PR #2129, opened 2025-04-28
Fix Issue #1848:  Medley script (and hence loadup script) now checks for lde (ldeinit) on PATH

fghalasz:
This PR modifies the medley script (and hence the loadup script) to check for and use lde (ldeinit) on PATH. 

The following is now how medley script looks for the lde (ldeinit) executable:

1.  If the --maikodir argument is given, use the lde in OSVERSION.MACHINETYPE subdir of the value of the argument; if not found ->error
2.  If the MAIKODIR env variable is set, use the lde in the OSVERSION.MACHINETYPE subdir of the value of $MAIKODIR; if not found -> error.
3.  If neither --maikodir nor MAIKODIR, then look for lde in the following order:
     a.  On the PATH
     b.  In the OSVERSION.MACHINETYPE subdir of MEDLEYDIR/maiko
     c.   In the OSVERSION.MACHINETYPE subdir of MEDLEYDIR/../maiko
     If lde not found in any of these -> error.

### DIFF
--- a/scripts/medley/medley.command
+++ b/scripts/medley/medley.command
@@ -931,7 +931,7 @@ do
         ;;
       --maikodir)
         # for use in loadups
-        if [ "$2" = "-" ] || [ "$2" == "--" ]
+        if [ "$2" = "-" ] || [ "$2" = "--" ]
         then
           maikodir_arg=""
           maikodir_stage=""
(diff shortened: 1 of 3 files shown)


pamoroso:
I tested this PR on Linux Mint 22.1 Cinnamon and my usual apps loadup built successfully. I have nothing unusual to report.

MattHeffron (approved):
Approved per discussion in meeting 4/28.

PR #2131, opened 2025-04-29
GITFNS gwc looks at all subdirectories, like prc

rmkaplan:
The gwc command was only looking at subdirectories that conformed to the hyphen convention (e.g. library>tedit>TEDIT-* for library).  This change makes it compatible with what the prc command was doing.

This was why my working loadup did not include XCL-LOOP, I had missed the update in internal/loadups.


MattHeffron (approved):
Looks correct

PR #2132, opened 2025-04-30
READ-BDF Fix incorrect detection of Italic font slope from BDF font metadata.

MattHeffron:
Discovered that Italic font slope was not correctly detected, so only the _later written_ of `MRR` or `MIR` would survive, as `MRR`.
This is why the example fonts in [discussion #2130](https://github.com/orgs/Interlisp/discussions/2130) were italic when not expected.


pamoroso (approved):
I confirm this PR now handles Italic correctly, here is what Comic Sans looks like in TEdit.

comicsans-tedit

PR #2140, opened 2025-05-07
JSON-GET indexes arrays with integer attributes

rmkaplan:
If an attribute path hits an array and the corresponding attribute is a natural number, it picks out that element of the array


pamoroso:
The new `JSON-GET` works fine and is useful.

nbriggs (approved):
Looks fine.  I approve.

PR #2146, opened 2025-05-10
Clicking on the graph after the main window has closed does't break

rmkaplan:
DINFO has many cross-linking, cyclic-pointer dependencies among a bunch of windows: main text window, graph window, history window, menu window, and submenu window.  

All of these need access to the DINFOGRAPH datastructure so that they all see the same state. Some of them store it as their own WINDOWPROP, others get it from the WINDOWPROP of the main text window through the attachment pointers.

It was previously noticed that clicking in the graph window caused an error (DINFOGRAPH NIL) if the main window had been closed (which takes down all the other windows) and then reopened with a new "man" command.  This is because the close function at least on the text window removed its DINFOGRAPH pointer as a way of breaking the cycles and allowing collection.  But the configuration of links was not completely restored at the next command, and the error ensued.

But cycle-breaking is not particular useful for this application, since the windows are being held onto so that they can come up again on the next command--someone outside the cycle is always holding on to something.  So for this PR I eliminated the error in the obvious way:  I took out the code that was trying to smash the links.

With a little (or maybe a lot) more work to undertand and rationalize all the relationships, it ought to be the case that there is one crucial link that maintains the cycle, say, the link from the text window to the DINFOGRAPH.  Then it would be clear how to break it if collection ever became an issue.  But for now, I think it is better to go for robustness.


pamoroso:
Prior to checking out this PR I can't reproduce the issue it addresses. If I close the main window the graph window closes too and I can't click anything. Opening DInfo again correctly opens both windows.

rmkaplan:
After you do the man to reopen, all the windows do come up.  The bug appears when you then click on a graph node to see another new man page.

pamoroso:
When I do `MAN` to reopen, only the text window comes up. Launching DInfo from the background menu does open both windows but clicking on graph nodes works as expected with no errors.

I can't find any way of keeping only the graph window open.

masinter (approved):
as per 6/16/25 meeting

Issue #2151, opened 2025-05-15
Pseudohost LI needs to be reset when LOGINHOST/DIR in reset in MEDLEY-INIT-VARS


fghalasz:
Currently the pseudohost LI is automatically set to the value of LOGINHOST/DIR whenever the PSEUDOHOSTS file is loaded.   This OK.  But if PSEUDOHOSTS is loaded during loadup (as is the case for the apps.sysout), then when the loadup'd sysout is run, the LI pseudohost is set incorrectly - to the loadup LOGINHOST/DIR instead of to the current LOGINHOST/DIR.

MEDLEY-INIT-VARS is the function we have set up to change the value of LOGINHOST/DIR as needed after logouts, sysouts, savevms, etc.  MEDLEY-INIT-VAR needs to be updated to reset the LI psuedohost (if PSEUDOHOSTS is loaded) whenever it resets LOGINHOST/DIR.

PR #2152, opened 2025-05-15
Fix Issue#2151 MEDLEY-INIT-VARS now resets the LI pseudohost whenever it resets the value of LOGINHOST/DIR

fghalasz:
MEDLEY-INIT-VARS now resets the LI pseudohost whenever it resets the value of LOGINHOST/DIR  - providing PSEUDOHOSTS is loaded and LI pseudohost already exists.  See Issue 2151 for more information.


pamoroso (approved):
On Linux Mint 22.1 Cinnamon this change fixes issue #2151. Approved.

sources/MEDLEYDIR
```diff
@@ -193,8 +190,12 @@
         (IRM.HOST&DIR (MEDLEYDIR '"docs/dinfo"))
         (IRM.DINFOGRAPH)
         (DIRECTORIES (APPEND LISPUSERSDIRECTORIES LISPSOURCEDIRECTORIES))
-        [LOGINHOST/DIR (DIRECTORYNAME (OR (UNIX-GETENV "LOGINDIR")
-                                          (UNIX-GETENV "HOME"]
+        (LOGINHOST/DIR (LET [(LHD (DIRECTORYNAME (OR (UNIX-GETENV "LOGINDIR")
```
MattHeffron, line 193: This doesn't appear to match what the PR description states. The `LI` pseudohost looks like it is reset anytime `LOGINHOST/DIR` is set. I.e., it isn't only when **changed**.
Does it matter?

fghalasz: Technically ypu are correct, should be set/reset.  But practically, whenever MEDLEY-INIT-VARS is executed (after sysout, after makesys, during greet, etc) LOGINHOST/DIR has already been set and it is be reset.  Only exception is somewhere in the loadup process where LOGINHOST/DIR is first set - but I am not sure whether by MEDLEY-INIT-VARS or otherwise.

PR #2154, opened 2025-05-16
Expand the (PSEUDOHOSTS) function so that it can be used cleanly with RESETSAVE

fghalasz:
Expanded PSEUDOHOSTS function to take a single argument NEW.HOSTS.  Function still always returns a list of the current pseudohost pairs.     IF NEW.HOSTS is NIL, that's all it does.  If NEW.HOSTS is a list of (host prefix) pairs, then all current pseudohosts will be removed (by sequentially calling PSEUDOHOST with NIL prefix) and new pseudohosts will be created by calling (PSEUDOHOST host prefix) for all the pairs in NEW.HOSTS.   If NEW.HOSTS is T, then all current pseudohosts will simply be removed.  Any other value of NEW.HOSTS results in an error.

This expanded PSEUDOHOSTS is designed to be used cleanly with RESETSAVE.

This change should not effect any current uses of (PSEUDOHOSTS), unless PSEUSDOHOSTS was called with a spurious argument which heretofore was simply ignored.

The TEdit documentation was updated accordingly.


library/PSEUDOHOSTS
```diff
@@ -151,10 +149,32 @@
...
-    (FOR DEV IN \FILEDEVICES WHEN (TYPE? FDEV (FETCH (PHDEVICE TARGETDEV) OF DEV))
-       COLLECT (LIST (FETCH (FDEV DEVICENAME) OF DEV)
-                     (FETCH (PHDEVICE PREFIX) OF DEV])
+  [LAMBDA (NEW.HOSTS)                                        (* ; "Edited 17-Jan-2022 18:15 by rmk")
+                                                             (* ; "Edited 16-May-2025 9:16 by fgh")
+
+    (* ;; "")
+
+    (* ;; " Returns existing list of PSEUDOHOST pairs.  If NEW.HOSTS is T, all current pseudohosts are removed by calling (PSEUDOHOST HOST NIL) on each current pseudohost in turn.  Otherwise, NEW.HOSTS should be a list of (HOST PREFIX) pairs and all current pseudohosts are r(PSEUDOHOSTSemoved (as above) and the NEW.HOSTS pairs are used to create new pseudohosts by calling (PSEUDOHOST HOST PREFIX) sequentially in reverse order of the NEW.HOSTS list. Reverse order to ensure that (PSEUDOHOSTS (PSEUDOHOSTS))  doesn't impact the ordering in the PSEUDOHOST list.  This function is designed to be used cleanly with RESETSAVE.")
```
MattHeffron, line 157: Typo in comment.

masinter (approved):
LGTM

I didn't look into Matt's "typo in comment" because I wasn't sure which comment with what typo and whether the typo was fixed.

PR #2156, opened 2025-05-16
Fix Issue 2139 (large parts of Medley tree not showing up in files.interlisp.org) and more improvements to files.interlisp.org

fghalasz:
Note:  This PR depends on PR#2154.  Should approve and merge 2154 first.

Here is what this PR does:

1.  Fix Issue 2139:  (MAKE-INDEX-HTMLS) was not handling pseudohosts correctly and an errant LI pseudohost was causing MAKE-INDEX-HTMLS to terminate early.  Adjusted MAKE-INDEX-HTML so it uses psuedohosts only for the top level directory and everything further down in the tree uses the truenames relative to the top-level pseudohost,  Results in a MAKE-INDEX-HTMLS run that works in the presence of random pseudohosts and in a collection of index.html files without difficult to understand and out of context references to pseudohosts.

2.  Remove loadups/build directory from all HCFILES runs (on desktop and via github actions)

3.  Added maiko source code and removed maiko lde executables from HCFILES outputs for github actions - thus adding maiko code and removing maiki executables @ files.interlisp.org.

4.  Fixed scripts/clean_hcfiles.sh so that it actually cleans off all of the index.html files - was missing some.

5. .github/workflow directory was being left out of HCFILES.  Put it back it.

Output from a run of HCFILES using this PR is now live at files.inter;isp.org.


pamoroso (approved):
It seems to be working well, I haven't found any abvious broken links or missing parts.

PR #2157, opened 2025-05-17
Some polish on the new loadup scripts - dribble files and lock overrides

fghalasz:
Added some polish on the new loadups scripts as requested by Ron.  Changes are:

1.  The dribble files are now copied into the loadups directory instead of being left only in the build directory.

2.  If the lock preventing simultaneous loadups is set when loadup starts (e.g., by a failed prior loadup), loadup will now ask the user if they want to override the lock or just exit.  Previously just exited.

3.  There is a new cl arg (--override or -ov) that will cause loadup to just automatically override the lock if needed without asking the user.

Man page updated accordingly.


rmkaplan (approved):
Works well

Issue #2155, opened 2025-05-16
loadup -f -b -x fails -- can't find full.sysout


MattHeffron:
doing a loadup from _scratch_ to create the `full.sysout` and the aux and db supplemental files fails when attempting the db stage.

1. cd to `medley/scripts/loadups`
2. `./loadup -f -b -x`
3. Builds the sysouts through `full.sysout`
4. Generates the aux stage files
5. It fails the `loadup-db-from-full` with error:
```
>>>>> START loadup-db-from-full
Error: cannot find /home/matt/medley/loadups/full.sysout.
Exiting.
``` 
6. No files have been copied/linked into `loadups/`

It seems that the `-b` , or equivalent, options shouldn't be processed until _after_ earlier stage files have been copied/linked, and those generated by `-b` should be handled separately.

PR #2160, opened 2025-05-21
Fix Issue #2155 by reordering  in the loadup script the component loadups and when their outputs are copied into loadups dir.

fghalasz:
Fix Issue #2155 (loadup -f -b -x fails can't find full.sysout).  

Reordered when things are done in the loadup scripts such that:
1.  The sysouts are copied to into the loadups dir before the aux and db subscripts are done
2.   Exports.all is copied into the loadups directory before the -db load is done.
3.   If full.sysout is newer than exports.all in the loadups dir, the loadup-aux is run before loadup-db to refresh exports.all.

This ensures that the depended-on files are in place before any particular component loadup is run.

No changes to the command line arguments.

### DIFF
--- a/scripts/loadups/loadup
+++ b/scripts/loadups/loadup
@@ -18,6 +18,7 @@ main() {
         thinw=false
         thinl=false
         override_lock=false
+        ignore_lock=false
         while [ "$#" -ne 0 ];
       	do
           case "$1" in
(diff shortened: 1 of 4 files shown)


MattHeffron (approved):
Looks good, and ran without error.

PR #2177, opened 2025-06-01
Fix POSTSCRIPTSTREAM to refine or default values for COLOR.

### LINKED ISSUE #2176: POSTSCRIPT doesn't take 0 as a color
There are a couple of Tedit files in lispusers that seem to call for a color change to 0:
CIRCLPRINT.TEDIT
EDITBG.TEDIT
EQUATIONEDITORPRGRMERGUIDE.TEDIT
HPGL.TEDIT

I think they all involve displaying a brush in DRAWCURVE, where the BRUSHCOLOR is 0.

(separately the file EQUATIONEXAMPLES.TEDIT uses font SIGMA 20, which postscript doesn't have)

MattHeffron:
(Don't require them to be FLOATP.)

This resolves #2176.


rmkaplan:
This corrected the zeros, but now I see that the file EQUATIONEDITORPRGMERSGUIDE.TEDIT is also drawing a curve where the brush has color 7. Are integer colors like this some sort of index into a table of colors?

nbriggs:
@rmkaplan there are a few things in can be, but one option is that it's something that is COLORNUMBERP, which is defined on library/LLCOLOR.

MattHeffron:
> Are integer colors like this some sort of index into a table of colors?

`COLORNUMBERP` checks against the color map for the screen, so I don't know what it would do for a B/W screen, nor for an imagestream that isn't to the screen.

This is the `BRUSHCOLOR` of a curve in the second embedded sketch. It appears that color `7` is specified in the sketch, but I don't see how to **set** the color of a `BRUSH` to use in the sketch. Is it actually a `TEXTURE`? So, I don't know how to find out where that "color" came from, and what it **should** do.
It looks like I might have to dig into `SKETCH`. How come `TEDIT` has no problem with that?

rmkaplan:
Tedit doesn't know about colors in image objects, and it doesn't interpret colors in any other way.  DSPCOLOR of a Tedit stream just puts the color-pointer (whatever it is) in the character-looks, and passes that to the back-end image stream through the image-stream's DSPCOLOR.

MattHeffron:
I did a bunch of digging through the source code and discovered:

- When a drawing operation targets the display, the `BRUSHCOLOR` is ignored if the target is 1BPP `ScreenBitMap`.
 If the target is `ColorScreenBitMap` , then `BRUSHCOLOR` is set into a bitmap of the same BPP as `ColorScreenBitMap` with no further interpretation, leaving that to the **use** of the `ColorScreenBitMap`.
- In `INTERPRESS` the function `\SETBRUSH.IP` ignores the `BRUSHCOLOR` if it is a `NUMBERP` that's `<=` 0. 
 (It doesn't do this safely. There are a couple of points there where the `BRUSH` function argument is used _as_ a `BRUSH` record type, but it hasn't verified that it **is** that type.)

So, I **propose** that `POSTSCRIPTSTREAM` should behave _similar to_ `INTERPRESS` and just _ignore_ a `BRUSHCOLOR` if it is a `SMALLP` that's neither 0 nor 1, and use the current `DSPCOLOR` of the stream. If someone wants a specific color, or grayscale, they can use the color descriptions as documented in `POSTSCRIPTSTREAM.TEDIT`.
This seems the best alternative, both for consistency with `INTERPRESS` and simplicity.

@rmkaplan @hjellinek @masinter  Any comments/objections?

rmkaplan (changes requested):
I tried it, and it worked fine on the problematic equation file.  But the paragraph in the documentation beginning "Anywhere that a texture or color..." seems to only allow 0 and 1, should that be updated to describe the other pass-through possibilities ?

PR #2180, opened 2025-06-04
Medley script currently forces -vnc when using WSL1.  This PR removes this infelicity.

fghalasz:
Medley script currently forces -vnc when using WSL1.  This PR removes this infelicity.  The -vnc flag works the same on WSL1 as on other linuxes (including WSL2).


masinter (approved):
as per 6/16/25 meeting
(also needs 2163)

PR #2181, opened 2025-06-04
Add a --vnc flag to loadup script, used to tell loadup scripts to run Medley with the --vnc option.

fghalasz:
Add a --vnc flag to loadup script, used to tell loadup scripts to run Medley with the --vnc option. 

Used primarily on WSL1 and WSL2 systems where loadup defaults to X Windows.  But can be used on any Linux platform that uses Xvnc.  

Not relevant on MacOS and Cygwin. 

Man page updated accordingly.


masinter (dismissed):
as per 6/16/25 meeting

InterlispMachineUser (approved):

PR #2184, opened 2025-06-05
PDFSTREAM opens the scratch stream

rmkaplan:
It was testing OUTFILEP and failing on a closed scratch stream.  This happened at least when doing a hardcopy to a printer from the file-browser


MattHeffron (approved):
LGTM

PR #2168, opened 2025-05-26
Support for MCCS and other font character codes, medley-font format

rmkaplan:
This is the first draft of changes aimed at the font problems discussed in #2120 and  #2135 and elsewhere.

It introduces and integrates a new, medley-specific font-file format that let's us write out and read in font information apart from the legacy font formats (strike, ac, ip...), and introduces support for font character-set manipulation (including clarifying and defining the MCCS encoding and its relationship to Unicode and XCCS.

The code for the medleyfont format is described in the doc/internal/ tedit file.  I'm also writing up description of the changes/extensions to FONT.

The point of this PR/draft is that things should work as they used to, that the integration of new stuff hasn't broken anything even though it may not yet have improved anything.

PR #2192, opened 2025-06-17
Rmk97  medleyfontfile format

rmkaplan:


rmkaplan:
This includes MEDLEYFONTFORMAT and changes in other files to include it in the loadup and to load fonts in the new format.  But no medley-formatted fonts are included.

The test case would be to load all of the charsets of a BDF-derived or HTML font into a font descriptor, and then execute (MEDLEYFONT.WRITE.FONT font).

You may need to explicitly create the fonts/medleydisplayfonts or fonts/medleyhtmlfonts/ directories.

The document is in docs/internal/MEDLEYFONTFORMAT.TEDIT.  (That may not be there--GITFNS said it did not need to be included, that it had been included before, but I'm  not sure.  LMK.)

This is part of what's included in #2168, which will eventually be closed.  The other part is the rearchitecting of FONTCREATE.  However, there is a minor update to FONT included here:  The FONTDESCRIPTOR has new fields to track the charencoding of the font and whether or not the font is complete.  The CHARSETINFO also has a complete flag, plus a slug flag, and a pointer field to an ALIST that typically would include at least the charencoding of the charset.

Fonts that are coded in MCCS should have MCCS at least in FONTCHARENCODING field.

rmkaplan:
A further comment:  on reflection, I believe that this has what is needed to write the new fonts, but additional updates to FONT will be needed to read those files in the normal operation of FONTCREATE.  I may have to backfill a few more changes to FONT to link that in.  But building comes first.

PR #2195, opened 2025-06-23
Recompiled all Tedit files

rmkaplan:
For a baseline comparison with rmk88.  A slight change to TEDIT-FNKEYS (constant dependency) that the recompiled revealed.


pamoroso:
I switched to this branch and so far I'm not noticing anything unusual on Linux Mint 22.1 Cinnamon.

rmkaplan:
Well, unfortunately, this fails the STRESSINSERT stress test, again showing the arrayblock crash.  

I updated the TEDIT-STRESS file here to include the updated tests, but that's no part of the Tedit runtime.

You might try LOAD(TEDIT-STRESS.LCOM) and let (STRESSINSERT) run for awhile.

pamoroso:
I updated to commit 0f17d71 and ran the stress test but, since it didn't break, I interrupted it after about 10 minutes of runtime and over 14K iterations. Approximately how many iterations are needed for the test to fail?

rmkaplan:
I see it at about 10, but it varies.

Are you running in the lispusers directory?

pamoroso:
No, I'm running in the default Medley directory. I'll try running from lispusers.

pamoroso:
I ran another test like this (`MEDLEY` is a pseudohost pointing to the Medley directory):

```
(LOAD '{MEDLEY}library>tedit>TEDIT-STRESS.LCOM)
CD '{MEDLEY}>lispusers>
(STRESSINSERT)
```

Since the test ran regularly for more than 6 minutes and over 220 iterations I stopped it.

rmkaplan:
As was suggesed in the meeting, I modified the stress test so that it saves the virtual memory (either SAVEFM or SYSOUT) before each insertion.  The hope is that we will catch the state of the virtual memory just before it crashes, and perhaps then run to watch more precisely where it goes bad.

This strategy doesn't seem to help.  When the pre-crash memory image is started up, it doesn't immediately crash as it expected.  It just keeps on doing the insertions as if nothing had gone wrong.

Is there any value in looking more carefully at the precrash image?  Or does this suggest that the problem is not in the Lisp memory but in some Maiko state?

rmkaplan:
(P.S.  These are always arrayblock crashes, either bad passwords or header/trailer length mismatch).

nbriggs:
I've added code that looks over all the memory of a block it is about to free to see if there are any words that look like block headers/trailers within it -- I have not found any cases yet in many hours of running the code.  I am running with ARRAYBLOCKCHECKING turned on (both from the Lisp side, and by forcing it on in Maiko directly), but am NOT running with TEdit calling CHECKARRAYS.  If there is a problem, and not something caused by the timing of when CHECKARRAYS is called, then it should become visible from the Maiko checking.

rmkaplan:
Are you suggesting that I stop the calls to CHECKARRAYS and see what happens?  And that we may be out of the woods if the Maiko checks don't fail?

Is that code in the current maiko master?

nbriggs:
It could be that CHECKARRAYS needs to be uninterruptable, but I think stopping calling it is a good test.  Still set ARRAYBLOCKCHECKING to T though, since you won't have it otherwise forced on in the Maiko code.

The extra checking/trap code is NOT in maiko master -- it's hacks to the code that makes it unusable in a release situation (it generates a dozen or so lines of output on stdout for every allocation/reclaim, and if it *does* find a potential problem (there are some false positives) you have to poke around in URaid and then continue.

nbriggs:
Compiling CHECKARRAYS with an UNINTERRUPTABLE around the PARSEARRAYSPACE doesn't make it behave better.
I will add more code in Maiko to see if I can see what the Lisp PARSEARRAYSPACE is doing differently.

rmkaplan:
So, there is no reason for the stress tests to continue calling CHECKARRAYS, just set ARRAYBLOCKCHECKING to T?

And, is there any value in saving the memory image before each event?

I.e., do the tests by setting ARRAYBLOCKCHECKING to T and just let it run?

masinter:
The medley process scheduler isn't preemptive ... no other process will run until the current process does a (BLOCK), or there is an interrupt by timer (for SPY) or keystroke (an interrupt character).

UNINTERRUPTABLE is only useful to let operations that really must be done in a single transaction lest the memory would be corrupted. 

It should never be wrapped around code that in the normal course of events would (BLOCK).

masinter:
at meeting 6/30/25 we decided to  close this as subsumed by rmk88. 

We should have some way of checking or recreating compiled code from Lisp sources so this wouldn't be necessary.

PR #2197, opened 2025-06-27
Fix online.interlisp.org/downloads/medley_downloads.html to reflect new wsl1 versus wsl2 ,releases (instead of just wsl)

fghalasz:
Recently split Medley releases for wsl into two seperate releases one for wsl1 and one for wsl2.  This change was not reflected in the https://online.interlisp.org/downloads/medley_downloads.html page.  This PR updates this page (or more properly the template from which this page is generated) to reflect this change in wsl releases.


masinter (approved):
LGTM (untested)

Issue #2196, opened 2025-06-25
Cleanup of display fonts


rmkaplan:
In the current standard loadup the displayfont files are spread out in a number of directories, as indicated by the value of DISPLAYFONTDIRECTORIES:

fonts>displayfonts>   fonts>altofonts>  fonts>adobe>  fonts>big>  fonts>other>

In the past it appears that some files have been copied from one directory to another (often into displayfonts>), so that there are multiple instances of the same file.  An initial goal is to have only one directory, fonts>displayfonts> hold all of the files that we need at runtime, and to move all of the copies to corresponding subdirectories of a separate archival directory (e.g. archive/fonts/, which we would not distribute.

To take a simple case,
    altofonts/timesroman6.strike
is a duplicate of
    displayfonts/c0/timesroman06-mrr-c0.displayfont
do the first one would be moved to archive/fonts/altofonts/timesroman6.strike

For the more obscure case that
     other/c0/sigma20-mrr-c0.displayfont and altofonts/sigma20.strike
are duplicates of 
     displayfonts/c0/sigma20-mrr-c0.displayfont
the would be moved to
     archive/fonts/other/c0/sigma20-mrr-c0.displayfont
and
     archive/fonts/altofonts/sigma20.strike

In these examples the files are not only duplicates but their names indicate that they represent the same family/size/face.  There is another situation where the different files appear to represent different fonts:
   displayfonts/c0/helvetica04-mrr-c0.displayfont
   displayfonts/c0/helvetica01-mrr-c0.displayfont
   displayfonts/c0/helvetica02-mrr-c0.displayfont
are all the same file, with presumably one of these names represents the actual/nominal size of the font and the others suggesting an attempt to fill in for missing fonts.  We want to pick one of these to keep, and archive the others.  Although our current character-set datatypes don't encode the nominal size (which may or may not be explicit in the file), a reasonable heuristic is to keep the one whose nominal size is closest to the internally represented font height (sum of its ascent and descent).  In this case size 4 would be kept, 1 and 2 would be archived.  But we would also record the intended substitutions, so that we can fake up 1 and 2 from 4 if there is a FONTCREATE request.  We would make entries
    ((HELVETICA 1)(HELVETICA 4)
    ((HELVETICA 2)(HEVETICA 4)
on the list DISPLAYFONTSUBSTITUTIONS.

[...]

Issue #2186, opened 2025-06-10
Should fonts/altofonts be removed ?


rmkaplan:
It has a bunch of strike files with names that don't conform to either the old or new conventions for font files, so they are never recognized/used.

I think many of them have been copied over (or at least correspond) to files in displayfonts/c0/.  If there are any that have any value and have not been moved over, they should be.

Is there a reason to keep the confusion of having this dead set of files?

A first step would be to remove fonts/altofonts/ from the DISPLAYFONTDIRECTORIES in MEDLEY-INITVARS.  A second step would be to move this directory to obsolete/fonts or internal/fonts.

rmkaplan:
FONT continues to support historical naming conventions for fonts.  Can these be obsoleted/removed, so we don't have to maintain this code?

E.g. a mapping for a certain number of fonts to an 8.3 DOS format:  TIMESROMAN 10 BOLD maps to TR10B000 with some particular extension.

rmkaplan:
The duplicate filescan showed a bunch of font files that are byte-for-byte copies of others, just with different names.  This was a way of compensating for the absence of different sizes for different fonts.  As an example, if Classic 6 Bold doesn't exist for a given character set but Classic 8 Bold does, then there might be a copy of the file for 8 with a name that says that it should stand in for 6. 

I propose to remove all of the copies, and drive any coercions that have been implemented this way by putting entries on a list of mappings to be done at runtime.  So, as we remove the redundant files, we would record their intent by putting entries on the list DISPLAYFONTCOERCIONS.  For this example we would add  ((CLASSIC 6 BOLD)(CLASSIC 10 BOLD) (or just ((CLASSIC 6)(CLASSIC 8)) if all faces are to be mapped in the same way).

FONTCREATE would interpret this mapping.  If it can't find a Classic 6 file, it will read the classic 8 file into a new font descriptor and change its fontprop to say that its size is 6 (so that FONTPROP - SIZE would return 6).

MattHeffron:
Just a thought: It might be a good idea to record on the `OTHERDEVICEFONTPROPS` or `FONTEXTRAFIELD2` an indication that this font was coerced from _fontXX_, otherwise when some text is printed at the _requested_ font size, there would be no way to trace _why_ it's different than on the display. Also, TEdit and Sketch, etc. could display an _informational message_ about the coercion and that printed output probably wouldn't layout the same as the display.

rmkaplan:
PR #2203 deals with this issue

PR #2203, opened 2025-07-03
Rmk99  cleanup display fonts

rmkaplan:
All display fontfiles are now in the single fonts/displayfonts/ directory.  The subdirectories adobe, big, other, altofonts have been deleted, and are no longer on DISPLAYFONTDIRECTORIES.  The few additional displayfonts from the envos repo are included.

This is the result of:
* Removing all byte-identical files, leaving a representative of each such equivalence class in displayfonts/

* Removing all files in displayfonts that are renamed byte-identical copies of other files in displayfonts/ (e.g. helvetica1-MRR-c0.displayfont is a fake-renamed copy of helvetica4-MRR-c0.displayfont).  These are to be constructed on the fly by the font substitution mechanism.

* Roughly comparing the bitmaps of non-identical files with the same font family/size/face/charset in the different subdirectories, and making displayfonts/ have the one that looks best (e.g. most glyphs, unless garbage), giving priority to ones that are already in displayfonts/.

* Moving the remaining odd-ball files from the subdirectories to displayfonts/ (e.g. gates, music).

* Copying from Envos the 30 or so files that are not byte-identical copies of files already in displayfonts/.


rmkaplan:
This deals with the discussion in #2196

fonts/displayfonts/c41/CLASSIC06-MIR.XMEDLEYDISPLAYFONT
masinter, line 1: this stands out -- is it intended?

fonts/displayfonts/c0/TESTFONT12-MRR-C0.DISPLAYFONT
masinter, line 1: does anyone use testfont ... is it a remnant?

rmkaplan:
I debated excluding that particular one (and a few others).  But on the theory that some unknown sofware somewhere might have made use of it, I kept it.

rmkaplan:
Same response:  I don't know, but there may be a dependency somewhere.

rmkaplan:
This also relates to issue #2186

rmkaplan:
Larry and I talked about the possibility of moving the fonts to a git submodule, and decided that more work needed to be done to set that up (see #2211 ), and that it isn't worth putting that in the way of finishing up this round of font cleanup.  It also isn't so clear what the advantage would be, even for the size of our git repositories and downloads.

This PR is just a clean up of the displayfonts/ directory.  Duplicates of files have been eliminated as the other font directories have been deleted, a relatively small number of files that were no in displayfonts/ have been copied over.   This PR does not separately copy all of the deleted directories over to obsolete where they would take up new git storage for no good purpose.

To make it easy to return the pre-cleanup state, should that be necessary, the current master should be tagged.

Otherwise, after testing that nothing has been lost, merging in this PR would be the first step of the cleanup.  A later step will gather all of the display charactersets into a single medley-format file and the displayfonts/ here will be deleted.  But I think there would be only a marginal difference in git storage if merging is deferred until after the files here are deleted.

MattHeffron (approved):
per 7/14/25 meeting

PR #2207, opened 2025-07-07
Correctly wait for edit of button to complete before updating

MattHeffron:
Fix editing of button...correctly wait for edit to complete before updating the button itself.
(The fix in PR #1594 wasn't correct; _or_, it no longer worked properly.)
Add (FIND-BUTTON label) function.
Minor updates to documentation.


pamoroso:
I don't know whether it's my setup but, with or without this PR, `(FILESLOAD BUTTONS)` returns `NIL` without loading the module. I need to connect to the LispUsers directory and load BUTTONS from there.

masinter:
re: "(FILESLOAD BUTTONS) returns NIL without loading the module." 
FILESLOAD returns NIL without printing a message when the files area already loaded.
As would happen if you start with the -a apps sysout, which already includes BUTTONS.

MattHeffron:
@pamoroso If `(FILESLOAD BUTTONS)` actually loads the file when connected to the LispUsers directory, but not if connected elsewhere, then it is probably the case that the LispUsers directory path is not on the `DIRECTORIES` variable. That's used as the list of locations that `FILESLOAD` searches for the file(s) specified.
Also, `(FILESLOAD (FROM LISPUSERS) BUTTONS)` should work without changing `DIRECTORIES`.

pamoroso:
The LispUsers directory is actually on `DIRECTORIES`:

```
4_ (PRINTDEF DIRECTORIES)
({DSK}<home>paolo>medley>notecards>library 
       "{DSK}<home>paolo>medley>medley>library>" 
       "{DSK}<home>paolo>medley>medley>lispusers>" 
       "{DSK}<home>paolo>medley>medley>internal>" 
       "{DSK}<home>paolo>medley>medley>greetfiles>" 
       "{DSK}<home>paolo>medley>medley>doctools>" 
       "{DSK}<home>paolo>medley>medley>sources>")
NIL
```

PR #2208, opened 2025-07-08
Use the loadup step script names as the --id string.

MattHeffron:
Shows where it is in the loadup process at a glance.

A trivial tweak, but it was annoying that **all** the steps showed "loadup_1".


fghalasz:
Merged this change into PR#2200. to avoid conflicts with other changes to loadup script.

PR #2209, opened 2025-07-08
Use a seperate repo for fonts

masinter:
Use the .git submodule feature for the fonts directory.

- **move fonts to separate repo**
- **remove obsolete fonts**
- **make fonts a submodule of medley**
- **update to latest fonts**


rmkaplan:
I cloned the fonts repo as a sister of medley.  

Is that the right place?  What do I have to do for this new project to replace what I currently have in medley/fonts ?  Just change e.g. displayfontdirectories to point to this locally?

Should fonts be a new project in GITFNS, like NOTECARDS, LOOPS, MEDLEY...?

PR #2220, opened 2025-07-17
Move all the proofreader and dictionary-related files to lispsuers/proofreader/, with hyphen names

rmkaplan:
These are interdependent but spread out in lispusers/.   All together they  provided the framework and implementation for spell-checking and dictionary lookup, in the early days. 

The dictionary resources (e.g. the server or wordlists) are no longer available, and these made use of internal Tedit fields and functions that no longer exist.  I made a rough pass to fix up the Tedit dependencies so at least these files compile.  And maybe would mostly work (except for one function that appears to be a version TEDIT.SUBSTITUTE with some additional bells and whistles).  

This subdirectory should eventually be moved to obsolete, this just pulls the files together, fixes the code, and does the hyphen filename renaming.


MattHeffron:
I checked out the rmk105 branch and all of the files, _except_ **README.TXT**, contain the `NUL` character at every position!!

rmkaplan:
Hmm, that's not good!

rmkaplan:
I found, moved, and pushed contentful versions of the proofreader files (I hope).

MattHeffron:
Better, but `PROOFREADER-PROOFREADER.TEDIT` is still all `NUL` characters. 
It is exactly the same size (9531 bytes) as `PROOFREADER.TEDIT` (which has good content), 
so it _seems_ that `PROOFREADER-PROOFREADER.TEDIT` shouldn't be there at all?

rmkaplan:
I think that PROOFREADER-PROOFREADER.TEDIT is the proper name, given the hyphen convention, I'll fix that.

I think there is something wrong with COPYFILE that led to the original wipeouts.  I'll investigate.

rmkaplan:
I believe these renamed files will load, but and some functionality might even work.  But more effort would have to go in to update to modern external formats and Tedit.  For example, it seems to read characters with BIN instead of either READCCODE or the more primitive \INCCODE, and it may be assuming that character lengths are the same as byte lengths.  Maybe only ever worked for 7-bit Ascii or ISO8859/1.

But it is worth grouping these together, and disposing of them together, one way or another.

MattHeffron:
> I think that PROOFREADER-PROOFREADER.TEDIT is the proper name, given the hyphen convention

For all the other `PROOFREADER-*.TEDIT` files, the corresponding source and LCOM files also have the `PROOFREADER-` prefix. For this file, the corresponding source and LCOM files do not. 
I thought that the prefix was to designate _component_ files of a _module_, and the _main_ file of the module wouldn't have the prefix. (For example, `library/tedit/TEDIT.LCOM` _vs_ `library/tedit/TEDIT-*.LCOM`) 
In that case, it seems that maybe some of the other files should have different prefixes as they are, according to their documentation, separate modules, but associated with `PROOFREADER`.

rmkaplan:
As I recall, there are 2 reasons (benefits?) for this foo>foo-xxx complexity.

One had to do with getting the code for image objects.  A function on foo-xxx that needed an image object definition could go to foo>foo as a last resort (strip off the -xxx).  But if the file was just xxx, it might be out of luck.

In this case, if there had happened to be a dependency like that (which there isn't), it would get the image object code from proofreader>PROOFREADER by decomposing the name PROOFREADER-ANALYZER.

The other is that if you want to find the code for a function in the XXX module (e.g. for PF or tf)  that was stored under just lispusers>foo>, the generic file searcher wouldn't find it unless lispusers>foo> was explicitly in DIRECTORIES.  But if you give it the name foo-XXX (e.g. from WHEREIS), it will go to foo>foo-XXX as a last resort.

In this case, (TEDIT 'PROOFREADER-PROOFREADER) should find lispusers>proofreader>PROOFREADER-PROOFREADER.TEDIT, and (TEDIT 'PROOFREADER-ANALYZER) should find lispusers>proofreader>PROOFREADER-ANALYZER.TEDIT).

(The PROOFREADER prefix is a little heavy, but that did seem to be the high-level entry for this).

MattHeffron (approved):
OK, since there's no effect to current loadups.

Issue #2210, opened 2025-07-09
[lafite/tedit] Lafite messages display with a bogus window title


nbriggs:
**Describe the bug**
A message displayed in Lafite should be in window with a title that reflects its source -- e.g.,

`Message 23 from {dsk}<Users>briggs>Active.mail;1 [1569 chars]`

but instead it shows up with

`See # <Input Nodircore Stream/147,7000>`

**To Reproduce**
1. Files loaded: LAFITE, LAFITE-UNIXMAIL
2. (LAFITEMODE 'UNIX) (LAFITE :ON) browse any Lafite .mail file, display any message from the mail file.
3. What happened? '(see above)

**Expected behavior**
(see above)

**Screenshots**

<img width="386" height="140" alt="Image" src="https://github.com/user-attachments/assets/38a3bcd3-d94d-4f3a-ad95-cab192a05010" />

PR #2222, opened 2025-07-21
Inspectcode scrolls, has the correct window title

rmkaplan:
INSPECTCODE now calls TEDIT - READONLY so that it has a process, instead of OPENTEXTSTREAM (which was also specified as readonly but without a process no sure what that does--TBD).

Also eliminated the override of the given title by the default.  That may also fix the lafite title problem.


MattHeffron:
How much, if any, of my `lispusers/INSPECTCODE-TEDIT` package (from 1985-1987) could/should be merged into `\TEDIT.INSPECTCODE`?

rmkaplan:
Given the changes I just made, the current OPENTEXTSTREAM... function and its changename have to be modified to use TEDIT instead, with the arguments for TEDIT.

I did that, and tried a little bit.  The inspectcode menu item worked, but mostly I got bells when I tried graphcalls--I didn't have it loaded at first, but still got bells after I loaded it.  

Is it correct that the prettyprint item only prints variables, not function definitions?  If TEDIT-PF-SEE is loaded, then meta-T (for "tedit-function") will print the function def in a new window.  Maybe that should be extended for other types.

If this is incorporated, probably the word-bound table should be initialized to the atom-bound table, like TEDIT-PF-SEE or Tedit of lisp sources.  So middle button then  tends to select the whole atom, even with dots, hyphens, slashes.

And maybe use TEDIT.PROMPTPRINT to print the error messages in the local prompt window, without the bells?

rmkaplan:
@nbriggs , can you check to see if this fixes the Lafite title problem in  #2210 ?  I get an NS error when I try to start up Lafite.

nbriggs:
@rmkaplan -- Yes, this fixes the Lafite title problem in #2210 .  Thanks.

nbriggs:
Any idea why every time you left-click or middle-click in the title bar of the Inspectcode window it adds a process (which seems to exit almost immediately) -- `ADD.PROCESS: Arg 1 = (TEDIT.DEFAULT.MENUFN (QUOTE {WINDOW}#157,nnnn)))`

rmkaplan:
I think it is trying to execute the menu actions in a different process.  I assume they had a good reason (not lock up the mouse for a long operation?), but maybe not?

rmkaplan:
I changed the FILETYPE for LLDISPLAY to :FAKE-COMPILE-FILE.  Note that the LCOM file say that it was COMPILE-FILEd even though it was actual FAKE-COMPILE-FILEd.

Also, (FINDCALLERS :FAKE-COMPILE-FILE  --) and (FINDCALLERS 'FAKE-COMPILE-FILE --) give the same hits.  Is that a good thing?

nbriggs:
Yes, I'd noticed that it still said "COMPILE-FILEd" - probably fine, since it's faking COMPILE-FILE...

I bet FINDCALLERS is using the equivalent of the PRIN1 name of the atom which ignores the package.  It's handy if you don't know what package the thing you're looking for is actually in (after any imports (and so on) are resolved)

MattHeffron (approved):
LGTM and seems to work with PR #2224

PR #2223, opened 2025-07-23
Removed a call to CENTRALWINDOW in Tedit's modernized buttoneventfn

rmkaplan:
Uses Tedit's own knowledge of its multi-pane window layout.  Test case is Lafite's mail scavenger


MattHeffron (approved):
LGTM

PR #2216, opened 2025-07-16
Rmk103  font and related code updates

rmkaplan:
This includes changes to cleanup and simplify (somewhat) the interface and some of the internals of the FONT architecture.  It also includes MEDLEYFONTFORMAT, the implementation of the new read/writable format for display and other device fonts.  The changes to FONT are documented in docs/internal/FONTCODECHANGES.TEDIT, the document for MEDLEYFONTFORMAT is in docs/internal/MEDLEYFONTFORMAT.TEDIT.

Around this central core there are a bunch of files with small changes for compatibility with some of the font changes.  E.g. the recursion surrounding the validation of font attributes has been flattened out, and almost all calls to \COERCEFONTDESC have been eliminated (addressing #2100).

And the loadup sequences and have been adjusted and a few functions have been moved so that medley-format fonts can be included in the MAKEINIT.

This PR includes only 2 medley-format fonts, for GACHA10 and HELVETICA10.  GACHA10 is created in the MAKEINIT. HEVETICA10 a little later, when MENU is loaded.  I've included them here to test that everything hangs together.

(Separately, I have created medley-format fonts for all the display fonts, but I want to release those in a separate PR after this code has settled down.  The new NS fonts will be MCCS-encoded, and that will require some other adjustments.)

PR #2241, opened 2025-08-04
Fix Gacha display fonts

rmkaplan:
Take out the goofy glyph at character 0 in Gacha 12, add underscores at 30Q for Gacha 8 and 10, and also add updated Medleydisplayfont for Gacha 10.


masinter (commented):
fonts/medleydisplayfonts/GACHA10-MRR.MEDLEYDISPLAYFONT
is also updated by Rmk103 font and related code updates #2216

PR #2245, opened 2025-08-08
GETMULTI in MULTI-ALIST is a Place

rmkaplan:
So it can be an argument to various left-handed (CL:SETF, change, add) operations.  This also simplified the implementation.

For example, to build a multidimensional histogram HIST with coordinates C1... CN, initialize with
(SETQ HIST NIL)
(PUTMULTI HIST C1 ... CN 0)
and for each instance do
(change (GETMULTI HIST C1 ... CN) (ADD1 DATUM)


rmkaplan:
I'm closing this one to get files in sync, will push one with additional changes.

Issue #2221, opened 2025-07-20
Remove  lispusers/PS-SKETCH-PATCH ?


rmkaplan:
It is not part of the loadup and is not loaded when Sketch is loaded or when Sketch hardcopies are done.

Whatever problem it was addressing must have been resolve in some other way. But it does overload some FONT functions, which creates a confusion.

Can it be moved to obsolete, or simply deleted?

MattHeffron:
I would suggest moving it to `obsolete`, just in case we later see some problems in sketch (possibly with the HTML stream). We'd have a starting place for looking at fixing.

(Does loading this **_break_** anything??)

rmkaplan:
The replacements for \BUILDSLUGCSINFO and \CREATECHARSET will probably screw things up.  The rest of it seems to just be fiddling around with different patterns of coercion.

PR #2248, opened 2025-08-09
Move lispusers/PS-SKETCH-PATCH to obsolete/lispusers/

rmkaplan:
Incompatible with current font interface, not loaded and sketch seems to work fine without it


rmkaplan:
Fixes #2221

PR #2249, opened 2025-08-10
Update loadup and medley scripts to use "tag" instead of "branch";  add ability to completely delete tagged loadups rather than just thin them 

fghalasz:
This PR implements the following:

1.  In loadups and medley scripts the term "branch" is replaced by the term "tag".  The new flags are "--tag" and "-tg".   Under the loadups directory, tagged loadups are now kept in MEDLEYDIR/loadups/tagged (as opposed to MEDLEYDIR/loadups/branches).  A symlink from "branches" to "tagged" is automatically created for backward compatibility.

2.  The --thinw argument to the loadup script now has an optional "+" parameter.  Without the "+", the versioned files in the working directory (usually called loadups/build or loadups/tagged/TAG/build) are deleted.  With the "+", all files in the working directory as well as the working directory itself are deleted.

3.  The --thinl argument to the loadup script now has an optional "+" parameter that work similarly to the that of the --thinw argument.  Note that --thinl does not delete any files in the working directory, just at the top level of the loadups directory.  If there is no working directory under the loadups directory (and the loadup is tagged), then the entire tagged loadup is deleted.

4.  There is a new --thin (-th) argument to the loadup scripts, which is equivalent to --thinw followed by --thinl.  If the "+" parameter is used with --thin and the loadup is tagged, then the entire tagged loadup (including the MEDLEYDIR/loadups/tagged/TAG directory) will be deleted.

5.  Fixed Issue#2063 - typo in medley man page.

6.  In loadup script, --help (-h) command line argument has been added as a synonym for --man (-z).

7.  In both medley and loadup scripts,  --man (-z) now accepts an optional parameter (+) which causes the man page to be displayed without a pager (or more precisely, with the pager set to cat).  This allows you to use your terminal's scrolling for moving back and forth in the man page instead of the arcane commands in less or more or whatever pager is the default on your system.

8.  The man pages for the medley and loadup scripts have been updated accordingly.

9.  Fixed Issue#2251:  LOADUPSDIRECTORIES in Medley now set correctly when using a tagged loadup.


pamoroso:
I checked out this PR on Linux Mint 22.1 Cinnamon. It doesn't break my build process and I haven't noticed anything unusual.

pamoroso (approved):
I built this PR on Linux Mint 22.1 Cinnamon and I'm not noticing anything unusual.

masinter (approved):
as per 8/20/25 meeting

PR #2254, opened 2025-08-12
Updated TMAX to separate window vs textstream properties

rmkaplan:
I did a lot of reworking some time ago, but didn't push.

Basically, TMAX (like a number of other Tedit applications) assumed that there would always be a window to hang its state on, not recognizing that only the textstream and textobj exist in many scenarios.  Moving things around and making them less depend on Tedit internals was the major focus here.

Plus some bug fixes and other improvements.

The reason I let it hang is that I know of one glitch that I haven't figured out how to fix.  There is a Find menu item on a cross reference which is supposed to find and display the main item.  It finds it, but the display is a little goofy.  I think that's because of an underlying design image object design issue, where the coordinates change from the window's coordinate system to the object's before some imageobject functions are called and not others.  I'll look more carefully at that.

In the meantime I want to get this out.  To get a sense of what this provides, look at its documentation (TEDIT 'TMAX).


pamoroso (approved):
I tested the PR on Linux Mint 22.1 Cinnamon and it builds and works with no apparent issues.

PR #2256, opened 2025-08-13
Add charset functions in LLREAD, move IOCHAR a little bit earlier

rmkaplan:
CHARSET.DECODE, CHARSET.ENCODE added.  CHARACTER.ENCODE moved here from TEDIT-FNKEYS. Some character names moved here from a few other files.

This anticipates the ability to use names for the charset argument to FONTCREATE.  IOCHAR needed to move a little earlier so that STRING.EQUAL/UPPERCASEARRAY is available before FONTCREATE creates the guaranteed displayfont.


rmkaplan:
P.S. This is simple and should be uncontroversial, but it is required for the Tedit cleanup that is queued up.  So it would be good to get this examined, approved and merged soon.

pamoroso (approved):
I built and tested this PR on Linux Mint 22.1 Cinnamon and I haven't noticed any issues.

PR #2259, opened 2025-08-15
Tedit PLOOKS cleanup plus default paralooks

rmkaplan:
As usual, quite a few files.  But most of them have to do with eliminating the confusion between PLOOKS sometimes meaning the _character_ looks of a piece and sometimes meaning the _paragraph_ looks of document.

This also adds a user settable TEDIT.DEFAULT.PARALOOKS that define the paragraph looks when an empty document is created (for an existing document, every paragraph has its own looks and new paragraphs inherit from their neighbors).

Also CHARCODE.ENCODE and some character names that now exist in LLREAD have been removed from TEDIT-FNKEYS.


pamoroso:
This builds and works fine on Linux Mint 22.1 Cinnamon but `TEDIT.DEFAULT.PARALOOKS` was already in master for me.

rmkaplan:
The para looks were included in the patch for the rmk103 merge.  This includes the related updates, plus the documentation.

pamoroso (approved):
It builds and works.

PR #2260, opened 2025-08-15
Remove the LEAVETTY from the PF-TEDIT Tedit window

rmkaplan:
Corrects a UI glitch wherein typing meta-t on an atom in a Tedit source window brings up a window with the corresponding function definition, but then the original window flipped back on top because it still had the ttyin.  

At the same time, I lower-cased the clisp words for stylistic consistency.


pamoroso:
I can't reproduce the issue in master. I open an Interlisp file in TEdit, select an atom, and press `Meta-t`: the mouse pointer briefly flickers but no other window is opened.

rmkaplan:
Are you looking at a source file or definition in a Tedit window,?   E.g.

tf COPYFILE
select DIRECTORYNAMEP
type meta-t

DIRECTORYNAMEP should show in a new window.  If it selected a region that overlaps with the one showing COPYFILE, COPYFILE was coming back on top.

pamoroso:
Never mind, I can reproduce the issue in master. I need of course to use `TF` or `PF-TEDIT`, not regular `TEDIT`. However, the original window doesn't flip on top but does still have the caret.

rmkaplan:
I think meta-t should work (with the same issue) if you TEDIT or SEE a source file and select there.  That is, you should be able to use regular Tedit, but meta-T only works on a Lisp source (although perhaps it should also be made to work on a man page).  (separately, meta-d should be armed here as in Sedit, so that it would bring up the man page, meta-o should open Sedit.)

For a Lisp source or definition, it gets the file's readtable and uses that to determine the boundaries of the atoms.  If you are pointing to the left part of ABC.FOO, it will pick the whole thing, whereas it would only pick ABC if it thinks you are looking at an ordinary text file.

pamoroso:
I confirm `Meta-t` works also with regular `TEDIT` and Lisp sources but I'm unable to make the original window flip on top.

pamoroso (approved):
I tested the PR on Linux Mint 22.1 Cinnamon. Now the new definition window brought up by `Meta-t` stays on top and does have the TTY.

PR #2261, opened 2025-08-15
Extensions to MULTI-ALIST especially for PLACE

rmkaplan:
The main improvement is that putting macros can be given commonLisp places and the getting macros return a commonLisp places (so they can be wrapped in CL:SETF, change, add...).

For example, to build a multidimensional histogram with indexing keys C1...Cn in a HIST field of a datatype bound to X, for each instance c1...cn  do
(change (GETMULTI (fetch HIST of X) c1 ... cn) (ADD1 (OR DATUM 0))))

Also added a little bit more to the interface, cleaned up the implementation, and updated the documentation.


pamoroso (approved):
No build or usage issue on Linux Mint 22.1 Cinnamon.

PR #2268, opened 2025-08-24
Fix MEDLEYDIR to handle filename if given

masinter:


pamoroso (approved):
I built and tested this PR on Linux Mint 22.1 Cinnamon with no issues.

PR #2270, opened 2025-08-24
Qualify record names in MASTERSCOPE COMPAREDIRECTORIES TTYIN

rmkaplan:
To avoid conflicts when creating fuller.database.

There are still conflicts left.  One in particular concerns INVERT.BITMAP.VERTICALLY.  There are different definitions of this function in 2 files, SCREENPAPER and EDITBITMAP.  

If they are equivalent, is one better than the other?  If not equivalent, what's the difference?


nbriggs:
The version in SCREENPAPER should definitely be faster, it has an extended argument list though it is compatible with the one in EDITBITMAP.  It's not just INVERT.BITMAP.VERTICALLY, though - there's also INVERT.BITMAP.HORIZONTALLY.

The SCREENPAPER versions use PILOTBBT directly to do line-at-a-time, the EDITBITMAP versions use BITMAPBIT to do a bit at a time.

rmkaplan:
So, should the versions in SCREENPAPER, which is not in the loadup, be moved to EDITBITMAP, which is?

nbriggs:
I would leave it be.  I don't think it's reasonable to require that we not have multiple function definitions because the function lookup code is deficient.

nbriggs:
Something that just occurred to me - the EDITBITMAP versions will probably work on BIGBM bitmaps - the SCREENBITMAP version almost certainly will not, since the PILOTBBT opcode won't know what to do with a BIGBM - perhaps it'll punt, but probably not.

rmkaplan:
SCREENPAPER is a lispusers idle hack.  Doesn't seem right for that to be the reason that the database is confused.

If the one in SCREENPAPER can't replace the other one, what if we eliminate the one in SCREENPAPER and let it use the slower one.  Does the speed for this hack matter on modern machines?

Or, rename the one in SCREENPAPER to something else?

masinter (approved):
is there an issue about INVERT.BITMAP.VERTICALLY

rmkaplan:
Yes, there are 2 different implementations, we should pick one.

nbriggs:
I think we need to fix things so that it is possible to locate the version of a function that is in use, and alternatives, not just eliminate the possibility of duplicate named functions in the files that are known

PR #2271, opened 2025-08-28
EDITFONT: Column-major order to match XCCS and Unicode tables

rmkaplan:


pamoroso (approved):
On Linux Mint 22 Cinnamon I tested up to commit ec6d820 and it's working as expected.

masinter:
would it be a good idea to make EDITFONT and FONTSAMPLER look and act the same?
I wanted to make HCFILES display "fonts" by printing them out (FONTSAMPLER) in a "non-interactive" manner.

Issue #2266, opened 2025-08-23
Move PRESS to obsolete?


rmkaplan:
There may already be an issue that discusses this, I can't find it.

library/PRESS is included in the loadup, but it doesn't work.

Taking GACHA to be the prototypical Press/Alto font,
     (FONTCREATE 'GACHA 10 'MRR 0 'PRESS )
gives you a font-not-found error.  Same for any other font you might ask about.

It appears to work in the Venue sysout (no error), but the font descriptor that is created has no character sets.

Is it worth trying to debug and maintain this, or just move it to obsolete (along with PRESSFROMNS)?

PR #2274, opened 2025-09-01
Rmk120  move press to obsolete

rmkaplan:
As discussed in the meeting 9/1/2025 and in #2266....

PRESS was not functional even in the Venue sysout, not worth fixing and maintaining.

This moves PRESS and PRESSFROMNS to obsolete, and takes PRESS out of the loadup.

The only dependency on PRESS outside of PRESS that I could find was the function PRESSFILEP used by BACKGROUNDIMAGES.  That particular function is now included in BACKGROUNDIMAGES, to break the file dependency.  The updated BACKGROUNDIMAGES still works and compiles, even in the absence of PRESS.


pamoroso (approved):
I tested this change on Linux Mint 22.1 Cinnamon and BACKGROUNDIMAGES still works with no issues.

PR #2278, opened 2025-09-06
Update HIPPO18-MRR-C0.DISPLAYFONT

rmkaplan:
One of the widths fields was about 100 times bigger than it should have been, causing bitmap allocation failures.


pamoroso (approved):
I tested the PR on Linux Mint 22.1 Cinnamon and the Hippo 18 font is working with no apparent issues.

PR #2279, opened 2025-09-06
Fix TEDIT object insertion

rmkaplan:
Internal function called with a textobj instead of a textstream (e.g. under COPYINSERT when snapping and inserting a bitmap object).


pamoroso (approved):
I tested the PR on Linux Mint 22.1 Cinnamon and noticed no anomalies.

Issue #2178, opened 2025-06-01
Alto fonts (SYMBOL, HIPPO, SIGMA) on hardcopy devices


rmkaplan:
We have files and datastructures (strings, image objects, Tedit files) that have characters in idiosyncratic (Alto/Ascii) encodings.  

Tedit coerces those character codes into MCCS codes and fonts, so that typing and searching is uniform.  But the various output devices have to make sure that their \OUTCHAR functions also do the proper coercions, based on the fact that a DSPFONT was executed to set up the font for the subsequent characters.

The coercion tables and translation functions into MCCS should be (will be) defined systematically.  But maybe the generic DSPFONT code should recognize from the font that what it will be asked to output is in the font's character encoding and set it up so that the outchar function does the "right" thing.  The right thing may be to select a different font for the output device and set up the translation from the font-encoded input characters to the charencoding of the selected output font.  (If the output font is always an MCCS font, then it would be one of the defined translation  functions to MCCSl otherwise the device would have to create its own translation for each substitution.)

Issue #2135, opened 2025-05-03
Medley font-file format


rmkaplan:
The problem that Matt pointed out with font recoding, is that we only have a function for writing strike-format files, not for AC format or other display formats, or other hardcopy formats.  And the strike format doesn't support all of the information that some of our fonts now have.

As mentioned in #2120 , this is a stumbling block in the effort to clean up the lingering confusion between our internal character codes (MCCS) and the various fonts we have for rendering those codes.  I now have the tools for moving character information around in and between fonts, but there is no obvious way of saving that information for later reuse--we would have to write/debug code that accurately conforms to the AC legacy format.

The alternative idea, doing the recoding on the fly when a font is being read, runs up against the fact that font-reading is done on a character-set-within-font basis; anything that crosses character sets or fonts is incompatible with this basic architecture.  That's also not really an option.

So, maybe time for a fresh approach:  define and implement our own Medley-natural font-file format.  We can then read legacy fonts from legacy formats, rearrange and otherwise modify them in memory, and then write them out in a Medley-specific format that we know has all the information contained in our font descriptor and character-set data structures.  These files would only be for Medley use; we have no need to export our fonts in a standard format that other applications might want.

For starters, here's a really simple strategy for writing a character set from a given in-core font-descriptor, no matter its provenance:  
Create a structure, say an alist, that has a pointer to the charset-info plus whatever properties of the fontdescriptor that we want to associate with the character set (e.g. family, size...charencoding...).  Then open a file with an appropriate name and device-extension, print an identifying header string (e.g. "Medley character set") to distinguish this from any other font-format, and HPRINT the structure.  When the generic charset-reading code has seen/read the header string, it calls HREAD to recover the alist and unpacks it into the various font-descriptor locations.  No need for fancy printing or parsing code.

A minor extension would be to include a character-set dispatch vector on the front, so that all the character sets of a given font could go into a single file. Character sets could still be loaded as needed, but without requiring the multi-level directory structure.

masinter:
Seems almost everything you need for a generic cache function-call facility.


cache-calls (fn) redefines fn to check a hash-table for EQUAL (or EQUAL-ALL) hash-key ... if found, return the hash value, otherwise compute fn and store the results in the hash table.

setup: cache-calls on FONTCREATE and READ-PDF.  to save from one session to the next, write out the hash-table as a hash-file.

I'm wary of baking in too much specifics about the font architecture ....

masinter:
See a;sp issues
[...]

Issue #2124, opened 2025-04-25
The WIDTH/IMAGEWIDTH distinction is not propagated through MOVECHARBITMAP PUTCHARBITMAP GETCHARBITMAP


rmkaplan:
I am looking at the infrastructure in FONT for moving character bitmaps, as part of the coercion of XCCS fonts to MCCS fonts (dollar, underscore, caret). 

MOVECHARBITMAP essentially does PUTCHARBITMAP of GETCHARBITMAP.  It appears to me that the bitmap is constructed using the IMAGEWIDTH, which defaults to WIDTH, so the bitmap is valid.  But the fact that the IMAGEWIDTH of a character might be different from the (advance) WIDTH is not carried over to the destination.  Nor is the leftkern information, if present in the source.

I think this means that e.g. diacritics and italic-advancing would be screwed up if those glyphs were moved from one place to another, even if in the same font descriptor.

So I think that GETCHARBITMAP should return a more complicated structure when the full collection of properties can't be determined just from the bitmap. Maybe a property list that includes the bitmap plus these other properties.

For backward compatibility, perhaps GETCHARBITMAP should take a flag to say that it should provide this information.

Also, these functions have a validity test for the character codes, allowing strings and litatoms and coercing them to their first character.  I think it would be better to call CHARCODE.DECODE.

rmkaplan:
On further thought, it would be better to leave GETCHARBITMAP alone, it just gets the bitmap.  But add a higher level function to fetch and store  all the information.  Maybe an entry MOVECHAR that does everything, keep MOVECHARBITMAP for backward compatibility.

rmkaplan:
I have written a function MOVEFONTCHARS that takes a code/font mapping and copies all of the relevant information.  I have also used an otherwise unused  field in FONTDESCRIPTOR to record the FONTCHARENCODING (XCCS, MCCS, ALTOTEXT...).

rmkaplan:
This has been accomplished

Issue #2120, opened 2025-04-23
Medley character encoding and external format


rmkaplan:
This is a proposal for cleaning up the confusion between the internal MCCS character encoding and external format, the Xerox XCCS encoding/standard, Unicode, etc.

To start, we clarify that MCCS is the character coding standard used internally.  For example, the mappings provided by CHARACTERSETNAMES and CHARACTERNAMES go from names to MCCS values.

MCCS is a derivative of, but not identical to, the Xerox character encoding standard.  The well-known differences are the codes for dollar, underscore, and caret. To make this explicit, the functions (MTOXCODE MEDLEYCODE) and (XTOMCODE XCODE) will map between the 2 different encodings.  MCODE x24 corresponds to XCODE xA4, and vice versa, underscore interchanges with left arrow, caret switches with uparrow.   

:MEDLEY will be the default external format if another format isn't specified for reading or writing.  (:XCCS must be specified to read a Xerox-encoded file, if we ever want to do that.)

The :Medley external format represents character codes using the shifting stringlets of XCCS,  but  reading and writing is completely transparent, no codes are translated.  In contrast, either XTOMCODE or MTOXCODE is applied to every code decoded from a :XCCS file. depending on reading or writing.

(Note that the key-action tables will produce the MCCS codes for the keys labeled $, ^, _.  We'll need some other keyboard convention to type in currency, caret, or underscore.)

UNICODE:
With respect to Unicode,  functions MTOUCODE and UTOMCODE will map from Unicode codes (in an UTF-8 external format) to Medley codes.  Those mappings will be derived from our XCCS-to-Unicode tables in the Unicode/xerox directory, essentially by executing either (XTOMCODE (UTOX CODE....)) or (MTOXCODE (XTOUCODE ...).  These compositions will be applied when the mapping vectors are compiled from the data files, to produce the resource files with explicit names UNICODE-TO-MEDLEY.TXT and MEDLEY-TO-UNICODE.TXT.  These will replace the current UNICODE-MAPPINGS.TXT and INVERTED-UNICODE-MAPPINGS.TXT, which are implicitly defined on XCCS.

FONTS:
In-memory fonts will have glyphs in positions that correspond to the MCCS codes.  If those are read from XCCS font files (Classic, Terminal...), the glyphs will be moved on reading according to the XTOMCODE mappings.  Thus, the uparrow glyph will end up at x5E, the caret glyph will end up at xAD.

[...]

PR #2280, opened 2025-09-08
rmk122--Next round on fonts and MCCS

rmkaplan:
Many files have changed as I have rearchitected the font interfaces, introduced the Medley font-format files, and worked to normalize to the MCCS encoding of internal data.

This PR is a draft that doesn't yet include the flood of files.  For starters, it just has the documentation files that describe the strategy and new and changed features.  It will be helpful to get some initial reactions so I can make adjustments before pushing everything.  The code and font files will come later.


pamoroso:
On Linux Mint 22.1 Cinnamon this PR builds with no issues and I didn't notice anything unusual. I tested DInfo and NoteCards and the fonts look fine so far.

rmkaplan:
I have uploaded all of the medleyformat fonts, and removed the draft status.

I expect/hope the only issue is that filenames with underscores and carets might not look the same in Medley vs Unix.  I would like to get this tested and merged, to avoid confusion with the changes to the imagefile/hardcopy interfaces.

rmkaplan:
I found that PRETTYFILEINDEX had its own compensation for underscore/arrow, at least for Interpress.  Removed it.

rmkaplan:
@MattHeffron , LOADUPFULLFONTS on internal/loadups/LOADUP-FULL has some special code for building the postscript fontcache.  I wonder whether this should be done when POSTSCRIPTSTREAM is loaded, as part of POSTSCRIPT.INIT.

pamoroso:
I updated to commit e46a1f3, nothing unusual to report.

pamoroso:
I updated to commit 3276822 and I get an error from GITFNS, here's how to reproduce it:

1. load GITFNS
2. evaluate `PRC`
3. click issue `#2290`
4. click `sources/COREIO.`
5. click `See`

A break window shows up with this error:

```
INTERLISP-ERROR
In ERROR:
ARG NOT DISPLAY STREAM
NIL
```

gitfns-prc-error

The backtrace:

```
CD-See/8(debug)BTV
   MESS1 "ARG NOT DISPLAY STREAM"
   MESS2 NIL
   NOBREAK NIL
ERROR
\DTEST.UFN
   DS #<IO Tedit Stream/145,142500>
   DONTCREATE NIL
   DD NIL
   HINTW NIL
WFROMDS
   CHILDREN (#<IO Tedit Stream/145,142500> 
#<IO Tedit Stream/127,20200>)
   PARENT {WINDOW}#120,150570
   C #<IO Tedit Stream/145,142500>
   ONE NIL
CLOSEWITH
   TBITEM {TABLEITEM}#120,117770
   MENUITEM See
   CDBROWSER {TABLEBROWSER}#120,116700
   KEY LEFT
   SOURCEWIDTH 770
   CHILDREN (#<IO Tedit Stream/145,142500> 
#<IO Tedit Stream/127,20200>)
CD-MENUFN
   CDBROWSER {TABLEBROWSER}#120,116700
   TBITEM {TABLEITEM}#120,117770
   CDENTRY (SOURCES>COREIO (
"{FROMGIT1}/master/sources/COREIO." 
" 5-Jun-2022 00:14:07" 58450 NIL SOURCE LF) < (
"{FROMGIT1}/pr/sources/COREIO." "11-Sep-2025 16:49:07" 
57998 NIL SOURCE LF))
   FILE1 "{FROMGIT1}/master/sources/COREIO."
   FILE2 "{FROMGIT1}/pr/sources/COREIO."
   TYPE SOURCE
   LABELS ("origin/master/sources/COREIO." 
"origin/rmk126--HARDCOPY-to-printers/sources/COREIO.")
   LABEL1 "origin/master/sources/COREIO."
   LABEL2 
"origin/rmk126--HARDCOPY-to-printers/sources/COREIO."
CD.COMMANDSELECTEDFNA0001A0002
TB.MAP.SELECTED.ITEMS
   WINDOW {WINDOW}#120,150570
   PWINDOW {WINDOW}#120,150264
   CDBROWSER {TABLEBROWSER}#120,116700
   USERDATA (LABELFN GIT-CD-LABELFN BRANCH1 
origin/master BRANCH2 
"origin/rmk126--HARDCOPY-to-printers" PROJECT (
GIT-PROJECT MEDLEY {MEDLEY} NIL & & 
"{DSK}/home/paolo/medley/medley" origin/master) CDVALUE 
(& & & & & & &))
   CDVALUE (("{FROMGIT1}<master>" "{FROMGIT1}<pr>" 
"23-Sep-2025 11:36:50" & & &) (SOURCES>COREIO & < &) (
SOURCES>COREIO.LCOM & < &) (SOURCES>FILEIO & < &) (
SOURCES>FILEIO.LCOM & < &) (SOURCES>HARDCOPY & < &) (
SOURCES>HARDCOPY.LCOM & < &))
   FN CD-MENUFN
   MIDDLE NIL
SI::*UNWIND-PROTECT*
   MENUITEM (See CD-MENUFN)
   MENU {MENU}#130,171054
   KEY LEFT
   LISPXHIST NIL
   SI::*RESETFORMS* NIL
   RESETSTATE NIL
CD.COMMANDSELECTEDFN
   *FORM* (CD.COMMANDSELECTEDFN (QUOTE &) (QUOTE {MENU}#
130,171054) (QUOTE LEFT))
   *ARGVAL* NIL
   *TAIL* NIL
   *FN* CD.COMMANDSELECTEDFN
\EVALFORM
   %#FORM# (CD.COMMANDSELECTEDFN (QUOTE &) (QUOTE {MENU}#130,171054) (QUOTE LEFT))
   *CURRENT-PROCESS* #<Process CD-See/173,125000>
   HELPFLAG BREAK!
   \CURRENTDISPLAYLINE 0
   \#DISPLAYLINES 1
   \LINEBUF.OFD #<IO Linebuffer Stream/167,74100>
   *READTABLE* #<ReadTable INTERLISP/173,161714>
   \PRIMTERMTABLE {TERMTABLEP}#173,154740
   \PRIMTERMSA {CHARTABLE}#173,155000
   TtyDisplayStream #<Output Display Stream/145,142000>
   SI::*RESETFORMS* NIL
   \INTERRUPTABLE T
   \TTYWINDOW NIL
   READBUF NIL
   \TERM.OFD #<Output Display Stream/127,4700>
   *STANDARD-OUTPUT* #<Output Display Stream/127,4700>
   *STANDARD-INPUT* #<IO Linebuffer Stream/167,74100>
\MAKE.PROCESS0
T
```

rmkaplan:
It looks like a minor change leaked in from the imagetype PR that I'm currently staging.  I backed that out of TEDIT-STREAM, please try again.

(The change is to now distinguish imagetype TEDIT from imagetype TEXT.  But WFROMDS in WINDOW also has a dependency on that Tedit-internal value, unfortunately.)

pamoroso:
Commit 49d15bd fixed the issue and `PRC` now works as expected.

rmkaplan:
I updated GITFNS again.  The final change to convert Unix filenames to and from MCCS is still pending.  Until then, it turned out that the move in GITFNS to use PR-named subdirectories of {DSK}/tmp/ instead of {CORE} for scratch files temporarily requires its own translation for PR names with underscores in their names (#1960 and #2292).

pamoroso:
Updating to commit b0c333e still works with no issues.

rmkaplan:
I thought that Palatino was an XCCS$ font, like Classic.  But it isn't.  Its 8-bit region seems to overlap with XCCS$ from time to time, but includes a bunch of accented characters and other characters in various positions.

Is this a different known font encoding, or totally idiosyncratic ?  It requires a different translation table to convert into MCCS.

nbriggs:
Envos was shipping these (listing only the MRR faces):
```
./xd0e/lde/fonts/adobe/palatino12-mrr-c0.displayfont
./xd0e/lde/fonts/adobe/palatino14-mrr-c0.displayfont
./xd0e/lde/fonts/adobe/palatino09-mrr-c0.displayfont
./xd0e/lde/fonts/adobe/palatino24-mrr-c0.displayfont
./xd0e/lde/fonts/adobe/palatino06-mrr-c0.displayfont
./xd0e/lde/fonts/adobe/palatino08-mrr-c0.displayfont
./xd0e/lde/fonts/adobe/palatino18-mrr-c0.displayfont
./xd0e/lde/fonts/adobe/palatino10-mrr-c0.displayfont
```
So I would assume it's in the Adobe Postscript StandardEncoding.   Pg 596 ff. of the PS language reference manual, 2nd edition.  or Wikipedia https://en.wikipedia.org/wiki/PostScript_Standard_Encoding

nbriggs:
... and you'll also find ISOLatin1Encoding, ExpertEncoding, ExpertSubsetEncoding and SymbolEncoding vectors.

rmkaplan:
Our Palatino has lots of character in the upper ctrl region, which seem to be unoccupied in the Adobe tables.  The Adobe tables are in row-major order, unlike the XCCS and Unicode tables.  I'll put that back as an option in EDITFONT.

Palatino (also Titan) is used in the 93 IRM and dinfo pages.  In sizes and faces that don't exist for the display and have to be faked up.

pamoroso:
I tested up to commit c9ed0b8 and I have nothing unusual to report.

pamoroso:
I tested up to commit 7c60baa, still nothing unusual.

pamoroso:
While on this branch I activated NoteCards, created a new notefile, created a text card, and shift-copied into it the code of function `CREATEW` open in SEdit.

Next, in the text card I selected all the text and, from the character looks menu, `APPLY`ed Terminal 10. This resulted in a break window with an error which, unfortunately, I didn't write down and no longer remember as I immediately tried to reproduce the issue without success.

I don't know whether this is related to the font work in progress but I'll keep an eye out.

nbriggs:
I haven't reproduced this particular failure, but it doesn't surprise me that you saw an error because there are a lot of weird TEDIT (mouse) selection issues, and one could have resulted in it trying to apply looks in a way that causes an error.

rmkaplan:
On Tedit mouse, is it that the highlighting of a selection doesn't come down when you abandon it?

nbriggs:
@rmkaplan - extending a selection with the right button, when starting with a line selection in the left margin, may (or may not) end up with an actual selection when you apply char looks.  It also seems to, for no obvious reason, end the selection at the right end of the last line, or put the caret on the next line after the inverted selection.  So perhaps it's related to whether the selection comes down when it should.

pamoroso:
On this branch, if I open a shell CHAT, call `DSPFONT` to set the font of the CHAT window to Terminal 10, and type the `_` key I get `←` in the CHAT window and when I type `^` I get `↑`: is this the expected behavior with MCCS? On master I get `_` and `^` in the CHAT window when typing the same keys.

rmkaplan:
Yes, that's the expected behavior.  This changes the keybindings so that shift-hyphen key transmits the MCCS left-arrow code, and shift 6 transmits the uparrow code.  To get the underscore and circumflex codes, you type function 10 and function 6.  (Also, to get the currency symbol code, should you ever want it, you type function 4).

This is just a change to the bindings in LLKEY, doesn't otherwise have to do with the fonts or other MCCS changes.

It's a question of whether we should try to match the labels on current keys, or labels on, say, the Alto keyboard--what would be more intuitive.

pamoroso:
In a VT100 CHAT window with the font set to Terminal 10, pressing the `F10`, `F6`, and `F4` keys, respectively, yields (second line):

chat1

If immediately later I press `Backspace` three times the caret advances by a few characters the first time as if I pressed `TAB`, then some spaces are deleted, and the caret finally ends up after `$` as shown here:

chat2

Continuing to press `Backspace` has no effect and the caret stays as in the above screenshot. If I immediately type `1` and then `ENTER` I get the error `1: command not found`:

chat3

This occurs on Linux Mint 22.1 Cinnamon, both with and without VNC. I'm on 7c60ba/rmk122--FONT-next-round for Medley and 853d133/master fot Maiko.

rmkaplan:
Can you send the equivalent screen shots from master, except using the underscore, cirumflex, and $ keys?

pamoroso:
The equivalent screenshots from master, first:

chat1

Second:

chat2

Third:
 
chat3

Pressing `Backspace` has a similar behavior.

rmkaplan:
FWIW I think the only distinct files were sizes 10 12 14 18 24.  6 and 8 were copies of 10, 9 was a copy of 12.  I no longer store files that are copies of other files, they are constructed on the fly according to the mappings on DISPLAYFONTCOERCIONS.  This has entries
(PALATINO 9) (PALATINO 12)
((PALATINO (<= * 8)) (PALATINO 10)

pamoroso:
On master I get the same `Backspace` behavior when typing anything, not just underscore or circumflex, but only with Terminal 10 as the default Gacha works fine.

rmkaplan:
I pushed a revised set of medleyformat fonts, plus a few code files with some glitch fixes.  The new fonts are built in a sequence that first moves character information in legacy encodings (XCCS$, Altotext...) to MCCS positions in all fonts, then replaces slugs with information it may grab from other MCCS fonts (according to DISPLAYCHARCOERCIONS).  Probably no difference from before for the standard encodings (which mostly had arrow/underscore shifts within a font), but I think the coercions for Palatino are now correct.

pamoroso:
I tested up to commit b747e0f and I have nothing unusual to report other the the persistence of the [Terminal font issue](https://github.com/Interlisp/medley/pull/2280#issuecomment-3385139945).

rmkaplan:
I loaded VTCHAT, selected Chat/shell from the background menu.  I see the left/right arrows when I type the keys they are assigned to (shift-6, shift hyphen), as I would expect, independent of the font.  But the function keys don't work.  I see a black string of f's for each of the keys, no matter which key of f4/46/f10 I press.  The other function keys just flash the window.

Screenshot 2025-10-10 at 9 21 43 AM

pamoroso:
I also see the left and up arrows when typing `_` and `^`. By function 10 and 6 do you mean the `F10` and `F6` keys?

rmkaplan:
I investigated a bit further, a little progress.

The input characters are handled by the function CHAT.TYPEIN.  It gets the codes from the keyboard (e.g. 0,255 for F-10, the MCCS underscore code) and puts it on the Chat OUTPUTSTREAM, a Unix-pty stream.  That's a UTF-8 stream, with outcharfn UTF8.OUTCHARFN.  That functions does the MCCS to Unicode coercion, mapping 0,255 to the Unicode underscore 0,137.

But CHAT.TYPEIN does a direct BOUT to the stream, so the translation doesn't happen.  That's a case that I apparently missed when I did the external format transformation.  With BOUT, the 0,255 gets put on the Unix stream directly, and presumably what I see as <fff> is the echo coming from the shell side--the shell doesn't like that character.

Changing the BOUT's to  \OUTCHARFN improves things.  The <fff> goes away but instead is replaced by the left-arrow glyph.  My bet now is that the window's display stream is actually just being used for echoing from the shell side, and the echo that's being provided is the already converted 0,137 (the MCCS left arrrow).  That is, the keyboards 0,255 is sent down as 0,137, and then send back for echoing as 0,137.

So that has to be translated the other way (UTOMCODE) before it is displayed in the new MCCS fonts.  I'm still looking at that.

nbriggs:
I think this is going to break the terminal emulation -- the CHAT code is processing the returned codes as ASCII bytes to emulate either (in the two common cases) a Datamedia DM2500 or a DEC VT100 terminal.  At times, the characters received are going to be used as binary values for things like cursor positioning.

nbriggs:
You should be able to download the ncurses examples from https://invisible-island.net/ncurses/ncurses-examples.html (untar the file, cd ncurses-examples-20211021, mkdir build, cd build, ../configure, make) then you'll find a collection of test/example/demo programs that make use of the "ncurses" package to drive ASCII graphic displays in the terminal.
Try, for example, `./xmas` in a dm2500 chat, and in a vt100 chat.

pamoroso:
Another useful resource is the [VTTEST](https://invisible-island.net/vttest) tool which "tests the compatibility (demonstrates the non-compatibility) of so-called “VT100-compatible” terminals"

rmkaplan:
> I think this is going to break the terminal emulation -- the CHAT code is processing the returned codes as ASCII bytes to emulate either (in the two common cases) a Datamedia DM2500 or a DEC VT100 terminal. At times, the characters received are going to be used as binary values for things like cursor positioning.

Is the problem that this interface doesn't use Unicode/utf-8, but something else (ISO8859/1 ?)

Should the MTOUCODE (or MCCS to ...) then be done explicitly in CHAT.TYPEIN, and then let it do the BOUTs as before on those translated codes?  And likewise the reverse.  So a code 0,255 from the keyboard would get translated to 0,137, and then bouted.  CHAT.TYPEIN already has a guard that it will only bout 8 bit codes, otherwise it flashes the window (which is what I was seeing for the other function keys).

And effectively do the reverse sequence the other way around--get the code back, maybe interpret it according to the protocol, but then translate actual characters back to MCCS.

Essentially not rely on the stream's external format--or define a separate external format that does the desired mappings

nbriggs:
There are a few things to consider

- there are CHAT.KEYACTIONS that you can adjust if you want -- standard keyaction entries, which take effect when the chat window has the focus
- characters sent down to the shell should probably be by default ASCII, so shift-6 `^` as 0x5E, shift-minus `_` as 0x5F -- I'm not sure what a non-US terminal (or terminal emulator) would do... that's an extended (and future) investigation
- characters coming up from the shell need to get processed as raw bytes before they get any translation and before they're displayed, and any actions that they imply would be performed then.  This, https://espterm.github.io/docs/VT100%20escape%20codes.html has a description of what you'd expect a VT100 terminal to respond to, so our CHAT architecture should be able to support an emulator that wants to implement them.  Presumably characters that are actually going to be displayed would need (1) translation to MCCS where appropriate, (2) translation to graphic symbols that aren't part of ASCII/ISO-8859-1 when special-graphics mode has been selected.

rmkaplan:
> There are a few things to consider
> 
> * there are CHAT.KEYACTIONS that you can adjust if you want -- standard keyaction entries, which take effect when the chat window has the focus

WRT CHAT.KEYACTIONS, I think they would have to be defined to return the MCCS codes that will translate to whatever the user wants the shell to see, they would then get converted
> * characters sent down to the shell should probably be by default ASCII, so shift-6 `^` as 0x5E, shift-minus `_` as 0x5F -- I'm not sure what a non-US terminal (or terminal emulator) would do... that's an extended (and future) investigation

If the user types Function-10, the keyaction table will produce the MCCS code for underscore 0,255, the translation will take it to 0,137, and reverting to the previous BOUT will just send that code.
> * characters coming up from the shell need to get processed as raw bytes before they get any translation and before they're displayed, and any actions that they imply would be performed then.  This, https://espterm.github.io/docs/VT100%20escape%20codes.html has a description of what you'd expect a VT100 terminal to respond to, so our CHAT architecture should be able to support an emulator that wants to implement them.  Presumably characters that are actually going to be displayed would need (1) translation to MCCS where appropriate, (2) translation to graphic symbols that aren't part of ASCII/ISO-8859-1 when special-graphics mode has been selected.

There is a function TERM.PRINTCHAR in CHATTERMINAL that seems to be the one that puts the character on the display, after all special interpretation has been done.  Putting the inverse translator there produces the MCCS code that the MCCS fonts expect. 

Assuming that the desired 8-bit codes are essentially Unicode charset 0/ISO8859-1, and putting MTOUCODE and UTOMCODE in one or two places seems to make it behave as expected, at least with respect to what shows up in the window with any of the new MCCS fonts:  underscores when you type F-10, circumflex when you type F-6.  Flashes the window when you type shift-6 (uparrow) or shift-hyphen (left arrow).

So this seems like an easy fix...except that back-space doesn't work as I would expect.  The Medley level keyaction table produces code 8 (CHARCODE BS) for the Delete key.  But when I type Delete  in the chat window the \GETKEY in CHAT.TYPEIN returns 127 (CHARCODE DEL).  This doesn't exist in MCCS, so MTOUCODE would convert it to an unassigned Unicode, and the window would flash.  To get it to work I have to brute-force exclude (CHARCODE DEL) from the translation.

But, I'm not sure why the DEL is coming up, if this is the way backspace is supposed to be communicated.  Or some sort of accident.

(One other note:  in release and in this PR, there seems to be an off-by-one bug in the backspace.  It shows up if you backspace over an underscore--the right most pixel remains black.)

rmkaplan:
A second thought re (CHARCODE DEL):  Another solution would be to simply add (CHARCODE DEL) to MCCS, with an identity mapping into Unicode/ISO.  And forget about this glitch.

pamoroso:
I also noticed some minor screen corruption when pressing `Backspace` in certain cases but couldn't reproduce it.

rmkaplan:
Chat with MCCS fonts now works for me, for simple inputs (underscore) and backspace.  I haven't done more elaborate testing.

Added DEL as an MCCS known character

rmkaplan:
One other Chat improvement:  I modernized its window, so you can drag it around by the title and reshape by the corners.

nbriggs:
What do I need to put on CHAT.KEYACTIONS to make shift-6 send ASCII caret and shift-minus send underscore?  There's no way I'm going to be able to change my touch typing to use different keystrokes between a regular terminal window and a CHAT window.

rmkaplan:
Also on Chat, I didn't look at the back-space display glitch.  And  noticed that now and in the release, when it starts up there are some garbage characters on the screen.  This is a snap from the release:
Screenshot 2025-10-11 at 1 43 07 PM

nbriggs:
The garbage is because the shell it's bringing up thinks it's talking to an  xterm-256color, but the CHAT terminal emulation is a dm2500 (by default) - xterm controls that the shell is sending show as garbage on a dm2500 terminal.  I suspect that something tried to pass down to the shell info about the terminal emulator type, but it failed.

rmkaplan:
If you want it to generally switch the bindings of shift-6/f-6 and shift-hyphen/f10 (but not shift-4/f-4), that's in the general keyaction table in LLKEY.

If people think it's better that the characters align with the labels on current keyboards, then we can revert that change.  I think it was discussed somewhere.  You would then have to type f-10 to get the left-arrrow in create expressions--maybe dwim could be fixed to coerce _ to  ←.

rmkaplan:
Is that something to worry about?

nbriggs:
Not generally, just in CHAT, which I think should be doable using the CHAT.KEYACTIONS, right?  I'm fine with using the underscore key to get a left-arrow in an exec (or SEdit), I'm not sure how I feel about it typing file name patterns in a file browser, but when I'm typing at a Unix shell through a CHAT window it's got to be like every other Unix shell.

We should fix the way it tries to tell the shell about the terminal type.  It works better if you have VTCHAT loaded.

nbriggs:
We're actually sending the correct info for the terminal type - it gets set to dm2500 in the default chat (implementation in DMCHAT).  I have a test for LDESHELL in my .zshrc so that I try to set a simpler prompt -- I guess I need to look at what DMCHAT.HANDLECHARACTER is doing with `<esc>[?2004h` --

```
1 - Enter DMCHAT.HANDLECHARACTER:
          CHAR = 27
1 - Exit DMCHAT.HANDLECHARACTER =>
1 - Enter DMCHAT.HANDLECHARACTER:
          CHAR = 91
1 - Exit DMCHAT.HANDLECHARACTER =>
1 - Enter DMCHAT.HANDLECHARACTER:
          CHAR = 63
1 - Exit DMCHAT.HANDLECHARACTER =>
1 - Enter DMCHAT.HANDLECHARACTER:
          CHAR = 50
1 - Exit DMCHAT.HANDLECHARACTER =>
1 - Enter DMCHAT.HANDLECHARACTER:
          CHAR = 48
1 - Exit DMCHAT.HANDLECHARACTER =>
1 - Enter DMCHAT.HANDLECHARACTER:
          CHAR = 48
1 - Exit DMCHAT.HANDLECHARACTER =>
1 - Enter DMCHAT.HANDLECHARACTER:
          CHAR = 52
1 - Exit DMCHAT.HANDLECHARACTER =>
1 - Enter DMCHAT.HANDLECHARACTER:
          CHAR = 104
1 - Exit DMCHAT.HANDLECHARACTER =>
```

nbriggs:
I was wrong about the source of the garbage -- it's the zsh line editor (zle) trying to say that we should "set bracketed paste mode".  CSI here is `<esc>[`, not sure why they used mixed Pm/Ps in the description.
```
CSI ? Pm h
          DEC Private Mode Set (DECSET).
[...]
            Ps = 2 0 0 4  ⇒  Set bracketed paste mode, xterm.
[...]
```
This is what gets you the reverse-highlighted text (and no activation if there's a linefeed) when you paste into a terminal window.

We could fix up the DMCHAT code to parse the set/reset modes, but the easiest thing (on a mac, with zsh) is to
`unset zle_bracketed_paste`, except there are some subtleties (see below)

My ~/.zshrc contains (with a few other things)
```
ssh-add --apple-use-keychain -q
export OSTYPE=$OSTYPE
[ -v LDESHELL ] && source ~/.zshrc-medley
```
and my ~/.zshrc-medley now contains
```
export PROMPT="%n %*> "
zmodload zsh/zle   # load it so we can change parameters
unset zle_bracketed_paste
```

pamoroso:
I tested up to commit ab5c281. Now pressing `F10`, `F6`, and `F4` in a Terminal 10 Chat window yields `_`, `^`, and `$`, respectively. However, the behavior of `Backspace` and `Del` is as described earlier, i.e. deleting characters doesn't work.

rmkaplan:
On the back space, that now works for me (Mac, DM2500 connection to the shell).  (I extended MCCS so that it now includes DEL).

I only have the chat-related files that come in the loadup.  Looking at the dribble file, the loadup includes
    CHAT, DMCHAT, CHATTERMINAL, UNIXCHAT, TEDIT-CHAT (which is unlikely to work).
Nothing more in my init other than to set DEFAULTCHATHOST to NIL.

Can you say/show again what happens when you push the Delete key?

pamoroso:
With DM2500 I get your same behavior for both `Backspace` and `Del`.

With VT100, after typing `1`, `2`, and `3` in a Terminal 10 CHAT window this recording shows what hapens when I press `Backspace` 3 times in succession, then type `A` and press `ENTER`:

https://github.com/user-attachments/assets/195d06ff-26d9-4545-b87d-2e5fb0f7a66a

pamoroso:
A clarification. With DM2500 I get your same behavior only for the default Gacha font but with Terminal 10 `Backspace` still behaves as in the recording. In the recording Medley runs under VNC but the behavior is the same without.

rmkaplan:
Whatever is going on, I don't think it has to do with the font or MCCS.

My bet is that the problem is that DSPFONT for a Chat window doesn't do the right thing.  It appears to just temporarily change the current font on the Window's display stream, but doesn't change the internal data that is supposed to correspond to that.

In the release system, try doing (DSPFONT '(CLASSIC 18) (WHICHW)), typing some stuff, and then backspaces.  You see big characters before the backspace, the screen is garbled after the backspaces.

But if you then type a return to go to the next line and type something, the characters show up in the original Gacha, as if you hadn't done the DSPFONT. 

I think this is old undesirable behavior, not related to MCCS or this round of font changes.

rmkaplan:
If VTCHAT is loaded, I don't see the 2400h.  But it is still a little bit odd.  When I hit the return, it shows an isolated inverted % on the next line, and then it shows an entire prompt string with another % on the line after that.

￼

nbriggs:
You shouldn't be using DSPFONT to change the chat font -- I think setting CHAT.FONT and restarting the CHAT is the only approved way to do it.  DSPFONT is likely almost never the right thing to do unless you're working directly on an image stream (not some application's window)

pamoroso:
When setting `CHAT.FONT` to Terminal 10 VTCHAT works as expected with respect to the function keys for underscore and circumfles, as well as deleting with `Backspace` and `Del`.

rmkaplan:
If you set CHAT.FONT to, say, Classic 18 before you start (e.g. a fresh release sysout), it's still bad.

You type the first character, it looks good.  You type the next character on the same line, the first character reappears after a big run of white space, then the second character.

And backspace is still garbage.

pamoroso:
However, pressing the `_` and `^` keys makes the CHAT window flash with any font under DM2500 and VT100, not display the left and up arrow.

rmkaplan:
> However, pressing the `_` and `^` keys makes the CHAT window flash with any font under DM2500 and VT100, not display the left and up arrow.

Yes, that's to be expected, given the new LLKEY assignments.  Those keys transmit the MCCS uparrow and leftarrow codes (0,255, 0,254), which the shell doesn't recognize.

As per a previous note, it may be more intuitive, especially for new users, to have those transmit according their labels, as they used to.  Users would then have to know that the function keys are available if they want to type those particular arrows (and that they do need to type F-10 when they want the left-arrow in a create expression (unless someone fixes DWIM).

rmkaplan:
BTW, none of these Chat/font/backspace problems are new:  same behavior in the Venue sysout.  Also in the Venue sysout, Tedit mode doesn't work by clicking the TeditMode button in the Chat menu.  Something probably needed to be done earlier to set up the textstream.

nbriggs:
The CHAT.FONT *must* be a monospace font.  It's never going to work with Classic.  The underlying shell may be using what it knows about the terminal emulator to move the cursor before erasing the character, and it can directly address each character cell on the (typically 24 x 80 character) display.  If the emulator didn't put the characters in that grid you're going to get garbage.

rmkaplan:
OK, then it's only Gacha and Terminal up to size 12 that should work, and they (mostly) do.  In playing around, sometimes I still do see the case where the already echoed first-typed character is repeated after some extra white space, before the second typed character appears.

Interestingly, the unwanted black pixel when backspacing over an underscore only appears for Terminal 12. Terminal 8 and 10 are OK, as are all the Gacha sizes.  Maybe something is wrong with its underscore bitmap?

nbriggs:
I'd expect TITAN, LETTERGOTHIC, and BOLDPS to work - they're printwheel fonts and I think they're monospace.  TITANLEGAL perhaps too.  But the character mappings in those may not really be XCCS.

nbriggs:
Actually, they do look like XCCS encoding at least in FontSample for character set C0.  They really don't work well in a CHAT window though, with what looks like mis-mapped characters.

rmkaplan:
I wrote a predicate to test for monopitchness.  The legacy (displayfonts/c0) Terminal 10 has a glitch in the 8-bit region, otherwise Gacha and Terminal are OK.  Titan seems to be OK, Boldps seems to have a lot of variation.

Of course, when slugs (missing characters) get filled in from other fonts to suppress black boxes (e.g. Terminal gets characters from Modern or Classic) there are no guarantees.  But presumably that doesn't happen in the 7-bit Ascii range.

pamoroso:
Part of issue #1157 is that the Calendar application displays some black boxes in the message editor window, so I tried to reproduce the issue on this branch where MCCS is supposed to take care of the boxes.

Now I no longer get black boxes in the message editor but the window content still looks off. And the error is different:

```
END-OF-FILE

In \EOSERROR:
End of file #<IO Tedit Stream/167, 64700>
```

calendar-error

fghalasz:
Running on this PR branch, the following files in docs/medley-irm fail when trying to hardcopy to PDf file:
14-ERRORS.TEDIT
16-SEDIT.TEDIT
25-USERIO-PACKAGES.TEDIT.

All fail with ```Invalid argument: (NIL NIL REGULAR)``` inside \FONTFACE underneath POSTSCRIPT.FONTCREATE where the FONTSPEC argument is ```(SYMBOL 1 (BOLD REGULAR REGULAR) 0 POSTSCRIPT)```

rmkaplan:
@pamoroso The Calendar bug is independent of the fonts.  Calendar fills in its template from a string, using OPENSTRINGSTREAM to first convert the string to a stream before giving it to Tedit (because we changed Tedit a long time ago to treat strings as file names).  OPENSTRINGSTREAM produces a fat (2-byte) stream even if the characters are only in charset 0.  Unlike thin files and thin streams, fat-stream pieces are not binnable at the stream level, so the subr falls out to the textbin code.  And there appears to be an extra SUB1 in that code.  The effect is that you were seeing every other character.

Removing that SUB1 makes it work, but I want to do some more testing to see whether any other piece types somehow depended on that.  And I won't install this fix in the font PR, the code in the TEDIT-STREAM file may already have things that depend on the next hardcopy/imagefile changes that are waiting on the sidelines.

rmkaplan:
@fghalasz , I updated POSTSCRIPTSTREAM (variables were set in the wrong order).  Those 3 files now work for me, but I did see a message flash by in the USERIO hardcopy saying that page 19 was complete full of headings.  The page didn't look odd to me, but I didn't investigate.

pamoroso:
I tested up to commit 016a716 and I have nothing unusual to report. I'm still using this branch as my daily driver.

nbriggs (changes requested):
Could you make it `MONOSPACEFONTP` rather than mixing "fixed pitch" (from typewriter terminology) and "monospace"  (more modern font terminology) in the name?

rmkaplan:
OK

pamoroso:
I updated to commit c727386, still looking good.

rmkaplan:
Forgot to mention:  this PR addresses issues raised in #2178 #2135 #2124 #2120  #2040 #2025  #1971 #1854 #1403

pamoroso:
Can [all the issues](https://github.com/Interlisp/medley/pull/2280#issuecomment-3424214222) addressed by this PR be closed? Any to leave open?

rmkaplan:
@pamoroso , I suspect that all of those issues can be closed, but I think they probably need to be scanned to make sure that there aren't some additional comments that still need to be addressed.

PR #2282, opened 2025-09-08
Workaround added to (HCFILES) to close TEDIT files left open after breaks in OPENTEXTSTREAM.  Temporary fix for HCFiles github action.

fghalasz:
OPENTEXTSTREAM has a bug whereby which the underlying file stream is left open after a break occurs while reading a TEdit file.  In addition, Maiko has some serious deficiencies in its handling of Unix Shell commands.  Together these bugs are causing HCFILES to fail without notice in github actions.  

Fixing OPENTEXTSTREAM and Maiko is the proper solution  But these fixes are way beyond my comfort level.  Issues have been (will be) created for these two bug separately.

Meanwhile, this PR creates a workaround in (HCFILES) that closes the errant open files after an error return from OPENTEXTSTREAM.  This allows the RunHCFILES.yml github action to run successfully.

A run made using this fix is already up at files.interlisp.org.


fghalasz:
Closing unmerged since PR#2284 fixes the HCFILES problem by fixing the root cause.  Hence this workaround is no longer necessary.

Issue #2281, opened 2025-09-08
OPENTEXTSTREAM leaves underlying file stream open after breaks cause premature return (and no textstream object)


fghalasz:
This issue was discovered in the failures we were seeing in HCFILES.  The TEDIT files (from HCFILES) that exhibit the issue are listed below.  For each of these files that I've tried.  OPENTEXTSTREAM breaks under \TEDIT.OPENTEXTSTREAM.PIECES, usually with some font issue.  When you ^ out of the break, the underlying file stream is left open with no easy handle to that file stream since no TEXTSTREAM is returned.

Example:
```
5<- (SETQ S T)
T
6 <- (NLSETQ (SETQ S (OPENTEXTSTREAM "docs/medley-irm/01-INTRO.TEDIT")))
NIL
7<-  S
T
8<- (OPENP)
(...  #<Input Stream on {DSK}<home>frank>il>medley>docs>medley-irm>01-INTRO.TEDIT;1/167,136000> ...)
```

One way to fix this is to wrap some sort of ERRORSET around \TEDIT.OPENTEXTSTREAM.PIECES inside OPENTEXTSTREAM to close the underlying file stream before its local binding disappears.  But TEdit just too complicated to know if this will have negative side effect elsewhere, so I implemented a work around in HCFILES that uses the (OPENP) at the HCFILES level to close the errant file streams.

This issue asks for a proper fix for OPENTEXTSTREAM that closes the underlying open file stream (if any) when OPENTEXTSTREAM errors out.

Problematic TEDIT files from HCFILES:  (Most testing done with files in docs/medley-irm)

docs/dinfo/CHAPBREAKING.TEDIT
notecards/docs/user-guide-v1.2/from_envos/000-OUTLINE.TEDIT
notecards/docs/user-guide-v1.2/from_envos/001-TITLE-PAGE.TEDIT
notecards/docs/user-guide-v1.2/from_envos/002-PREFACE.TEDIT
notecards/docs/user-guide-v1.2/from_envos/003-TOC.TEDIT
notecards/docs/user-guide-v1.2/from_envos/01-INTRODUCTION.TEDIT
notecards/docs/user-guide-v1.2/from_envos/03-SOFTWARE-INSTALLATION.TEDIT
notecards/docs/user-guide-v1.2/from_envos/04-SYSTEM-USE-ISSUES.TEDIT
notecards/docs/user-guide-v1.2/from_envos/05-NOTECARDS-BASICS.TEDIT
[...]

PR #2284, opened 2025-09-09
Make sure TEDIT backing files close on error

rmkaplan:
Addresses #2281.

The files should close on internal errors, and no textstream will be returned.


fghalasz (approved):
LGTM.  Ran doHCFiles.yml in Github Actions as well as do_hcfiles.sh on Ubuntu 24.04 on WSL2.   All runs completed successfully.  Tests showed no extraneous open files during the runs.

pamoroso:
HCFILES now runs to completion but the formatting of the produced prettyprinted Lisp files looks off. For example, there's too much bold text.

rmkaplan:
Is that generally true, or just specific files?

pamoroso:
It's generally true, it happens with all Lisp files. TEdit files look fine though.

PR #2303, opened 2025-09-29
script/do_compiles.sh  to automate compile of all files in sources directory. 

fghalasz:
do_compiles.sh script iterates thru .LCOMs and .DFASLs in sources and compiles each from source in a separate Medley invocation after loading exports.all.  It uses "grep :FAKE-COMPILE-FILE" on the existing LCOM to decide whether to use TCOMPL or FAKE-COMPILE-FILE.

Still very much a WIP.

### DIFF
--- a/scripts/do_compiles.sh
+++ b/scripts/do_compiles.sh
@@ -1,0 +1,191 @@
+#!/bin/sh
+#
+#  do_hcfiles.sh
+#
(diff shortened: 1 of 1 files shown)


fghalasz:
Deleting this PR, but keeping branch for when we get around to returning to the project of compiling the entire system.

PR #2304, opened 2025-09-29
Add function COMPILEFILETYPE

rmkaplan:
(COMPILEFILETYPE CFILE) returns CL:COMPILE-FILE IL:FAKE-COMPILE-FILE or BCOMPL, depending on how the compiled file CFILE was compiled.  NIL if CFILE is not a compiled file.

Reads the filecreated expression.


rmkaplan:
BCOMPL is returned if it was compiled by the byte-compiler in any way:  TCOMPL, RECOMPILE, BCOMP, BRECOMPILE.

This may be useful for #2303

masinter:
isn't this just what COMPILE-FILE? does?  
Perhaps COMPILEFILETYPE is a better name? But having two different entrypoints that don't agree doesn't seem like a good idea.

rmkaplan:
I think COMPILE-FILE? works after a file has been loaded, looks at the property to see what compiler to use now.  This looks at an existing compiled file, tells which compiler was used then.

pamoroso (approved):
Works as expected on Linux Mint 22.1 Cinnamon.

PR #2320, opened 2025-10-21
Rmk131  Mapping MCCS filenames to (Mac?) UTF-8 file names

rmkaplan:
On the Mac, you can type a file name with underscore (Function 10) and it will match a Finder or Unix file name.  You can also use Greek or other MCCS characters in filenames.

This needs to be tested in Windows environments, it may be that different environments use different filename encodings.


rmkaplan:
I fixed Unicode so that it canonicalizes large (non-SMALLP) codes before it translates them to 16 bit codes in the MCCS-unknown range.  It can read and write filenames and file contents with emojis and preserve their identities, for example, even though those emojis will display as black boxes.  The important point is that the same unknown MCCS code will always be assigned to the same large unicode code within the same memory image, so that EQ can be used to compare the values of NTHCHARCODE.

We could go further:  if we know the range of emoji unicodes, we can extend the set of translation tables in Unicode/xerox to provide mappings to codes in some number of currently undefined MCCS character sets (e.g 20Q to 40Q).  Then the translations would be constant across memory images, and we might even be able to import black-and-white images for the emojis from BDF fonts.

rmkaplan:
Has any one tried this on a Windows machine, at least for a file with an underscore in its name?  If that works, I'll remove the draft status.

rmkaplan:
In the worst case the extensions won't work as intended in Windows environments, but it should do no harm wrt past behavior.  I upgraded from Draft.

hjellinek:
Here's the minimal test:

On the Windows side, create a file with an underscore in its name.
In Medley, try to find or open that file, given a name with an (MCCS) underscore in its name.
If it fails, there's a problem.

Do the reverse: create a file with such a name in Medley, see if there's an identically-named file in Windows.

hjellinek:
More liberal test: create a file in Windows with a Greek letter, see if the file appears in Medley with a black box in the name.

In Medley, copy the black box character into another string, create a file with that name, see if the same Greek character appears on the Windows side.

PR #2322, opened 2025-10-21
WHICHKEY returns the keynumber as well as the keyname, for convenience

rmkaplan:


pamoroso:
I'm not sure I see the change. When evaluating `(WHICHKEY)` and pressing `A`, `B`, and `C` I only get the keyname for each key.

whichkey

rmkaplan:
You have to hold down a key.  I added a documentation file.

pamoroso:
I confirm `WHICHKEY` works as described when holding down a key.

PR #2323, opened 2025-10-21
EDITFONT bug fix--compatibility with new FONT conventions

rmkaplan:


rmkaplan:
(The SAVE command was broken, after editing a character's bitmap)

pamoroso:
I tested the change on Linux Mint 22.1 Cinnamon and found no unexpected behavior.

PR #2326, opened 2025-10-22
TEDIT:  Better heuristic estimate of initial region, better management of margin bar

rmkaplan:
For SEBINDS.TEDIT the initial suggested region should be wide enough for the longest line


pamoroso:
This PR does indeed improve TEdit's initial suggested region, here's the default for SEBINDS.TEDIT:

sebinds

rmkaplan:
A small further embellishment: when it looks for a region to recycle, it bounds the maximum and minimum sizes.  So before, if it had looked at SEBINDS, then closed that window and applied Tedit to an ordinary documentation file, the region manager would have recycled the large SEBINDS region.  With this commit it ignores that because it is more than 10% bigger than the heuristic suggestion

pamoroso:
I tested up to 8bce8f4 and the updated heuristics works well. But consider opening with TEdit the ASCII file PLAINTEXT.TEDIT that contains only:

```
Line 1
Line 2
```

Although the TEdit window correctly frames the content it looks cramped and adding more text requires resizing:

tedit-window

What about setting a wider minimum window size if the first few lines are shorter?

rmkaplan:
Should it just always provide as a minimum, plain or formatted, what it defaults to for an empty document?  200 x 100 ?

pamoroso:
That would be fine.

pamoroso:
I updated to commit 97e0147 and the new default for empty documents works well:

tedit-default-window

pamoroso:
The window size heuristics of commit 13e85e7 works as described.

pamoroso:
I updated to commit b2e8fa3 and I haven't noticed anything unusual.

PR #2331, opened 2025-10-28
Extend GITFNS/COMPAREDIRECTORIES so that the See and Compare commands work after files have been rearranged

rmkaplan:
When a file is deleted from the gwc menu, it isn't there to be seen or compared again.  Same thing is true if a file is copied to {UNIX}, the original isn't there.

This changes the CDBROWSER in COMPAREDIRECTORIES so that it associates private copies of the original files when they would otherwise have been lost, and adjusts the compare/see commands so that it uses the private copies if the originals are no longer accessible.  The copies are NODIRCORE files associated with the browser, so they will be collected when the browser is closed.  GITFNS is adjusted to align with the few interface changes.

The saved information is also enough to support undoing of the menu actions, but that interface is left for another time.


pamoroso:
I tested up to commit b589a3f and `PRC` is still working fine.

pamoroso:
I tested up to commit 3f5e6c0, nothing unusual to report.

PR #2335, opened 2025-10-30
Add :IMPORT-FROM option to DEFPACKAGE

MattHeffron:
This resolves #2334

PR #2350, opened 2025-11-02
Add :import-from option to defpackage (import into newly defined package)

### LINKED ISSUE #2334: DEFPACKAGE has :IMPORT but not :IMPORT-FROM
`DEFPACKAGE` can `:IMPORT` actual symbols into the package being defined, _but_ if this is saved on a file in the package created, then the symbols may be written without the original package prefix, and may not specify the correct symbol. 
If `:IMPORT-FROM` were implemented, from standard Common Lisp, then the correct package could be specified, and the symbol names be specified as strings. This would always be written back to the file unambiguously.

masinter:
based on PR #2335 but add PACKAGE value in call to IMPORT -- without it, it modifies *PACKAGE* instead of the package being defined with DEFPACKAGE.

@pamoroso in comments on #2335 I had a false conjecture that the problem had to do with passing symbols vs. strings to name what is being imported.

This resolves #2334


masinter:
now subsumed by #2335

pamoroso:
I confirm that `:IMPORT-FROM` now works fine with `IN-PACKAGE`, also with strings.

PR #2351, opened 2025-11-03
Lispusers package KINETIC:  add CLOSEFN to Kinetic window so that kinetic activity stops and window closes when CLOSEW is called ..

fghalasz:
... and when the Close item on the window menu is clicked.

This makes the Online Demo using Kinetic a little more user friendly.


pamoroso:
This change hasn't already been incorporated in the [example here](https://github.com/orgs/Interlisp/discussions/1097#discussioncomment-14854761), right? Closing the window doesn't interrupt the process.

PR #2352, opened 2025-11-03
Add function (ShellWget URL OUTFILE) that calls the shell to execute 'wget URL -O OUTFILE'

fghalasz:
This PR adds a function (ShellWget URL OUTFILE) that executes ```wget URL -O OUTFILE``` in a separate Unix shell.   URL is a string containing an http(s):// url representing a file on the web.  That file will be downloaded from the web and placed in OUTFILE (which is a versioned {DSK} file).  No format or other checking is done on the URL before it is handed to wget.

ShellWget will ERROR out if ```wget``` is not available on the current machine.

.


pamoroso:
It works well but, if `ShellWget` is called from an Exec, the download pauses for the user to press a key when the page is full. What about running `wget` with the `-q` switch or another option that generates less output?

fghalasz:
Moved to Online repo; renamed ShellWget to ShellHttpGet; moved to file ONLINEUTILS; loaded only in Online Medley image

PR #2356, opened 2025-11-04
Fix SLASHIT bug

### LINKED ISSUE #2355: Infinite loop in SLASHIT("{MEDLEY}<docs>Sun Users Guide>")
this problem appears in master when running HCFILES under MAKE-INDEX-HTMLS but it seems to not be specific to that context -- even happens with SLASHIT("a b c")

SLASHIT is in UNIXUTILS last edit 22-Oct-2025

rmkaplan:
In the case of a file name containing spaces, it was smashing a constant list.

This fixes #2355


masinter (approved):
LGTM

PR #2357, opened 2025-11-05
Add BACKGROUND-YIELD to Lisp.sysout (and thus to FULL.SYSOUT) so you don't need to load it in INITs.

masinter:


masinter:
> Maiko is still pegging my machine at 100% CPU usage.  I understand you've fixed that but I think the fix should be the default and not some special configuration.  IMO, many new users will delete Medley when they see that rather than investigating how to fix it.

This is just an issue of the default configuration ... 


(from @blakemcbride)

pamoroso:
I've started using this branch as my daily driver on Linux Mint 22.1 Cinnamon and I'll report any issues.

nbriggs:
I think the `(SUBRCALL CAUSE-INTERRUPT)` should be removed from BACKGROUND-YIELD before it's made part of the loadup.  It seems to add no value, and definitely interferes with debugging anything that actually does go through the interrupt path (like the ethernet interface)

nbriggs:
It's also unclear to me whether it should be in the LISP.SYSOUT, or only in the FULL.SYSOUT.

pamoroso:
I updated to commit fafb667, still looking good.

pamoroso:
While I still haven't run into anything broken I'm noticing some subtle lags or misfirings in the reactions of the system to some actions I carry out.

For example, sometimes I click on a permanent menu option and nothing happens or the action is not instantaneous. I experience something similar when clicking the NOTECARDS or ROOMS buttons in the apps sysout: it takes a fraction of a second for the system to initiate loading the application, and at times nothing happens after clicking and one or two more clicks are needed for the system to react. Finally, sometimes characters don't show up immediately when typing at an Exec.

blakemcbride:
Interesting. I'm not too surprised, though. If I understand correctly how you're easing the CPU, you're putting in wait states when the CPU is idle. The problem with this is that you'll get the types of delays you're talking about. The fix for the pegged CPU is a kluge. Fiddling with the kluge will never truly solve the problem.

What should happen is that when the system is idle, an infinite wait should occur until an interrupt happens. Keyboard, mouse, and timer (and possibly other) events should cause an interrupt to restart the main loop. This way, there will be no random delays and better reduction of wasted CPU cycles.

masinter:
@pamoroso This PR makes no change to the actual configuration of Medley -- it just loads into FULL.SYSOUT the BACKGROUND-YIELD.LCOM which is also loaded by the default system GREET file (not the user GREET file).   The additional FILESLOAD at GREET is not repeated because the file has already been loaded.  So I would be surprised to see any differental behavior between this PR and the non-PR state.

pamoroso:
From this branch I ran `git pull` to update to commit 4297dd8 but got the following error:

```
paolo@lispmachine:~/medley/medley$ git status
On branch lmm-add-background-to-sysouts
Your branch is up to date with 'origin/lmm-add-background-to-sysouts'.

nothing to commit, working tree clean
paolo@lispmachine:~/medley/medley$ git pull
remote: Enumerating objects: 19, done.
remote: Counting objects: 100% (19/19), done.
remote: Compressing objects: 100% (5/5), done.
remote: Total 11 (delta 9), reused 6 (delta 6), pack-reused 0 (from 0)
Unpacking objects: 100% (11/11), 2.03 KiB | 297.00 KiB/s, done.
From github.com:Interlisp/medley
 + fafb667b...4297dd88 lmm-add-background-to-sysouts -> origin/lmm-add-background-to-sysouts  (forced update)
hint: You have divergent branches and need to specify how to reconcile them.
hint: You can do so by running one of the following commands sometime before
hint: your next pull:
hint: 
hint:   git config pull.rebase false  # merge
hint:   git config pull.rebase true   # rebase
hint:   git config pull.ff only       # fast-forward only
hint: 
hint: You can replace "git config" with "git config --global" to set a default
hint: preference for all repositories. You can also pass --rebase, --no-rebase,
hint: or --ff-only on the command line to override the configured default per
hint: invocation.
fatal: Need to specify how to reconcile divergent branches.
```

Did I screw up my working copy?

masinter:
I did a rebase rather than a merge commit to update the PR. 
But this isn't a good idea (sorry) because of this problem
I _think_ you can patch your local clone by
```
git checkout master
git branch -D lmm-add-background-to-sysouts. # delete old meaning for the PR
git merge origin/lmm-add-background-to-sysouts
```

pamoroso:
Your commands fixed my working tree and I'm now able to pull and build again.

PR #2371, opened 2025-11-14
TEDIT.INSERT ends with a point selection

rmkaplan:
This addresses the problem observed with the INSPHEX display, as reported in discussion #2362.  A misunderstanding of what was supposed to happen when an insert did not come the keyboard.

With respect to INSPHEX, I think the tedit stream should be opened with a fixed-pitch font (e.g. with PROPS  (... FONT (GACHA 12)..), or else (FONT (GACHA 12)) should be passed as the LOOKS argument of TEDIT.INSERT


pamoroso:
I'll change Insphex to open the TEdit stream with the Terminal font.

pamoroso (approved):
This change does address the problem with Insphex reported in discussion #2362.

pamoroso:
I changed Insphex to set the TEdit stream font to Terminal 12.

Issue #2363, opened 2025-11-09
Calling `TEDIT` with title-less window as second arg blocks prompt area


pamoroso:
Pass a window with no title as the second argument of a call to `TEDIT`, like this:

```
(SETQ W (TEDIT NIL (CREATEW)))
```

and swipe out a small window like this:

<img width="273" height="230" alt="Image" src="https://github.com/user-attachments/assets/c085f0c8-9757-4ade-a138-626755eae111" />

Most of the prompt area is blocked by the title bar of the TEdit window. Next, print something into the prompt area:

```
(TEDIT.PROMPTPRINT W "Hi there" T)
```

Now the prompt area blocks most of the TEdit title bar:

<img width="271" height="233" alt="Image" src="https://github.com/user-attachments/assets/28bc60f7-16f0-4b39-b1af-fc742c508ff5" />

If you make the TEdit window large enough both the prompt area and the title bar will show up at full size:

<img width="426" height="343" alt="Image" src="https://github.com/user-attachments/assets/32ab6664-4e0a-4e4b-a2dd-6334c6e9d969" />

This interaction between the prompt area and the title bar may lead to more issues when attaching a menu window to TEdit called with a title-less window. An example is [documented here](https://github.com/Interlisp/online/pull/64#issuecomment-3506322065), where the fix was to set the window title *before* calling `TEDIT`.

This happens on both Linux Mint and Medley Online.

rmkaplan:
I think the problem here is that there are 2 different (incompatible) notions of how to interpret the provided window.  The code that puts on the title considers that the given window should include only the text area, and so the title comes on top of that.  Since the title is usually taken out of the given region of a window, to compensate it grows the window to be enough bigger than the window provided so that the title sits on top of the provided text area.

But the code that attaches the promptwindow considers the given window to already have included space for a title, and attaches to the top of that.  So the title and the promptwindow overlap.

If instead of giving the CREATEW window, you give the value of that window's region, then everything looks right...but the height of the text region is reduced by the space that was reserved for the title.

If from the user's point of view, a Tedit window is a single "object" that always has a title region and a promptwindow region, then a consistent behavior would be for space for the title and promptwindow be subtracted out from a given or swept out region, or the region of a given window.  The text region would always be reduced accordingly, but the space occupied on the screen would be what the user specfied.

[...]

PR #2375, opened 2025-11-15
Fix Tedit promptwindow overlap when given a title-less window

rmkaplan:
This makes sure that the promptwindow is attached as the last thing when a window is set up for a Tedit document.  This also ensures that Tedit windows always have titles.

If a title-less window (as opposed to a region or NIL) is given as the WINDOW argument to TEDIT or OPENTEXTSTREAM,  a title is constructed and installed first, before the prompt window is attached.

This should eliminate the odd behavior reported in #2363.

However, Tedit windows are not guaranteed to be subtractive, in the sense discussed there, so the title and/or promptwindow may show up above the region provided by the user.  There is a conceptual confusion as to whether what the user specifies is the region of the final window constellation (with title and promptwindow) or the region of the text area of the window (the clipping region without title or promptwindow).  The former is subtractive, the latter is additive.

This interacts with the heuristics for suggesting a default size for a window based on properties of the document that it is going to display--those heuristics calculate the text region, assume an additive model.


pamoroso:
In my initial tests the change works as expected. I checked out Insphex, DInfo, NoteCards, and the example in issue #2363.

pamoroso (approved):
It works with no issues in my testing.

PR #2383, opened 2025-11-24
Rmk143  INITVARS command for MEDLEY-INIT-VARS, FLUSHFONTCACHE

rmkaplan:
The loadup scripts are changed to set MEDLEY-INIT-VARS back to NOBIND before loading MEDLEYDIR, and MEDLEYDIR is changed to use an INITVARS command for MEDLEY-INIT-VARS.  Other files can add new entries to MEDLEY-INIT-VARS without that value being sucked up into MEDLEYDIR the next time it is edited.  (Except for variables, like the font cache variables, that appear in files like FONT that are loaded during the make-init phase--the NOBIND would wipe out their values.)

This sets up the font cache variables so that they will be nulled when running in a new system, also adds a more general function FLUSHFONTCACHE for taking specified entries out of the caches (generalization of FLUSHFONTSINCORE).  Documentation in FONTCODECHANGES.TEDIT.

This addresses discussion points raised in #2358 and also the "negative caching" issue raised by @nbriggs in Nov 18 2025 private email:
"and the negative caching is still a problem - I don't see an obvious way to force it to look for the font file again.  It's not fixed by doing a (LOGOUT) and restart -- which it really should, since the sysout may be starting in a different environment."

### DIFF
--- a/.github/ISSUE_TEMPLATE/primer.yml
+++ b/.github/ISSUE_TEMPLATE/primer.yml
@@ -1,0 +1,68 @@
+name: Report an issue with the **Medley Interlisp for the Newcomer** primer
+description: Use this template to report issues or make suggestions.
+title: Title of your issue
+labels: 
+  - primer
+  - documentation
+body:
+  - type: dropdown
+    id: problemType
+    attributes:
+      label: "What type of issue are you reporting?"
+      options:
+        - Suggested improvement
+        - Incorrect explanation / code sample
+        - Confusing explanation
+        - Outdated information
+        - Broken link
+        - Typo / Grammar
+    validations:
+      required: true
+  - type: dropdown
+    id: location
+    attributes:
+      label: "Section of the primer where the issue occurs"
+      options:
+        - Introduction
+        - Medley online and Medley Local
+        - Understanding and Navigating the Interface
+        - Understanding Lisp Syntax
+        - Atoms, Functions and Lists
+        - Variable Bindings and Scope
+        - Iterators and Conditionals
+        - The File Browser
+        - Debugging
+        - Editing functions with SEdit
+        - Build Your First Interactive Program
+        - Saving Your Work
+        - TEdit, The WYSIWYG Editor
+        - Drawing and Displaystreams
+        - Making a Graph with Grapher
+        - Additional Resources
+        - General Feedback (not specific to a section)
+    validations:
+      required: true
+  - type: textarea
+    id: issueLocationDetails
+    attributes:
+      label: "Please provide more details about the location of the issue"
+      description: "For example, the specific page title, section heading, or url."
+    validations:
+      required: false
+  - type: textarea
+    id: issueDescription
+    attributes:
+      label: "Description of the issue"
+      description: "Please provide a detailed description of the issue you encountered."
+    validations:
+      required: true
+  - type: textarea
+    id: suggestedFix
+    attributes:
+      label: "Suggested fix or improvement"
+      description: "If you have a suggestion for how to fix or improve the issue, please provide it here."
+    validations:
+      required: false
+  - type: markdown
+    attributes:
+      value: "## Thank you for helping us improve the **Medley Interlisp for the Newcomer** primer!"


rmkaplan:
As per discussion with @nbriggs , FONTSPECFROMFILENAME no longer defaults to (MEDIUM REGULAR REGULAR) when the filename doesn't have a valid sequence of face characters.  This makes FONTSAVAILABLE consistent with FONTEXISTS/FONTCREATE (at least when font substitutions are not involved).

pamoroso:
It builds and works with no issues on Linux Mint 22.1 Cinnamon.

nbriggs:
@pamoroso - I'd expect it to build and work in the normal case, did you happen to check whether

1. negative caching is flushed over a LOGOUT - ask for a font that doesn't exist, LOGOUT, put a font file in place with the name that didn't previously exist, resume the existing virtualmem, test to see that it now can find the font.
2. that FLUSHFONTCACHE has the same effect as above - nonexistent font request, put a font file in place with the name that didn't previously exist, check that it still doesn't find it, call FLUSHFONTCACHE, check that it DOES now find the font
3. that an otherwise well-formed font without the "-mrr" in the name is neither found by FONTCREATE nor FONTSAVAILABLE, but when the -mrr is added is found (you'd probably have to call FLUSHFONTCACHE, or LOGOUT/restart)
4. that requesting a font that exists, but with a face that doesn't exist but is otherwise valid, can be satisfied as above if the file is put in place, and that this works with files that are currently available as medleyfont format, but where the additional face is NOT in medleyfont format.
5. any other things that affect font file naming - rotation, weight, slope, expansion, are treated correctly when locating the font file.

rmkaplan:
> 4. that requesting a font that exists, but with a face that doesn't exist but is otherwise valid, can be satisfied as above if the file is put in place, and that this works with files that are currently available as medleyfont format, but where the additional face is NOT in medleyfont format.

You probably have to call FLUSHFONTCACHE on this one, because at least for the display there is code to fake up italic/bold from other existing faces.  The faked-face font would be in  the incore cache, as it has always been, and it wouldn't then look for or notice the new file.

pamoroso:
I carried out the following tests.

## 1.

1. `(FONTCREATE '(PAOLO 10))`
2. `(IL:SYSOUT 'FONTS)`
2. `(LOGOUT)`
3. under fonts/medleydisplayfonts, copy MODERN10-MRR.MEDLEYDISPLAYFONT to PAOLO10-MRR.MEDLEYDISPLAYFONTS
3. start a new session with `medley -e -n -v + -y ~/il/FONTS.SYSOUTS`
4. `(IL:FONTCREATE '(IL:PAOLO 10))`

I get this error:

```
FONT NOT FOUND
(PAOLO 10 (MEDIUM REGULAR REGULAR) 0 DISPLAY)
```


## 2.

What should I supply as the first argument of `FLUSHFONTCACHE`? Should the following arguments be `PAOLO` and `10`?


## 3.

1. under fonts/medleydisplayfonts, copy MODERN10-MRR.MEDLEYDISPLAYFONT to PAOLO10.MEDLEYDISPLAYFONTS
2. start Medley

The calls to `FONTCREATE` and `FONTSAVAILABLE`:

```
4_ (FONTCREATE 'PAOLO 10)
In ERROR:
FONT NOT FOUND
(PAOLO 10 (MEDIUM REGULAR REGULAR) 0 DISPLAY)

5_: 
6_ (FONTSAVAILABLE 'PAOLO 10)
NIL
```


3. `(LOGOUT T)`
4. under fonts/medleydisplayfonts, rename PAOLO10.MEDLEYDISPLAYFONTS to PAOLO10-MRR.MEDLEYDISPLAYFONTS
5. start Medley

The calls to `FONTCREATE` and `FONTSAVAILABLE`:

```
4_ (FONTCREATE 'PAOLO 10)
{MODERN10-MRR/135,72420}
5_ (FONTSAVAILABLE 'PAOLO 10)
((PAOLO 10 (MEDIUM REGULAR REGULAR) 0 DISPLAY))
```


## 4.

As noted I'm unsure how to call `FLUSHFONTCACHE`.


## 5.

Some tests:

```
11_ (FONTCREATE 'MODERN 10 'STANDARD)
{MODERN10-MRR/146,24210}
12_ (FONTCREATE 'MODERN 10 'BIR 90)
{MODERN10-BIR/135,72210}
13_ (FONTCREATE 'TIMESROMAND 36)
{TIMESROMAND36-MRR/173,104210}
15_ (FONTCREATE 'TIMESROMAND 36 'BIR)
{TIMESROMAND36-BIR/135,72104}
17_ (FONTCREATE 'TIMESROMAND 36 'BIR 90)
{TIMESROMAND36-BIR/135,72000}
```

rmkaplan:
It appears that hanging the cache resets on MEDLEY-INIT-VARS is not the right thing to do.  Part of the logic there is correct (don't do anything if MEDLEYDIR hasn't changed, presumably on the same machine), but otherwise it assumes that there is a list of strings with old medleydir prefixes that need to be updated to new medleydir prefixes, and that's all it does.  Nothing happens if the old values are not of that form.

I modeled this after the entries for ShellBrowser and ShellOpener on UNIXUTILS, I think those are also probably incorrect (old values won't get set back to NIL).

I think the INITVARS part of this PR is good, MEDLEY-INIT-VARS should not be accumulating entries accidentally.

But another resetting function and another list of variables is needed so that they are completely reset when the MEDLEYDIR changes.

I think I'll put this also in the MEDLEYDIR file, since at least it depends on that change in context.

rmkaplan:
Ahh, it seems the MEDLEY-INIT-VARS has a RESET flag that does what I expected.  So I think it will be sufficient just to had that flag to the variables in FONT and UNIXUTILS.

As of now, the reset feature is not conditioned on a MEDLEYDIR change, so caches will be flushed whenever you restart a sysout, even on the same machine/path.

rmkaplan:
I updated the various files so that the cache variables are RESET.

The caches are flushed whenever a sysout or makesys is started, even in the same environment.  

The TYPE argument to FLUSHFONTCACHE is one of :INCORE :EXISTS :AVAILABLE (or NIL meaning flush all of them).

pamoroso:
I updated to commit b6d5384.

## 2.

1. `(FONTCREATE '(PAOLO 10))`
2. under fonts/medleydisplayfonts, copy MODERN10-MRR.MEDLEYDISPLAYFONT to PAOLO10-MRR.MEDLEYDISPLAYFONTS
3. `(FLUSHFONTCACHE)`
4. `(FONTCREATE '(PAOLO 10))`

I get an error:

```
4_ (FONTCREATE '(PAOLO 10))
In ERROR:
FONT NOT FOUND
(PAOLO 10 (MEDIUM REGULAR REGULAR) 0 DISPLAY)

5_: 
6_ (FLUSHFONTCACHE)
((:INCORE 1) (:EXISTS 0) (:AVAILABLE 0))
7_ (FONTCREATE '(PAOLO 10))
FONT NOT FOUND
(PAOLO 10 (MEDIUM REGULAR REGULAR) 0 DISPLAY)
```

Do I need to start a new Medley session after step 2?

## 4.

I wanted to test with XEROXLOGO 48 whose only file is 
XEROXLOGO48-MRR.MEDLEYDISPLAYFONT but Medley creates `(XEROXLOGO 48 BOLD)` with no issues even if the file doesn't exist. Any suggestions for a font to test?

rmkaplan:
(FLUSHFONTCACHE) with no arguments was interpreting that as the default font for all caches, not as all fonts. Now fixed.   To be more specific try (FLUSHFONTCACHE :EXISTS  '(PAOLO 10)).  You should see (:EXISTS 1) in the result to tell you that it deleted 1 entry.  Then hopefully the test will succeed

pamoroso:
I updated to commit cf6c489.

## 2.

It looks like the test succeeded (same steps as my previous attempt):

```
4_ (FONTCREATE '(PAOLO 10))
In ERROR:
FONT NOT FOUND
(PAOLO 10 (MEDIUM REGULAR REGULAR) 0 DISPLAY)

5_: 
6_ (FLUSHFONTCACHE :EXISTS '(PAOLO 10))
(:EXISTS 1)
7_ (FONTCREATE '(PAOLO 10))
{MODERN10-MRR/135,73420}
```


## 4.

What font can I use for this test? XEROXLOGO 48?

rmkaplan:
@pamoroso , if I understand it, there are 2 separate tests suggested in 4. One is about faces that don't exist and then do exist, the other is about the medleyfontformat vs some other displayfont format. 

I think the format issue is not at stake in this PR.  That's the question, independent of face, as to whether adding a font that was previously known not to exist is then recognized after flushing the caches and adding it as a displayfont instead of a medleyfont.

The face question is more interesting because of the fact that it fakes up missing faces from other existing faces.  So pick a medleyfont that has both an MRR and a BRR.  First copy only the MRR to a new family name, ask for the Bold.    You should get a font with the fake-bold glyphs.  Use EDITFONT to show them on the screen.

Then flush all the caches (including :INCORE), copy the BRR to the new family name, ask for Bold again.  The glyphs in EDITFONT should look better.  (In fact you could copy any other random font to the Bold name, if you really wanted to see the contrast).

pamoroso:
## 4.

Copy MODERN10-MRR.MEDLEYDISPLAYFONT to PAOLO10-MRR.MEDLEYDISPLAYFONT.

```
8_ (EDITFONT (FONTCREATE '(PAOLO 10 BOLD)))
{PAOLO10-BRR/135,73420}
9_ (FLUSHFONTCACHE NIL '(PAOLO 10))
((:INCORE 2) (:EXISTS 2) (:AVAILABLE 0))
```

The resulting font:

font1

Copy HELVETICA12-BRR.MEDLEYDISPLAYFONT to PAOLO10-BRR.MEDLEYDISPLAYFONT.

```
10_ (EDITFONT (FONTCREATE '(PAOLO 10 BOLD)))
{HELVETICA12-BRR/135,73356}
```

The resulting font:

font2

nbriggs:
For the non-existant face, try, for example, BIC (bold italic condensed) which no fonts have, and which shouldn't be derivable from any existing face.

rmkaplan:
Probably wouldn't make a difference:  it coerces nonexistent COMPRESSED to REGULAR, without trying to change the bitmap (unlike faking bold and italic).

rmkaplan:
On @pamoroso 's test, the only thing of note is that the font name for the copied font comes back as HELVETICA12 instead of PAOLO10, so it is ends up with the name that is written in the font as opposed to the user-level name that was requested.  Probably haven't considered before the case of a filename that lies about the font it contains (which the Medley format knows about, not sure about strike and ac). 

So should FONTCREATE always produce a font with the asked-for name (i.e. based on the filename)?  That would make it consistent with FONTSAVAILABLE.  Or should FONTSAVAILABLE look inside the file to see what's really there?

I think this is a separate issue, not related to this PR.

pamoroso:
A test for a non-existant face:

```
5_ (EDITFONT (FONTCREATE '(MODERN 10 BIC)))
{MODERN10-BIC/135,73420}
```

nonex-font

rmkaplan:
As expected

pamoroso:
I updated to commit 857c865, still looking good.

pamoroso:
I updated to commit 85b3cbc, nothing unusual to report.

rmkaplan:
If this is looking good, can it be approved/merged?

PR #2386, opened 2025-11-26
APPS button points to new primer

masinter:


pamoroso:
I'd rename the BASICS button to DOCUMENTATION or DOCS and move PRIMER to the top spot.

PR #2394, opened 2025-11-30
Remove markdown from titles and other minor cleanup

stumbo:
Markdown in titles is not rendered.

Fixes some minor formatting issues I noticed after the initial commit.


pamoroso (dismissed):
Looks good to me.

Issue #2364, opened 2025-11-12
Create "complete" MEDLEYDISPLAYFONT files from Noto


MattHeffron:
Create a set of _complete_ `MEDLEYDISPLAYFONT` files from Noto to parallel those in the upcoming `HTMLSTREAM` module.
I'm looking at @hjellinek 's Web Font project to see if I can use that process to (help) get "complete" fonts from `ttf` to `BDF`, and then to `MEDLEYDISPLAYFONT`.

These comments are copied from PR 2360:
#### From Matt:
I tried this (PR 2630), converting _various_ Noto Sans, Noto Sans Mono, and Noto Serif **ttf** fonts to **bdf** and then to original-style `DISPLAYFONT` files. 
I didn't write `.MEDLEYDISPLAYFONT` files. (@rmkaplan what should I change in the `FONTDESCRIPTOR` objects I create to write _better_ MEDLEYFONT format files _vs._ pre-MEDLEYFONT `FONTDESCRIPTOR`.)
I'm still analyzing just how much coverage of the MCCS glyphs are covered in the Regular, Bold, Bold Italic, and Italic _styles_, and would they be close to **full** fonts.  
E.g., Noto Sans Mono has only Regular and Bold, not Italic nor Bold Italic.

@hjellinek Did you determine a _good_ set of google fonts to get good coverage of the MCCS glyphs? Anyone else
#### From @hjellinek 
Yes, take a look at the "font stacks" (that's Google/CSS terminology) that I defined in my [WebFontMetrics project](https://github.com/hjellinek/WebFontMetrics). The Java source file [Main.java](https://github.com/hjellinek/WebFontMetrics/blob/main/src/main/java/org/interlisp/Main.java) lines 56-70 might help. I'm happy to explain any of it. The companion [FontCoverage](https://github.com/hjellinek/WebFontMetrics/blob/main/src/main/java/org/interlisp/FontCoverage.java) tool examines font stacks for characters present in the code but missing from a font stack.
#### From Matt:
@hjellinek (I have no experience with java...) I've downloaded the WebFontMetrics project from github (as the .zip, and unpacked it).  
I'm trying to follow the README.md, so:
* I installed java (`sudo apt install openjdk-21-jre-headless`),
* connected to the WebFontMetrics directory, 
* created the destination directory (beside the WebFontMetrics directory) and 
* gave the java command as in the README.md (`java -cp WebFontMetrics.main org.interlisp.Main -d ../wfm`)

I get the error:
```
[...]

PR #2395, opened 2025-12-01
Complete the READ-BDF changes for MEDLEYDISPLAYFONT format

### LINKED ISSUE #2313: READ-BDF need to be updated for new MEDLEYDISPLAYFONT format
READ-BDF needs to be updated for font encoding changes and MEDLEYDISPLAYFONT format.

### LINKED ISSUE #2365: READ-BDF should be able to build a single FONTDESCRIPTOR from a sequence of BDF fonts
Extend READ-BDF to construct a single _composite_ font from a sequence of `BDF-FONT` instances, or `.bdf` filenames, or a mix.
Glyphs would be added sequentially as needed to _complete_ the font.
This would be instead of deferring this to the runtime font completion functionality.
The resulting fonts should be saveable with `(MEDLEYFONT.WRITE.FONT ...)`

MattHeffron:
This should complete the changes for the MEDLEYDISPLAYFONTFORMAT.
The building a composite font from BDF files implemented.

Resolves #2313 
Resolves #2365 
This should make #2364 possible.
(I created a size 20 Noto Sans _composite_ font from the same Google fonts as used by @hjellinek's HTMLSTREAM.)

Added files with version of partial implementation of this for old (STRIKE) file format to `obsolete/lispusers/READ-BDF-old/`


rmkaplan:
Can you include one of the medleyfonts that you are now producing?

MattHeffron:
Here are 2 font files, and PDFs of FontSampler of the fonts. (Zipped so github will accept them.)
[bdf derived medleydisplayfonts.zip](https://github.com/user-attachments/files/23861769/bdf.derived.medleydisplayfonts.zip)

rmkaplan:
On the Tedit file:
Paragraph beginning  READ-BDF reads.., 2nd sentence, remove "he first is"
Paragraph beginning "The building of"  "procedes" → "proceeds"
Pargraph beginning ":FAMILY".  FAMILY → ":FAMILY" halfway down.  Also, instead of say a list like FONTCREATE, say a FONTSPEC like FONTCREATE.  (Are you calling \FONT.CHECKARGS to decode the arguments in the standard way?)

In the notes:  it is also unwise to have a hyphen in a family name.

Why do you recommend not putting them in one of the DISPLAYFONTDIRECTORIES?  Why do you think they would be lost if Medley is updated, if they have been committed?  You might clarify that this warning is for the situation when somebody is creating fonts in their own environment, and not committing them to Medley.

rmkaplan:
The only thing I saw in the font files is that the FONTCHARENCODING property should be MCCS, not e.g. NOTOSANS.  Is that coming from \READCHARSET or from your own function?

pamoroso:
I copied the sample font files to medley/fonts/medleydisplayfonts/ and they seem to be working fine in TEdit:

bdf-fonts

masinter (dismissed):
I don't think moving the old versions of READ-BDF to obsolete/ is a good idea.  Old versions are readily available in the git hisstory.

in some cases where we're depricating old software and we're not quite sure of the deprecation it might be OK but that doesn't apply here.

MattHeffron:
> The only thing I saw in the font files is that the FONTCHARENCODING property should be MCCS, not e.g. NOTOSANS. Is that coming from \READCHARSET or from your own function?

It's not in my code. I don't call \READCHARSET.

It looks like it is in FNS `FONTPROP`:
```common-lisp
;; Under (SELECTQ
  (CHARENCODING [OR (ffetch FONTCHARENCODING of FONT)
                    (freplace FONTCHARENCODING of FONT
                       with (if (MEMB (fetch (FONTDESCRIPTOR FONTFAMILY) of FONT)
                                      NSFONTFAMILIES)
                                then 'XCCS$
                              elseif (MEMB (fetch (FONTDESCRIPTOR FONTFAMILY) of FONT)
                                           ALTOFONTFAMILIES)
                                then 'ALTOTEXT
                              else (fetch (FONTDESCRIPTOR FONTFAMILY) of FONT])
``` 
that last else is defaulting to the FAMILY.

MattHeffron:
@rmkaplan 
Made the Tedit changes.
(I'm open to wordsmithing suggestions for the DISPLAYFONTDIRECTORIES note.)

No, I don't use `\FONT.CHECKARGS` because there is an _extra_ intermediate level of defaulting from within the `BDF-FONT` (the XLFD ("X Logical Font Description" within the .bdf file)

MattHeffron:
> I don't think moving the old versions of READ-BDF to obsolete/ is a good idea.

Removed.

pamoroso:
I updated to commit c25da55 and it's still working with no issues.

pamoroso:
I updated to commit d2b1726, still no issues.

MattHeffron:
@rmkaplan
I just made CHARSETENCODING and FONTCHARENCODING values of CHARSETINFO and FONTDESCRIPTOR, respectively, explicitly be MCCS

pamoroso:
I updated to commit 2effafc and the test fonts still work.

MattHeffron:
> I updated to commit [2effafc](https://github.com/Interlisp/medley/commit/2effafc5fbf7a9d4a934d98f64f29ea3f63ae745) and the test fonts still work.

@pamoroso This change wouldn't have affected fonts already created from BDF, but only _newly created_ ones. (And, even at that, the changes were only for _internal correctness_; as shown by the fact that the fonts previously created already worked correctly.)

rmkaplan (approved):
LGTM

PR #2396, opened 2025-12-01
lmm revert causeinterrupt for now

masinter:
- **Add BACKGROUND-YIELD to Lisp.sysout (and thus to FULL.SYSOUT) so you don't need to load it in INITs.**
- **remove CAUSE-INTERRUPT call; doesn't add value**
- **for now, revert change to BACKGROUND-YIELD so that it subrcall CAUSEINTERRUPT**


pamoroso:
This PR does improve the slowdown on Linux Mint 22.1 Cinnamon.

masinter:
@pamoroso not sure what happened, but when I tried it, this PR #2396 didn't cure the slowdown, and I tried again. Please check out #2397.

pamoroso:
PR #2397 works well too.

PR #2398, opened 2025-12-02
TEDIT.ABBREVS entry with LITATOM on CHARACTERNAMES failed to process

MattHeffron:
If `CADR` of a `TEDIT.ABBREVS` entry is `LITATOM` that can `CHARCODE.DECODE` (i.e., it's on `CHARACTERNAMES`), use it.

Clear the TEdit prompt window of any old messages if `\TEDIT.ABBREV.EXPAND` finds an expansion in `TEDIT.ABBREVS`.

`"="` is duplicated as an expansion _KEY_, so modified the `TEDIT.ABBREVS` entry: 
```
("=" "357,121")
```
to
```
("lra" "357,121"  ; U+21C6 Leftwards Arrow Over Rightwards Arrow ("lra" is "left-right-arrow")
                )
```
because `"="` is also the _KEY_ in entry:
```
("=" NONBREAKING-HYPHEN)
```
which was never seen as a possibility (the above was earlier in the `ALIST`). 
(This change exposed the issue with `LITATOM` as `CADR` of entry.)

Cleaned up some comments, typos. Added a few more `TEDIT.ABBREVS` entries.

(Discovered, that if the `CADR` of the entry is valid _itself_, then the `CDDR` of the entry is ignored and can be a comment.)

### Further issue:
@rmkaplan 
Three entries on `TEDIT.ABBREVS` depend on entries on `CHARACTERNAMES` that aren't present in a FULL sysout. They're added in `TEDIT-SCREEN` under 
```common-lisp
[DECLARE%: EVAL@COMPILE DONTCOPY
               (EXPORT etc
``` 
They only get added to `CHARACTERNAMES` by loading `tedit-exports.all`. 
I tried to **move** them out of the above structure to the top of the `TEDIT-SCREENCOMS`:
```common-lisp
((COMS
   ;; This shouldn't be under the DONTCOPY.
   ;; There's no reason for these not to be added to CHARACTERNAMES on any normal LOAD
   (ALISTS (CHARACTERNAMES SOFT-HYPHEN NONBREAKING-HYPHEN NONBREAKING-SPACE)))
               (EXPORT etc
``` 
but was unable to remake the file or compile it.


pamoroso:
On Linux Mint 22.1 Cinnamon it seems to be working fine, event for `b` and `c` that yield the warning `No abbreviation to expand` on master.

abbrevs

I'm not sure about `<<` and `>>` though as the expanded text looks nearly identical to the abbreviation.

MattHeffron:
> I'm not sure about `<<` and `>>` though as the expanded text looks nearly identical to the abbreviation.

@pamoroso
That may be due to the particular font. I tested with Modern 10.
The expansion doesn't look correct. It ought to be smaller and a single character glyph, not 2 characters.

pamoroso:
With Modern 10 `<<` and `>>` expand as expected along with other abbreviations that don't expand with Gacha.

abbrevs-modern

rmkaplan:
It shouldn't depend on the font--Missing Gacha should have been filled in from Terminal and then Modern when the medleyfont was created.

MattHeffron:
Resolved by PR #2439

PR #2399, opened 2025-12-03
Sort the sublists in the value returned by CALLS

rmkaplan:
Sorting the functions and variables in alphabetical order makes it a little bit more useful to find what you might be looking for.

But the main motivation is that sorting would simplify the eventual implementation of a gitfns comparison for compiled files, at least comparing the function signatures, as mentioned in #2314


masinter:
I'd like to try this out before commiting -- isn't the order kind of derived from the order the symbols qppear in the function?  Is it better to lose information? should sorting be an option -- or even a separate function that you could either qpply or not? 

there was some idea of noticingv all symbols used - even in constants. 

another way to comppqre two compiled functions woould be to decompile into bytelqp and compare the results whidh i think woould not pick up neqrly as many differences -- and might be useful for 3 <==> btte atom comparision and transformation.

masinter:
closed by accident, but would like to discuss

Issue #2404, opened 2025-12-04
Too many calls to INFILEP under FINDFILE


rmkaplan:
@nbriggs noticed that there were an inordinate number of calls to INFILEP when looking up a font, and @hjellinek also noticed a lot of time is going into INFILEP under FONTEXISTS?.

The problem seems to be at a lower level, in the implementation of FINDFILE. Below is a trace of FINDFILE and INFILEP to look for a nonexistent file in a single directory.   There are 5 calls to INFILEP, 2 are make no sense (the filename XXX appears also as a subdirectory) and the other 3 all have the same argument--one call should have been enough.  I haven't looked further, but FINDFILE goes to SPELLFILE (even with NSFLG T), which goes to FINDFILE-WITH-EXTENSIONS.  

<img width="392" height="469" alt="Image" src="https://github.com/user-attachments/assets/85d1a976-3e4a-4854-9f90-392dbe6b17ad" />

nbriggs:
> (the filename XXX appears also as a subdirectory)

Back in 2022, you added this comment to `SPELLFILE`:
```
(* ;; "This does the FOO to FOO>FOO correction, with only directory scanning and no other transformations.")
```

rmkaplan:
Yes, that's the problem:

We wanted to group all the Sketch, Tedit, Lafite, Tmax... files together in their individual subdirectories with a consistent naming convention (SKETCH-OPS, TEDIT-STREAM...), so that a search for SKETCH-OPS would also look for SKETCH>SKETCHOPS, parsing at the first (I think) hyphen.

And then we also put main files down inside the same directory, to be in the group, so that the file SKETCH would be SKETCH>SKETCH.  So, any lookup that failed for a file FOO without a hyphen would look for FOO>FOO, and any lookup that failed for FOO-FIE would then look for FOO>FOO-FIE.  

The double lookup isn't so bad (it's typically only the entry-point file (TEDIT, SKETCH) that has a double lookup, the hyphen names for all the subfiles are built-in the FILES commands.

But this is compounded for fonts, because of the unrelated convention that the fields of a fontfile name are also separated by hyphens.

We could do a little reworking (e.g. put SKETCH at the top-level with all of its subfiles in a parallel directory SKETCHFILES>) and maybe change the subfile convention to use characters other than hyphens so that fonts don't fall into this.

[...]

PR #2405, opened 2025-12-04
FONT: eliminate hyphen INFILEPs for font lookup

rmkaplan:
\FINDFONTFILE does its own directory searching, avoiding the hyphen-processing of FINDFILE, as discussed in #2404 .

Looking up a non-existent display font with 2 directories and 2 extensions does 8 INFILEP's, 4 for the directory/extension combination times 2 for the charset/no-charset.

Before it did 32, because of parsing the family off as also a subdirectory.

Other devices, with single directories and single extensions (HTML?), should only do 2.

With more thought, we could also build in a little more knowledge (another FONTDEVICEPROP) about whether it is worth probing for the charset subdirectories c0>.


pamoroso:
On Linux Mint 22.1 Cinnamon tracing `INFILEP` does indeed yield 8 calls for the test case of creating a nonexistent font.

rmkaplan:
If all the information in the Interpress and postscript font-files is instantiated when the charsets are read into font descriptors, then we can convert them all to equivalent medleyfont format files and systematically eliminate the extra factor of 2 that comes from also searching charset subdirectories.

At this point probably we can also set the default DISPLAYFONTDIRECTORIES and DISPLAYFONTEXTENSIONS in MEDLEY-INIT-VARS to point only to the medleyfont directory/extension.  Anybody who wants to look at files in the old directories/formats can adjust their private values for these parameters.

In the end , with default settings there should be only one INFILEP the first time there is a probe for a given font (existing or not) on any device--the caches would record the result for later probes.

But this PR does not depend on those future updates.

rmkaplan:
Updated SPREADFONTSPEC macro: @hjellinek noted that it should use the FONTSPEC record rather than popping down the list

pamoroso:
I updated to commit 707a415, still looking good.

PR #2406, opened 2025-12-05
FontSampler sample sheet display in column major order.

### LINKED ISSUE #2273: Fontsampler in column major order ?
It would be nice if FontSampler would print out its tables in column-major order, like the change I made to EDITFONT.  That would make it easier to compare with the layout of XCCS and Unicode tables.

Also, the documentation doesn't say explicitly how to produce a table for viewing on the screen, not on a printer.

I discovered that you can give a file name for the printer argument (e.g. FOO.PDF) and that will secretly produce a FOO.PDF pdf file.

The value of the final CLOSEF in FontSample should be returned.  But in the case of having given a FOO.PDF argument for the printer, the value returns is a ps file on tmp, not the pdf file that was created (by the call to OPENIMAGESTREAM I suspect).

MattHeffron:
FontSampler sample sheet display in column major order.
Added alternative CharacterSets designations.
Updated documentation.

Resolves #2273


pamoroso:
I tested this with the output sent to a window and it works, but FontSampler doesn't seem to handle window repaint and reshape. Is this expected?

MattHeffron:
> FontSampler doesn't seem to handle window repaint and reshape. Is this expected?

Yes, this is expected. It's just treating the window as a printer. One and done.

pamoroso:
When I call `FontSample` with `'DISPLAY` as the printer and `NIL` as the stream type the output goes to the primary output and the sample sheet is not shown in a window:

fontsampler1

To actually show the sample sheet in a window I need to pass `T` as the printer and `'DISPLAY` as the stream type:

fontsampler2

MattHeffron:
@pamoroso I fixed the documentation.

rmkaplan:
When I call it with
(FontSample '(HELVETICA 10) 0 NIL 'DISPLAY)) 
I get a blank window.

When I replace the NIL with T, I have gotten misaligned pointer errors and hard crashes

pamoroso:
With commit bed9819 the documentation is now correct and clearer.

PR #2415, opened 2025-12-12
Add to `DEFINE-RECORD` expansion providing of arglist info for generated macros.

MattHeffron:
**NOTE:** This uses the function `IL:CLSMARTEN` which is from the file `CLSMARTARGS`. 
The file `CLSMARTARGS` isn't loaded until almost immediately **after** `XCL-EXTRAS` during making loadups.
There are no uses of `DEFINE-RECORD` in making the `lisp.sysout` loadup, so this _ought_ to be safe, but this must be verified!


pamoroso:
On this branch Medley builds and runs with no apparent issues on Linux Mint 22.1 Cinnamon.

masinter:
I'd like a little background on this PR -- what problem is it solving? 
What is DEFINE-RECORDS? Not fmailiar with it.

MattHeffron:
> @masinter  What is DEFINE-RECORDS?

`xcl:define-record` is a DEFINER that "Creates a structures object named by the symbol name that provides Common Lisp accessors, settors, predicates and constructors for an Interlisp record" 
(from the [Medley Release Notes, September 1988](https://files.interlisp.org/medley/docs/ReleaseNote/SEC7-CLIMPLMNTN.TEDIT.pdf), Page 7-3 ).

I'm trying these to reduce the Interlisp (CLISP) code in READ-BDF, which already is primarily Common Lisp code. 
I discovered that the macros are defined in such a way that their argument lists are not available to `?=`, _etc._ 
This change just adds the argument list information to the property lists of the defined macros.

It isn't _actually_ a _structures object_ created, but macros and SETF places, that can be used _as if_ the record was defined by `defstruct`.

pamoroso:
I updated to commit 0da9bac and `DEFINE-RECORD` still works with no apparent issues.

rmkaplan (approved):
Looks good--12/29/2025 meeting

Issue #2413, opened 2025-12-11
\ETHEREVENTFN exits too early when no ethernet detected


nbriggs:
The `\ETHEREVENTFN` is responsible for cleaning up in a sysout that was created on an ethernet enabled host when it is restarted on another where the ethernet is NOT enabled -- but in this case it exits before clearing the ethernet addresses and perhaps other state left in the sysout.  It should at least do `(\SETLOCALNSNUMBERS)` (which will give the host the generic 0#0.12345.0# address), and quite likely '(\SETETHERFLAGS)` and `(\FLUSHNDBS EVENT)` before its early exit.

masinter:
The assumptions about the network don't match current capabilities. There's code in Loops, for example, that uses the Ethernet ID as a unique ID address -- kind of assuming that
* every machine has a single EThernet unique ID : there's only one per machine but every machine has one. There are other assumptions that are clearly not valid.

Should the ethernet address be cleared when restarting without network available? Or treat it as a per-VM value which can be used persistently even when the VM might be started sand saved even when ethernet isn't available?

nbriggs:
I agree that the assumptions don't match the current capabilities, and I'm working to resolve some of that.

It looks as though back in 1985 BvM changed the code in LLETHER `\SETLOCALNSNUMBERS` so that if a sysout is started and there is no ethernet (hardware says 48 bits of 0) then it sets its NS address to `0#0.12345.(serialnumber)#` apps don't see a zero address, and the addresses are unlikely to be duplicated in the local environment.  Otherwise, it ensures that the Lisp ethernet address related variables get reset from the hardware address that the microcode/emulator provides, so that you don't end up with a mismatch between the hardware and the variables.

The problem (for Loops) is that the current code exits in the no-ethernet case (i.e., always) *before* calling `\SETLOCALNSNUMBERS` so you end up with `\MY.NSADDRESS` and `\MY.NSHOSTNUMBER` set to zero.  Any hacks that were done to fix that in Loops were done in the wrong place.

With the advent of more easily set up ethernet access -  if the sysout was created with ethernet enabled and then it's started with no network, it has someone else's hardware address, which is bad.

But then we have other issues to deal with:

(1a) With people running multiple medley instances on the same machine - even with the fix, there's only one serial number - though if they're not running with ethernet it _might_ not matter, but I don't know where Loops might embed the address in objects that could be shared in other ways.

[...]

PR #2418, opened 2025-12-17
Updates \ETHEREVENTFN to relocate exit when the ethernet is unvailable.

nbriggs:
In the case that the ethernet is unavailable the \ETHEREVENTFN should still call \SETETHERFLAGS, \SETLOCALNSNUMBERS, and \FLUSHNDBS, or a newly started image will retain information from the system on which it was built (such as ethernet MAC address in the NS address).  This will also ensure that the NS address is not set to 0, instead initializing with a fake address based on the system serial number (calls \SERIALNUMBER).


nbriggs:
Fixes issue #2413

pamoroso:
I'm running this branch on Linux Mint 22.1 Cinnamon and so far I haven't noticed anything unusual.

nbriggs:
Check \MY.NSADDRESS with the new and old Maiko in fresh sysouts, they should be different.

pamoroso:
What new Maiko? Isn't this change limited to Medley?

nbriggs:
Sorry, yeah, I haven't pushed that.  You should still check \MY.NSADDRESS with the new Lisp code though.

pamoroso:
Under Medley/nhb-fix-ethereventfn and Maiko/master:

```
4_ \MY.NSADDRESS
0#0.12345.401#
```

nbriggs:
Perfect, thanks.  That's what I'd hoped we'd see.  The .12345. is constant for a non-ethernet Medley, the 401 is coming from the "host serial number" (low 16 bits of the result of gethostid()).

nbriggs:
This PR should wait for an additional change -- `\FIND.LOCALPUPHOSTNUMBER` should not call `\LOOKUPPUPNUMBER` if there's no ethernet available.  That is, starting with
```
          [COND
             [(AND (\ETHER-AVAILABLE)
                   (SETQ NEWNUMBER (\LOOKUPPUPNUMBER \MY.NSHOSTNUMBER NDB)))
          ...
```
I'd just make the change to the branch, but I'm having trouble getting a clean update to PUP.

MattHeffron:
> This PR should wait for an additional change

Probably should convert this PR to "Draft" until it's ready...

pamoroso:
When switching to this branch from rmk126--HARDCOPY-to-printers I get this error:

```
paolo@lispmachine:~/medley/medley$ gh pr checkout 2418
Switched to branch 'nhb-fix-ethereventfn'
Your branch and 'origin/nhb-fix-ethereventfn' have diverged,
and have 1 and 48 different commits each, respectively.
  (use "git pull" if you want to integrate the remote branch with yours)
Checkout complete!
hint: Diverging branches can't be fast-forwarded, you need to either:
hint: 
hint: 	git merge --no-ff
hint: 
hint: or:
hint: 
hint: 	git rebase
hint: 
hint: Disable this message with "git config advice.diverging false"
fatal: Not possible to fast-forward, aborting.
failed to run git: exit status 128
```

How should I proceed?

nbriggs:
I force pushed an update so change to another branch, delete the local copy of this branch, then check it out again

pamoroso:
I eventually checked out the branch correctly and the new features seem to work:

```
4_ (\ETHER-AVAILABLE)
NIL
5_ \MY.NSADDRESS
0#0.12345.401#
6_ \MY.NSHOSTNUMBER
(NSHOSTNUMBER 0 5349 257)
```

Are these the expected values?

nbriggs:
The last 16 bits aren't what I expected, but they're good anyway.  It fills it in with `(\SERIALNUMBER)` which is derived from the Unix hostid, which on macOS always ends up 0, but apparently Linux has a more useful value that's hardware dependent, so that's good.

pamoroso:
I updated to commit d79f1ae and it's still working as expected.

nbriggs:
Summary of changes --

      Updates \ETHEREVENTFN to remove early exit when the ethernet is unvailable.
      When no ethernet \FIND.LOCALPUPHOSTNUMBER should immediately return NIL
      Updates \SETETHERFLAGS to avoid setting \10MBFLG when no ethernet
      Avoids making copy of \10MBLOCALNDB as \MAIKO.10MB.NDB for \MAIKO.ETHER-INTERRUPT
      RESTART.ETHER at end of loadup-lisp to ensure network structures are set up.
      Removes OCTALSTRING from PUP - now in APRINT

pamoroso:
Nothing unusual to report at commit c160b0a.

rmkaplan:
I loaded this up on my mac. The first thing I tried was

DIR {FOO} for FOO an unknown host.  Previously, this instantly caused an error:  file not found {FOO}*.*;*

In this loadup it gives the same error, but it takes a few seconds.  Is the delay expected/intended?

rmkaplan:
Actually, I think the reason I didn't see the delay in an old loadup is that there I had \LEAF.DEVICEP defined as NILL.  Without that I see the same delay, which appears to be under EXCHANGEPUPS under ETHERPORT.

Are there/will there ever be any leaf servers?

nbriggs:
Someone has an IFS running on a Raspberry Pi.  I thought I'd guarded all the paths into EXCHANGEPUPS, there must be one I've missed.  It shouldn't be waiting trying to EXCHANGEPUPS when there's no ethernet to exchange them on.

nbriggs:
It's `\LOOKUPPORT` which doesn't ensure that PUP is ready before trying and timing out.  It needs to be the test for PUP being ready rather than just whether the ethernet is available because we don't want it to take the timeout when you've got only XNS running.  I've pushed the fix.

pamoroso:
At commit fb8e46f `DIR {FOO}` instantly yields an error on Linux Mint 22.1 Cinnamon but I get a delay on Medley Online.

rmkaplan:
I confirm:  also no delay on Mac

nbriggs:
@pamoroso that tracks -- online is neither running the most recent Medley code, nor Ron's `NILL` replacing `\LEAF.DEVICEP` change.

rmkaplan (approved):
LGTM

PR #2430, opened 2025-12-29
Fix FILETYPE property of HLDISPLAY

rmkaplan:
It was COMPILE-FILE, but needed to be :FAKE-COMPILE-FILE


pamoroso:
I don't know how to test it but this branch builds and works with no apparent issues on Linux Mint 22.1 Cinnamon.

rmkaplan:
This was a compile-time issue.  It was attempting to create a DFASL rather than an LCOM, and failing.

nbriggs (approved):
Looks like the right change.

Issue #2353, opened 2025-11-03
DIR and FILEBROWSER columns ?


rmkaplan:
Some odd things:

DIR * LENGTH
   shows a column of lengths

DIR * LE
   shows a column of lengths, plus a BYTESIZE column that has (     ) on every line.

Shouldn't those be the same?  And, does BYTESIZE make any sense any more?

For the FB command, the default presentation has columns
  Name Length  Pages Created Author

Would a modern user know (or care) how pages are measured?  A KB column would be more useful, albeit redundant with Lenght.  And, isn't the Author now always the current user?

nbriggs:
Author is the unix user that created that file - if you're looking at a directory created by another user of  the machine then it would be their login name.  There are certainly files created by "root" that you might look at.

I think there may be issues with the filing code returning the correct attributes when looking at `{unix}` files.

masinter:
I like the idea of modernizing FB, but I'm not sure it's clear what the 'right' thing to do is.

I think fixing this is better done by individual users' INIT. Whether the OS can return meaningful creator IDs for example probably depends on the backend.

Every change introduces the possibility of destablization. Obvious bugs and errors and completely useless stuff are legit targets, and I wanted FB defaults to change. 
It might depend on the file system mounted.

If it were up to me I'd mark this as WONTFIX.

rmkaplan:
Maybe author is useful

Generally, I think default displays should at least make sense to new users, the ones that wouldn't yet know how to fix their INIT files, and changing default displays shouldn't be destabilizing.

LEN ought to be an abbreviation for LENGTH, but it isn't.

rmkaplan:
Some other changes/enhancements to the file browser were discussed at the 11/5/25 meeting.

The question was raised whether the submenu options under the See menu  (fast see pretty, fast see unscrollable, scrollable & pretty) should be removed.  The default is scrollable & pretty (using readonly Tedit or a browser pdf window), and that is fast enough now that there is no need for the less useful hidden under the main menu.  It is also odd that there is a "file browse" submenu item since that's redundant with the behavior that you get when you "See" a directory.

There was also a general consensus that it would be useful to be able to move back in the directory hierarchy, not just down, and to have an option for reusing the current window for browsing a directory rather than opening up a new window for each directory you look at.  

[...]

PR #2431, opened 2025-12-29
#2353Remove vacuous (BYTESIZE) from DIRECTORY LENGTH column

rmkaplan:
As mentioned in #2353, it's silly to have the BYTESIZE always there for
DIR -- LENGTH

#2353 has a much longer discussion of filebrowser column defaults, that's not part of this.


pamoroso:
With this change the File Browser works with no apparent issues on Linux Mint 22.1 Cinnamon.

PR #2432, opened 2025-12-30
Recompile \INITDISPLAYBCPL in ATERM

rmkaplan:
This has been hanging around for about a year, I don't remember exactly why I did this.  But probably to make sure that the (create STREAM) tracks changes that had been made to the STREAM declaration quite some time ago, for external formats


pamoroso:
It builds and runs with no apparent issues.

nbriggs (approved):
Looks good.

PR #2439, opened 2026-01-03
Rmk150  update to tedit.abbrevs

rmkaplan:
This includes and should supercede the changes in #2398 (please test).

But the major improvement is in the algorithm for matching the abbreviations against the text.  If you have a point or 1-character selection, it searches backwards for the substring that matches the longest abbreviation key.  So for example, if -- is expands to an n-dash and --- expands to an m-dash, a point selection after --- will get an m-dash.

If you have a multi-character selection, it only tries to match the selected characters.  So if you really want the last 2 characters of --- to be an n-dash, you would select just those 2 and then type ^X.

In the multicharacter case, if it can't find a match in the abbreviations list, it tries to interpret the selection as a character name.  So selecting INFINITY will give you the infinity character.

A shortcoming of this version is that it does literal case-matches only.  It doesn't first try for a literal match and then back off to a case-insensitive match.  This is a simplified implementation using MULTI-ALIST to build a digital search tree.  The easiest way to get case-insensitive matching would be to upgrade the slow case of MULTI-ALIST to take a functional test argument, like commonlisp ASSOC (@hjellinek asked why it wasn't already doing that).  I'll fiddle that first, then do another update for this.


pamoroso:
On Linux Mint 22.1 Cinnamon expanding `INFINITY` with Modern 10 in TEdit yields the following error:

```
Help! Shouldn't happen:
```

abbrev

The backtrace:

```
TEdit/5(debug)BTV
   MESS1 "Shouldn't happen:"
   MESS2 NIL
HELP
   MESS NIL
SHOULDNT
   TSTREAM #<IO Tedit Stream/161,21700>
   SEL {SEL: 12-8-L _ 100/48946}
   LASTCHAR 19
   FIRSTCHAR 12
   ABBREV NIL
   STRING NIL
\TEDIT.ABBREV.PARSE
   TSTREAM #<IO Tedit Stream/161,21700>
   TEXTOBJ {TEXTOBJ}#121,62600
   SEL {SEL: 12-8-L _ 100/48946}
\TEDIT.ABBREV.EXPAND
   TSTREAM #<IO Tedit Stream/161,21700>
   CHARCODE 24
   TEXTOBJ {TEXTOBJ}#121,62600
   FN \TEDIT.ABBREV.EXPAND
\TEDIT.COMMAND.FUNCTION?
   SI::*DUMMY-FOR-CATCH* T
   SI::*CATCH-RETURN-FROM* (&)
   CHARCODE 24
   TCH NIL
   FN NIL
\TEDIT.COMMAND.LOOPA0001A0002
   SI::NLSETQ-VALUE NIL
   *PROCEED-CASES* (&)
   SI::*NLSETQFLAG* NIL
SI::*UNWIND-PROTECT*
   TSTREAM #<IO Tedit Stream/161,21700>
   TEXTOBJ {TEXTOBJ}#121,62600
   LISPXHIST NIL
   SI::*RESETFORMS* ((&))
   RESETSTATE NIL
\TEDIT.COMMAND.LOOP
   TSTREAM #<IO Tedit Stream/161,21700>
   WINDOW {WINDOW}#121,55664
   UNSPAWNED NIL
   TEXTOBJ {TEXTOBJ}#121,62600
\TEDIT1
   *FORM* (\TEDIT1 (QUOTE #<IO Tedit Stream/161,21700>) 
{WINDOW}#121,55664 NIL)
   *ARGVAL* NIL
   *TAIL* NIL
   *FN* \TEDIT1
\EVALFORM
   %#FORM# (\TEDIT1 (QUOTE #<IO Tedit Stream/161,21700>)
 {WINDOW}#121,55664 NIL)
   *CURRENT-PROCESS* #<Process TEdit/173,120000>
   HELPFLAG BREAK!
   \CURRENTDISPLAYLINE 0
   \#DISPLAYLINES 13
   \LINEBUF.OFD #<IO Linebuffer Stream/161,21400>
   *READTABLE* #<ReadTable INTERLISP/173,153714>
   \PRIMTERMTABLE {TERMTABLEP}#173,146740
   \PRIMTERMSA {CHARTABLE}#173,147000
   TtyDisplayStream #<Output Display Stream/161,104600>
   SI::*RESETFORMS* NIL
   \INTERRUPTABLE T
   \TTYWINDOW NIL
   READBUF NIL
   \TERM.OFD #<Output Display Stream/160,120100>
   *STANDARD-OUTPUT* #<Output Display Stream/160,120100>
   *STANDARD-INPUT* #<IO Linebuffer Stream/161,21400>
\MAKE.PROCESS0
T
```

I get the same error when expanding `<<`, `>>`, `lra`, `!`, `?`, and `u`.

rmkaplan:
Hmm, they all work for me.  Can you load TEDIT-EXPORTS.ALL, break \TEDIT.ABBREV.PARSE, and send me a screen shot of (INSPECT SEL)?

pamoroso:
I loaded TEDIT-EXPORTS.ALL, broke `\TEDIT.ABBREV.PARSE`, and expanded `INFINITY` in TEdit. This is the output of `(INSPECT SEL)`:

inspect-sel

rmkaplan:
Thanks, I see the problem.  Your selection is at the paragraph level (middle button in the margin), not the character/word/line level.  That is, the entire paragraph is the string INFINITY.

I had forgotten about that possibility.  I'll do an upate later today.

rmkaplan:
The parsing algorithm is simpler, and the paragraph and line selection should now work better.  When you select a line or paragraph, if the caret is at the end, it looks at the string before the terminating EOL.

This also does the case-insensitive match if the case-sensitive match fails, and does the character-name match if both case sensitive and insensitive matches against the abbreviation table have failed.

pamoroso:
Commit 8573b94 fixed the issue and I can now enter with no errors the above abbreviations, also with paragraph selection.

rmkaplan:
I implemented a convention for explicit marking of the string that is to be matched and replaced.  If the point is just after a \, then the matching string is everything between that and the nearest preceding \.  So, if you are typing along, you can type /Greek,a/^x to the the slashes and everything between replaced by alpha.  Otherwise, you would have to stop and explicitly select a Greek,a substring to get the alpha replacement.  This also gives control when you want the expansion of a shorter strings that is also a suffix of a longer string.  So if you type ---^x you would get an em-dash, but if you want a hyphen followed by an en-dash, you can type -\--\^x.  Finally, an initial back-slash will be consumed as part of the longest match, even if there is no final backslash, effectively following the pattern of Latex.  You can type in^x or \in^x to get the membership symbol.

pamoroso:
I updated to commit 2a59880 and backslash expansion works as intended except for a few cases in which I'm not sure the output is expected:

* `\Greek,b\^X` yields `\Greek,` followed by a bullet
* `\Greek,c\^X` yields `\Greek,` followed by a copyright symbol
* `---^X` yields an em dash
* `-\--\^X` yields an en dash

backslash-abbrevs

pamoroso:
I updated to commit 6bea84e and now:

* `\Greek,b\^X` yields `β`
* `\Greek,c\^X` yields a black box
* `---^X` yields an em dash
* `-\--\^X` yields an en dash

tedit-abbrevs

rmkaplan:
I think those are now the correct results.  Greek,c doesn't exist in XCCS/MCCS, but the charname decoder (CHARSET.DECODE) only knows how to combine code values of the character set name and character name without any knowledge of whether the resulting number means anything.  So it returns the code that would correspond to a c in the Greek character set, even thought there is no actual character assigned to that code position.  The result of the expansion is a slug black box.

MattHeffron:
In `TEDIT-RELEASENOTES.TEDIT`, I think that the paragraph referring to _name of a character_ ought to explicitly state that it uses `CHARCODE.DECODE` (which looks on `CHARACTERNAMES`). This informs users _where to look_ for potential abbreviations.
Also, you mention this as with a "multi-character selection", but it might be good to be explicit that this doesn't work with a point selection.
The `\abbrev\` pattern paragraph _could_ reference the earlier paragraph: `If a multi-character selection...`.

In `TEDIT-ABBREV` a simple space is the _abbreviation string_ for non-breaking space. (This is unchanged.) I think that it might be a good idea to **add** a _text_ string for that expansion (e.g., "nbsp"). With the simple space, when expanded, there's no visible indication that it was changed to non-breaking. Using some text, then something visibly changes, reducing potential confusion.

MattHeffron:
`TEDIT-ABBREV` doesn't clear the `TEDIT.PROMPTPRINT` if it is successful. 

- Try to expand misspelled abbrev string.
- `TEDIT.PROMPTPRINT` shows "No abbreviation to expand"
- Redo expansion with fixed abbrev string.
- Expansion succeeds, but message remains.

rmkaplan:
Responded to @MattHeffron 's comments.  The prompt message on success now tries to say more explicitly what happened.

MattHeffron:
Sorry ... In `TEDIT-RELEASENOTES.TEDIT` the `This document was last edited in October 2025` should be `January 2026`

MattHeffron (approved):

pamoroso:
I confirm the updates up to commit 74c41b7 fix the prompt window issue.

PR #2440, opened 2026-01-05
Fix MANAGER Warnings issued at load time.

MattHeffron:
Manager issued the warning:
```
Warning: The variable Manager.ACTIVEFLG was unknown and has been declared SPECIAL.
``` 
11 times when it was loaded as DFASL.

This is from the 11 functions that were ADVISED with references to `Manager.ACTIVEFLG` in the advice expression.
The `SPECVARS` and `GLOBALVARS` file package commands are expanded under `(DECLARE%: DOEVAL@COMPILE DONTCOPY` so they aren't set when the DFASL file is loaded. (They're not even present in the DFASL file, except as part of the COMS.) 
Even though `Manager.ACTIVEFLG` was in a `SPECVARS` file package command, this wasn't in effect when the advice was being applied. 
Ultimately, declarations were required in each of the advice expressions.


hjellinek (approved):
I loaded the new dfasl and didn't see any warnings.  MANAGER seemed to work as before.

pamoroso (approved):
I confirm this change makes the warnings disappear and MANAGER still works.

PR #2450, opened 2026-01-21
Rmk150  tedit and sketch doc files

rmkaplan:
Added the scanned pdfs to library/tedit/docs/ and library/sketch/docs, as per 1-21-2026 meeting


pamoroso:
Shouldn't the file extensions be all uppercase like `.PDF`?

rmkaplan:
My reading of HCFILES is that it upper-cases before it does any extension comparison.

pamoroso:
The new manuals haven't shown up yet at file.interlisp.org, so I wondered whether the extension had anything to do with it. But perhaps the HCFILES automation hasn't run yet since the meerge of the PR.

MattHeffron:
I manually ran the HCFILES action and the files are there now.

PR #2454, opened 2026-01-23
index.html's exsternal-format is UTF-8

masinter:
set :exsternal-format of index-html to be UTF-8


rmkaplan:
Why does it need the advice?

masinter:
@rmkaplan 
> why does it need the advice?

at one point, errors were being reported in the prompt window and not included in the dribble file.

MattHeffron:
I manually ran the HCFILES action and the [index.html](https://files.interlisp.org/medley/) appears correct.

PR #2462, opened 2026-01-27
Rmk151  Remove old fontfile conventions from \FONTFILENAME

rmkaplan:
Some minor FONT cleanups:

No need to maintain code for *OLD-FONT-EXTENSIONS* and *USEOLDFONTDIRECTORIES*.

Fixed it so \FONTFILENAME can take a FONTSPEC as its FAMILY argument, also calls FONTFACETOATOM instead of doing is own version.

Also fixed MEDLEYFONT.FILENAME to call \FONTFILENAME and not try to replicate the filename logic.

Added AVGCHARWIDTH as a FONTPROP property.

If IMAGESTREAMTYPES doesn't have a FONTCREATE entry, \CREATEFONT defaults to just creating a fontdescriptor from the fontspec (generic behavior should be good for most devices--CHARSET function would need to insert the scale if it's not 1.

FONT calls OCTALSTRING for the charset, that was in PUP.  Move it to APRINT (but won't remove it from PUP until Nick is finished with the network revisions).


pamoroso:
This change builds and run fine. Font-heavy TEdit documents are displayed with no issues.

PR #2467, opened 2026-01-28
Update to new imagefile architetecture

rmkaplan:
This now fails because the \NULLFDEV doesn't have a GETNEXTBUFFER function.  I don't understand the logic for having that device.

I believe this is otherwise compatible with the new architecture:  (COPYFILE.TO.IMAGEFILE xxx yyy 'HPGL) gets down to the first \BOUT in COPY.TEXT.TO.IMAGE, and fails.


pamoroso:
I get an error when calling `COPYFILE.TO.IMAGEFILE`:

```
4_ (FILESLOAD HPGL)

{DSK}<home>paolo>medley>medley>lispusers>HPGL.LCOM;1
compiled on 28-Jan-2026 01:01:30
File created 28-Jan-2026 01:01:29
HPGLCOMS
(PRINTFILETYPES reset)
(PRINTERTYPES reset)
(IMAGESTREAMTYPES reset)
({DSK}<home>paolo>medley>medley>lispusers>HPGL.LCOM;1)
5_ (COPYFILE.TO.IMAGEFILE 'BITMAPFNS.TEDIT 'BITMAPFNS.HPGL 'HPGL)
COPYFILE.TO.IMAGEFILE is an undefined function.
```

rmkaplan:
Call CONVERT.TO.IMAGEFILE.

I expect this to fail because the HPGL stream is broken at a fundamental level--the HPGL image stream doesn't support BOUT. But with respect to the new interface, the failure should be underneath TEDIT.TO.IMAGEFILE.  The HPGL entry in PRINTFILETYPES should now be quite minimal.

rmkaplan:
You could also try (COPYFILE 'BITMAPFNS.TEDIT '{LPT}.HPGL), that should fail for the same reason.

pamoroso:
Still not the expected failure mode:

```
4_ (FILESLOAD HPGL)

{DSK}<home>paolo>medley>medley>lispusers>HPGL.LCOM;1
compiled on 28-Jan-2026 01:01:30
File created 28-Jan-2026 01:01:29
HPGLCOMS
(PRINTFILETYPES reset)
(PRINTERTYPES reset)
(IMAGESTREAMTYPES reset)
({DSK}<home>paolo>medley>medley>lispusers>HPGL.LCOM;1)
5_ (CONVERT.TO.IMAGEFILE 'BITMAPFNS.TEDIT 'BITMAPFNS.HPGL 'HPGL)
In ERROR:
Can't convert TEDIT file to HPGL
{DSK}<home>paolo>il>BITMAPFNS.TEDIT;1

6_: ^

7_ (CONVERT.TO.IMAGEFILE 'STRINGSCOPE 'STRINGSCOPE.HPGL 'HPGL)
In ERROR:
Can't convert TEXT file to HPGL
{DSK}<home>paolo>il>STRINGSCOPE.;1

8_: ^

9_ (PRINTDEF PRINTFILETYPES)
[(HPGL (EXTENSION (HPGL PLOT))
       (BITMAPFILE (HARDCOPYW.HPGL FILE BITMAP SCALEFACTOR REGION ROTATION 
                          TITLE)))
 (TEDIT (TEST TEDIT.IMAGESOURCEP)
        (EXTENSION (TEDIT TED)))
 (BRAVO (TEST BRAVOFILEP)
        (EXTENSION (BRAVO)))
 (PDF (TEST PDFFILEP)
      (EXTENSION (PDF))
      (CONVERSION (POSTSCRIPT PDF.POSTSCRIPT))
      (BITMAPSCALE POSTSCRIPT.BITMAPSCALE)
      (BITMAPFILE (PDF.HARDCOPYW IMAGEFILE BITMAP SCALEFACTOR REGION ROTATION
                         TITLE)))
 (INTERPRESS (TEST INTERPRESSFILEP)
        (EXTENSION (IP IPR INTERPRESS))
        (BITMAPSCALE INTERPRESS.BITMAPSCALE)
        (BITMAPFILE (INTERPRESSBITMAP IMAGEFILE BITMAP SCALEFACTOR REGION 
                           ROTATION TITLE)))
 (POSTSCRIPT (TEST POSTSCRIPTFILEP)
        (EXTENSION (PS PSC PSF POSTSCRIPT))
        (BITMAPSCALE POSTSCRIPT.BITMAPSCALE)
        (BITMAPFILE (POSTSCRIPT.HARDCOPYW IMAGEFILE BITMAP SCALEFACTOR REGION
                           ROTATION TITLE)))
 (HTML (TEST HTMLFILEP)
       (EXTENSION (HTML HTM)))
 (BITMAP (TEST BITMAPP))
 (WINDOW (TEST WINDOWP))
 (SCREENREGION (TEST REGIONP))
 (BITMAPFILE (TEST BITMAPFILEP)
        (EXTENSION (BM BITMAP))
        (CONVERSION (BITMAP BITMAP.TO.BITMAPFILE)))
 (TEXT (TEST LISPSOURCEFILEP)
       (EXTENSION (TXT TEXT)))
 (DEFAULT (CONVERSION (BITMAP BITMAP.TO.IMAGEFILE WINDOW WINDOW.TO.IMAGEFILE
                             SCREENREGION SCREENREGION.TO.IMAGEFILE 
                             BITMAPFILE BITMAPFILE.TO.IMAGEFILE TEXT 
                             TEXT.TO.IMAGEFILE BRAVO TEDIT.TO.IMAGEFILE TEDIT
                             TEDIT.TO.IMAGEFILE]
NIL
10_ (COPYFILE 'BITMAPFNS.TEDIT '{LPT}.HPGL)
DEFAULTPRINTER cannot print files of type HPGL
```

rmkaplan:
Yes, that's what I would expect.  A very low-level error (GETNEXTBUFFER is undefined in the file device) propagates upwards and is caught and reported as can't convert (which is true).

On the COPYFILE, I'll check to make sure that that's the way it is supposed to report the error in that context.

nbriggs:
On this branch, doing `(COPYFILE "lorem-list.txt" "{LPT}.HPGL")` results in
Screenshot 2026-01-28 at 12 56 03
where it tried to rename a scratch PDF file (OK) to `#<Output Scratch Stream/160,117700>`

rmkaplan:
I get a different message:  "DEFAULTPRINTER can't print files of type HPGL", which is true, whether or not HPGL is loaded (VIEWER is the default printer).  I don't see it getting mixed up with PDF.

But the logic maybe should be a little different, since the printer wasn't specified.  In that case it shouldn't have landed on the defaultprinter, it should have found a printer (PLOTTER) that could in fact print that format (which would then fail).  I thought that's how I set it up, but...

When I put in the explicit printer {LPT}PLOTTER.HPGL, it gives the expected error that even PLOTTER currently fails at printing that format.

rmkaplan:
I looked a little bit more.  I think the message I saw "DEFAULTPRINTER..." is the right one.

When you load HPGL, it defines a PRINTERTYPE (e.g. PLOTTER) that can print HPGL files.  But there isn't any printer of that type on the list of known printers.  So it falls back to the default printer, and reports correctly that it can't deal with HPGL files.

I still don't know why you saw PDF getting into the mix, I think that what I see is the expected behavior.

rmkaplan:
I added some minor adjustments to IMAGEIO and HARDCOPY.  In HARDCOPY now is more careful about how it finds the proper combination of printertype and imagetype, when either or both of those are defaulted.  This also collects the TEXT imagestream and printertype information in one place, with the code in HARDCOPY.

HPGL still doesn't work because of the FDEV issue, but this is an improvement.  So removing draft status, ready for testing (merging?).

pamoroso:
I updated to commit d292ea6.

When printing from TEdit with `Alt-P` > `Send to a printer` > `UNIX` I get the error:

```
lpr: Error - The printer or class does not exist.
```

hardcopy-error

The value of `PRINTERTYPES`:

```
4_ (TEDIT 'BITMAPFNS.TEDIT)
#<Process TEdit/173,113000>
5_ (PRINTDEF PRINTERTYPES)
(((UNIX)
  (CANPRINT (PDF))
  (STATUS TRUE)
  (PROPERTIES NILL)
  (SEND UnixPrint))
 ((DOCUPRINT)
  (CANPRINT (INTERPRESS POSTSCRIPT))
  (HOSTNAMEP NILL)
  (STATUS NSPRINTER.STATUS)
  (PROPERTIES NSPRINTER.PROPERTIES)
  (SEND NSPRINT))
 ((INTERPRESS 8044)
  (CANPRINT (INTERPRESS))
  (HOSTNAMEP NSPRINTER.HOSTNAMEP)
  (STATUS NSPRINTER.STATUS)
  (PROPERTIES NSPRINTER.PROPERTIES)
  (SEND NSPRINT))
 ((FAX TELECOPIER)
  (CANPRINT (INTERPRESS))
  (HOSTNAMEP FAX.HOSTNAMEP)
  (STATUS FAX.STATUS)
  (PROPERTIES FAX.PROPERTIES)
  (SEND FAX.SEND.FILE))
 (VIEWER (CANPRINT (PDF HTML))
        (STATUS TRUE)
        (PROPERTIES NILL)
        (SEND VIEWERPRINT)))
NIL
```

Tracing `ShellCommand`:

shellcommand

As far as I know there's no printer named `UNIX` on my Linux system.

nbriggs:
The old HPGL code thought that you could get an HPGL file from a text file with TEXTTOIMAGESTREAM (as called from MAKEHPGL, which is now gone).  It created a stream with OUTCHARFN of \OUTCHAR.HPGL - yet the created stream does not have that OUTCHARFN, instead it has \MCCSOUTCHAR, which I don't think should happen.

rmkaplan:
The stream created in OPENHPGLSTREAM is explicitly trying to set the OUTCHARFN to \OUTCHAR.HPGL, but the create expression may also be trying to get the OUTCHARFN from the dummy \NULLFDEV, which has \MCCSOUTCHAR as its initialization.

So the create-using may not be doing what is expected--the external format is being smashed in after the fact.

Setting the OUTCHARFN after the stream has been created seems to work

rmkaplan:
For HPGL, I explicitly replaced the OUTCHARFN outside of the create-using.  CONVERT.TO.IMAGEFILE now does not cause an error, does create the file.

For the UNIX printer, the code now does a more consistent job of interpreting the various printername formats.  Choosing UNIX should work

nbriggs:
OK, but that seems at odds with the documentation for `(CREATE ... USING ...)` doesn't it?
```
USING form           Specifies that for all fields not explicitly given a value. the value of the corresponding field in FORM is to be used.
```
Shouldn't `\EXTERNALFORMAT` be avoiding smashing fields that have already been set?

rmkaplan:
Well, yes.  That's another issue--can it recognize what's been set?

nbriggs:
I ran CONVERT.TO.IMAGEFILE with the new HPGL to convert a text file to HPGL format and successfully viewed it in an HPGLviewer.  I haven't tried the graphics operations on the stream yet.

There may be another problem lurking in the stream changes -- when I created an SPP stream I ended up with a stream that I believe used to support standard printing to - its STRMBOUTFN is \BUFFERED.BOUT (what I expect) but its OUTCHARFN is NIL and so printing to it fails.  On the input side, it tries to call \INCCODE.EOLC and also picks up a NIL INCCODE function it's trying to use.   It has a NIL external format on the stream, but the SPP device it's based on is default external format :MCCS

rmkaplan:
I changed \EXTERNALFORMAT so that the non-NIL values for explicit fields in the create expressions won't be overwritten.  That's controlled by a flag, because it is also called from other places.  I reverted HPGL to put OUTCHARFN back in the create expression, and it seems to work.

I didn't look at SPP, maybe just recompiling will shape it up.

pamoroso:
I updated to commit 6add6b4. Now printing to the `UNIX` printer works as expected and `CONVERT.TO.IMAGEFILE` correctly generates an HP-GL file.

rmkaplan:
With respect to HPGL, another (and perhaps more consistent) solution would have been to define a separate external format for the HPGL nullfdev--it might also deploy the other functions in the standard way, FWIW.  Perhaps also for SPP.

Another question about HPGL:  What should happen if the outchar function is given an MCCS character outside of MCCS's 7-bit range?  UTF8?  XCCS?

nbriggs:
HPGL is 7-bit ASCII - I am not sure what provision is in the OUTCHAR (or BOUT) for catching things not in 0..127 (we don't have a :ASCII or :ASCII8 format, I guess) - but it would be fine, I think, for it to error.

At the actual SPP low level I think it's a pass-through 8-bit binary stream - but clients can interpret the data they receive as they wish -- e.g., if you're doing XNS Mail using Courier on an SPP stream the stream is binary for the Courier RPC info, but the data you're transferring is almost certainly XCCS, but the SPP device shouldn't be interpreting what's transferred.

rmkaplan:
I implemented an ascii format, but then decided it didn't have much value and never pushed it.  At best it would just throw an error on an 8-bit byte...or revert to XCCS.  There is a :THROUGH format that just passes all bytes, and an ISO8859/1 (and also an old mac format somewhere).  Those probably need to be updated for the MCCS differences.

rmkaplan:
By the way, which 7-bit Ascii does the HPGL expect?  If not MCCS, the outcharfn presumably has to do a translation.

nbriggs:
The one where when you dump a text string with
```
         (for CHARCODE in (DREVERSE PD.TEXT) do (BOUT PD.STREAM CHARCODE))
```
you get an ASCII caret (94) and an ASCII underscore (95) for what someone typed as ^ or _ when they wrote the code?

pamoroso (approved):
Looks good to me.

masinter:
please update release note when merging, including noting what changed

Issue #2463, opened 2026-01-27
Have man bring up FB on library/tedit/docs/ and library/sketch/docs/


rmkaplan:
We don't have a DINFO index or the sources for the sketch and tedit documentation, but now we are now distributing the pdfs.

It would be nice if

man tedit
man sketch

At least brought up the filebrowser on the library/--/docs/ directories.  The See command would then show selected files in external windows.

Looks like an extension to GENERIC.MAN.LOOKUP, parallel to the commonLisp lookup, but I'm not sure.

pamoroso:
Does PR #2468 supersede this issue?

rmkaplan:
Yes

PR #2468, opened 2026-01-28
man tedit and man sketch

rmkaplan:
I took a stab at extending HELPSYS so that it will show the various pdf files in the library/sketch/docs and library/tedit/docs/ directories.  It uses ShellOpen to bring up the directory windows on the desktop instead of using FILEBROWSER in Medley.  Works on the Mac, don't know about other platforms.

Address #2463


pamoroso:
On Linux `MAN TEDIT` and `MAN SKETCH` open the system file manager on the corresponding directory of the local Medley source tree, here it is for TEdit:

tedit-man

From there I can double-click on a document to open it.

But what if the user doesn't have the source tree locally? Wouldn't it be better to call `ShellBrowse` on https://files.interlisp.org/medley/library/tedit/docs/?

pamoroso:
On the other hand a local installation of Medley includes the relevant files, so it can probably be assumed they are available.

PR #2474, opened 2026-02-01
MCCS coding for UNIX-GETENV, fix MTOUTF8STRING

rmkaplan:


pamoroso:
The loadup fails with the following error on Linux Mint 22.1 Cinnamon:

```
paolo@lispmachine:~/medley/medley$ ./scripts/loadup-all.sh -apps
>>>>> START loadup-init
"/home/paolo/bin/lde" "/home/paolo/medley/medley/internal/loadups/starter.sysout" -id "loadup_init_1" -title "Medley::loadup_init_1" -g 1024x768 -sc 1024x768 -noscroll
LOGINDIR: "/home/paolo/medley/medley/loadups/build/logindir"
GREET FILE: "/home/paolo/medley/medley/loadups/build/loadup-init.init"
REM.CM FILE: ""
VMEM FILE: "/home/paolo/medley/medley/loadups/build/logindir/vmem/lisp_loadup_init_1.virtualmem"
+++++ SUCCESS +++++
..... files created .....
-rw-rw-r-- 1 paolo paolo 4411904 Feb  1 12:01 /home/paolo/medley/medley/loadups/build/init.dlinit
-rw-rw-r-- 1 paolo paolo 118109 Feb  1 12:01 /home/paolo/medley/medley/loadups/build/init.dribble
-rw-rw-r-- 1 paolo paolo 4264960 Feb  1 12:01 /home/paolo/medley/medley/loadups/build/init.sysout
-rw-rw-r-- 1 paolo paolo 55826 Feb  1 12:00 /home/paolo/medley/medley/loadups/build/RDSYS
-rw-rw-r-- 1 paolo paolo 43012 Feb  1 12:00 /home/paolo/medley/medley/loadups/build/RDSYS.LCOM
-rw-rw-r-- 1 paolo paolo 89183 Feb  1 12:00 /home/paolo/medley/medley/loadups/build/I-NEW
-rw-rw-r-- 1 paolo paolo 70676 Feb  1 12:00 /home/paolo/medley/medley/loadups/build/I-NEW.LCOM
<<<<< END loadup-init

>>>>> START loadup-mid-from-init
"/home/paolo/bin/ldeinit" "/home/paolo/medley/medley/loadups/build/init.dlinit" -id "loadup_mid_from_init_1" -title "Medley::loadup_mid_from_init_1" -g 1024x768 -sc 1024x768 -noscroll -NF
LOGINDIR: "/home/paolo/medley/medley/loadups/build/logindir"
REM.CM FILE: "/home/paolo/medley/medley/loadups/build/loadup-mid-from-init.cm"
VMEM FILE: "/home/paolo/medley/medley/loadups/build/init-mid.sysout"
evaluating initial expressions:
PACKAGE-CONVERSION-TABLE.EXPRESSIONS

[...]

MAKEINIT.EXPRESSIONS
 
+++++ SUCCESS +++++
..... files created .....
-rw-rw-r-- 1 paolo paolo 5097984 Feb  1 12:01 /home/paolo/medley/medley/loadups/build/init-mid.sysout
<<<<< END loadup-mid-from-init

>>>>> START loadup-lisp-from-mid
"/home/paolo/bin/lde" "/home/paolo/medley/medley/loadups/build/init-mid.sysout" -id "loadup_lisp_from_mid_1" -title "Medley::loadup_lisp_from_mid_1" -g 1024x768 -sc 1024x768 -noscroll
LOGINDIR: "/home/paolo/medley/medley/loadups/build/logindir"
REM.CM FILE: "/home/paolo/medley/medley/loadups/build/loadup-lisp-from-mid.cm"
VMEM FILE: "/home/paolo/medley/medley/loadups/build/logindir/vmem/lisp_loadup_lisp_from_mid_1.virtualmem"
System call error: open errno=13 Permission denied
----- FAILURE loadup-lisp-from-mid-----
..... files created .....
<<<<< END loadup-lisp-from-mid

----- loadup-all: FAILURE -----
```

Some files are missing:

```
paolo@lispmachine:~/medley/medley$ ls -l /home/paolo/medley/medley/loadups/build/loadup-lisp-from-mid.cm
ls: cannot access '/home/paolo/medley/medley/loadups/build/loadup-lisp-from-mid.cm': No such file or directory
paolo@lispmachine:~/medley/medley$ ls -l /home/paolo/medley/medley/loadups/build/logindir/vmem/lisp_loadup_lisp_from_mid_1.virtualmem
ls: cannot access '/home/paolo/medley/medley/loadups/build/logindir/vmem/lisp_loadup_lisp_from_mid_1.virtualmem': No such file or directory
```

The dribble files: [loadups.zip](https://github.com/user-attachments/files/24991698/loadups.zip)

nbriggs:
This is likely due to the underscores embedded in the names of environment variables that are present in (UTF-8) text in the loadup script files, but are being interpreted as MCCS as they're read in to Lisp.

Needs some rethinking of how this works.  One alternative that just sidesteps the problem is to NOT use underscores in environment variable names in the scripts.  Unfortunately that just leaves the problem to bite someone else later.

rmkaplan:
During the loadup sequence it fakes the :UTF-8 format with :XCCS, which maps the underscore/caret to/from MCCS in the right way.  The scripts shouldn't otherwise go out of 7-bit ascii.

There are some left-arrow-to-underscore edits needed in the apps greet file, I've had trouble working there because some packages aren't available.

nbriggs:
It looks as though you have to run in the apps sysout in order to edit the apps-init greet file.

rmkaplan:
As noted, above the loadup now goes through...but the characters in the exports.all file are screwed up.  Something somewhere is using the wrong format.

pamoroso:
With the commits up to 4b38bab the apps loadup completes successfully.

Setting these environment variables before starting Medley:

```
paolo@lispmachine:~/medley/medley$ export UNDERBAR=_
paolo@lispmachine:~/medley/medley$ export CARET=^
paolo@lispmachine:~/medley/medley$ echo $UNDERBAR
_
paolo@lispmachine:~/medley/medley$ echo $CARET
^
```

yields these `UNIX-GETENV` values in Medley:

unix-getenv

pamoroso:
I updated to commit 3b58828, still looking good.

rmkaplan:
I extended EXTERNALFORMAT do deal systematically with the translation of strings from MCCS to/from other encodings, and not just to translate the contents of files.  The primary motivation is to have a clean interface for mapping between MCCS strings and their representation in the format of the host operating system (typically UTF-8).  This interface is implemented by the  new functions MTOSYSSTRING and SYSTOMSTRING.  Calls to those functions have been inserted in appropriate places (mostly in LLSUBRS and UFS).

(After all this, I still don't know why the script makes EXPORTS.ALL in the wrong format, whereas a directly call from my code(with or without my greeting) makes it with the right format.  (The test is whether a search for dollar-sign finds something--the wrong format has that has the currency character in its place.)

So this is still a draft, but worth poking at.

pamoroso:
On Linux Mint 22.1 Cinnamon I updated to commit ee0fa13 and did some tests. I evaluated the following expressions and recorded the session to this dribble file: [DRIBBLE.TXT](https://github.com/user-attachments/files/25150894/DRIBBLE.TXT)

dribble

To insert the last 3 nonblank characters of the string of event 6 I typed `F10`, `F6`, and `F4`.

rmkaplan:
This looks right.

The dribble file is MCCS so is goofy unless viewed inside Medley

rmkaplan:
This fixes the problem with EXPORTS.ALL.  The default reader environment had a hidden specification of :XCCS, not changed to the value of *DEFAULT-EXTERNALFORMAT*, which is :MCCS.

I'm removing the Draft designation, this is ready for testing

pamoroso:
Still nothing unusual to report at commit 603b099.

rmkaplan:
:MCCS external format does not allow runs of 2 byte characters.  New format :THROUGH16 is a new 2-byte format with no code translations, used by linebuffer.

pamoroso:
It still works at commit 43ef965.

pamoroso:
Still looking good at commit 75d527d.

rmkaplan:
With these current commits the ambiguity of MCCS encoding is resolved in favor of character-set runcodes (e.g. bytes 255 [greek] A B instead of 255 225 0 [greek]A [greek] B).  

I'm having second thoughts about this, since the input/output code (particularly for Tedit) is more direct if fatstrings are written in a single run without charset switching in the middle.

So the new proposal is that a thin string would be written somewhere after a preceding 255 0, and a fat string would be written as 2-byte codes after a preceding 255 255 0, and there would be no need to track changes in its high-order bytes.

We would still read both formats, but we would only write charset-0 thin runs, or fat runs.  The only glitch is that FILEPOS on current files with 1-byte runs outside of cs-0 might fail. (With current commits it would fail for all 2-byte runs).

pamoroso:
Nothing unusual to report at commit 18692f2.

masinter:
@rmkaplan 
The distinction between 'thin' and 'fat' strings should be at a different level and somewhat independent of the nature of run coding.The replacement of a thin string by a fat one can happen at any time. I don't think there is any process that turns /fat/ strings into thin ones.  The decision of run codes and runs should be made independently -- the string is a sequence of chatacter codes -- the representation is either 8 or 16 bits, where you need 16 bits for character codes > 255 but other than that you have no information.

at the string representation level there are no "runs".

rmkaplan:
I'm talking about the complexity of decoding/encoding the byte sequences that correspond to individual characters or sequences of characters in files, which is conceptually different from the thin/fat distinction of strings.

The runs are in the MCCS file representation, which includes an XCCS space efficiency hack for sequences of characters in the same character set.  That creates an ambiguity in how any 16-bit character may be represented.  We could do 1-byte runs of characters in the same character set, or runs of 2-byte sequences for characters outside of charset 0, but not allow both schemes as alternatives (as XCCS does).   (We could go so far as to use 2-byte sequences for all characters, even characters in character set 0--but that would double the typical space and even Ascii would not be readable outside of Medley).

My first proposal was to do 1-byte same-charset runs everywhere, eliminate the ambiguity by getting rid of the 2-byte possibility.  But I think implementation is simpler and less disturbing to go the other way, get rid of the 1-byte runs of non-0 charset characters in favor of 2-byte representation for those characters.  So still retain space efficiency for charset 0.

rmkaplan:
I reverted to the previous version of the MCCS file that did not try to eliminate the ambiguous byte encoding.  I'll deal with that in a separate PR after further discussion.

pamoroso:
Nothing unusual to report at commit e68c69c.

pamoroso (approved):
Looks good to me.

masinter:
FILEPOSd FFILEPOS can work when the encoding is "ambiguousa" -- you just have to compare the encodings of the search string each way.

rmkaplan:
There are exponentially different ways

masinter:
@rmkaplan you only need to consider the alternative encodings of the end of the match segment.

rmkaplan:
That is still some number of characters at the end of the pattern with alternative representations, but then you don't know how many bytes to skip over without decoding.

Issue #2476, opened 2026-02-04
TEDIT.MAP.OBJECTS returns wrong info when COLLECT?='VALUE


fghalasz:
The function TEDIT.MAP.OBJECTS when the COLLECT? argument is 'VALUE returns a list of the character positions of the image objects.  It is supposed to return a list of the values returned by (APPLY* FN CH# OBJ FNARG).

fghalasz:
Fixed in PR# 2478

fghalasz:
Changes incorporated into PR#2479

PR #2478, opened 2026-02-04
Fixed Issue #2476 - TEDIT.MAP.OBJECTS now returns correct info when COLLECT? argument is 'VALUE

fghalasz:
TEDIT.MAP.OBJECTS was returning a list of character positions when the COLLECT? argument is 'VALUE.
It now returns - as expected - a list of the values returned by (APPLY* FN CH# OBJ FNARG).


fghalasz:
Closing this PR since changes have been incorporated into PR#2479.

PR #2484, opened 2026-02-09
Rename 'NOBIND' table name to 'UNBOUND' to allow running interpreted

masinter:
evaluating a variable bound to NOBIND causes errors when run interpreted (doesn't when run compiled).  It's not clear why this design decision was made, but in this case compiling MASTERSCOPE was hung up trying to run a subfuncion involved in macroexpansion after LOAD(MASTERSCOPE).


rmkaplan (approved):
Now possible to compile

pamoroso:
Nothing unusual to report on Linux Mint 22.1 Cinnamon.

PR #2488, opened 2026-02-11
Fixed COMPAREDIRECTORIES See command so it doesn't fail if only one file to show

rmkaplan:
Put in NIL guards before calling PDFFILEP


pamoroso:
The change seems to work as intended on Linux Mint 22.1 Cinnamon.

PR #2490, opened 2026-02-12
Fix POSTSCRIPTSTREAM loading of PSCFONT so unaccented fonts aren't re-encoded.

### LINKED ISSUE #2489: POSTSCRIPTSTREAM mishandles ZapfDingbats glyph encoding.
There’s a bug in the `POSTSCRIPTSTREAM` that doesn’t exclude the **ZapfDingbats** font from re-encoding of the font to get access to _some_ extended characters. This causes many to be incorrect in a `FONTSAMPLER` page (and in general). 

The incorrect encoding: [ZD.pdf](https://github.com/user-attachments/files/25247554/ZD.pdf)

The correct encoding: [ZDX.pdf](https://github.com/user-attachments/files/25247558/ZDX.pdf) created by hand editing the PostScript from the incorrect `FONTSAMPLER` output.

MattHeffron:
Fix POSTSCRIPTSTREAM loading of `.PSCFONT` files so unaccented fonts (_e.g._, ZapfDingbats) aren't re-encoded.
The problem was that `PSCFONT.READFONT` didn't set `*PACKAGE*` so some symbols were being read in the current _arbitrary_ package. 
Also, creating the FULL loadup was prepopulating the PostScript font cache, so all PS fonts had this issue, in any environment based on full.

(This means that the FULL loadup/sysout must be rebuilt to test this.)

This resolves #2489 
Here is a FONTSAMPLER page for ZapfDingbats created with this fix: [ZD-fixed.pdf](https://github.com/user-attachments/files/25273597/ZD-fixed.pdf) Compare with files in the issue noted above.


rmkaplan:
Is FIND-PACKAGE the only change?  No changes to any of the mapping tables?

MattHeffron:
> Is FIND-PACKAGE the only change? No changes to any of the mapping tables?

Yes, this is **_just_** the fix for the error in handling the "unaccented" fonts.
I'll deal with the mapping tables, separately.

rmkaplan (approved):
Lgtm

Issue #2492, opened 2026-02-13
INSPECTCODE using TEDIT formatting very broken


MattHeffron:
`(INSPECTCODE 'FUNCTION)` at the latest master branch commit (dc86cd8f80) has very broken formatting. It looks like corrupted font width info.
<img width="508" height="537" alt="Image" src="https://github.com/user-attachments/assets/1342f4e8-5a8e-47f4-8b25-737cefd81cdb" />

In an older sysout (commit b0f92834), it's fine.
<img width="538" height="425" alt="Image" src="https://github.com/user-attachments/assets/ee050f90-3f02-474e-95db-1a31e28234d0" />
(This was a .virtualmem file from Jan 27.)

nbriggs:
This appears to be because it is formatting in Classic now, but it needs to be in a fixed-pitch font.  It used to be in Gacha.

nbriggs:
DEFAULTFONT for DISPLAYFD has changed from GACHA to CLASSIC.

New:

<img width="385" height="141" alt="Image" src="https://github.com/user-attachments/assets/412096b8-e79a-47fa-9314-a80f2fb55005" />

Old:

<img width="382" height="123" alt="Image" src="https://github.com/user-attachments/assets/cd6a909a-e335-4767-a815-76364fd5db41" />

MattHeffron:
So, are there **two** problems?
1. It's using Classic (variable width) instead of Gacha (fixed width)
2. The widths appear to be bizarrely messed up.

nbriggs:
The Classic font it has its hands on is messed up -- if I start a TEDIT, type some characters, display the Character looks -- it says it's Classic, but the glyphs it is displaying are actually Gacha.  I tried flushing the font cache, recreating Classic 12, and setting that as the DISPLAYFD in DEFAULTFONT, and then a normal TEDIT window displayed in Classic instead of Gacha, and a code inspector looks better (though NOT good) - so I think it's an issue with how the required fonts are getting instantiated in the basic image.

<img width="427" height="347" alt="Image" src="https://github.com/user-attachments/assets/f636e20e-5b3b-478e-ada4-2ffeead8c464" />

nbriggs:
DEFAULTFONT is OK in the lisp.sysout (it has GACHA 10 MRR in the DISPLAYFD slot), but it's got the broken CLASSIC 12 MRR in the full.sysout built from that lisp.sysout.

rmkaplan:
[...]

PR #2496, opened 2026-02-14
Reading indirect charsets in MEDLEYFONT doesn't smash the default font

rmkaplan:
Addresses the major symptom described in #2492.  Faking italic and bold character bitmaps still requires further examination.  But this should fix the INSPECTCODE formatting problem and hopefully related glitches that have been encountered.

\READCHARSET is also changed slightly so that it returns NIL if given a NIL FONTSPEC.  This may allow cleaner implementations of some simple device charset functions.


pamoroso:
On Linux Mint 222.1 Cinnamon this PR fixes the `INSPECTCODE` and shell VTCHAT issues. This is what a shell VTCHAT looks like now:

vtchat-shell

PR #2502, opened 2026-02-17
delete old fonts no longer needed

masinter:


rmkaplan:
Or move to obsolete?

masinter:
we might need to keep some of the strike-format display fonts? not for releasing but in case we want to fix some problem?

rmkaplan:
I would be enough if they are kept under fonts/displayfonts/ in the obsolete repo.  We could refer to them systematically with {OBSOLETE}/fonts/displayfonts/, where {OBSOLETE} is a pseudohost pointing to obsolete's local clone.

nbriggs:
How about the (Xerox) logo font?

rmkaplan:
There is a Medley font for XEROXLOGO48

masinter:
> We could refer to them systematically with {OBSOLETE}/fonts/displayfonts/, where {OBSOLETE} is a pseudohost pointing to obsolete's local clone.

i don't think it's a good idea to require anyone to maintain a local clone of the 'obsolete' repository. As a fallback it's fine, but if you need to refer to it systematically implies it's not really "obsolete".

rmkaplan:
Nobody should be "required" to maintain an obsolete repository with obsolete fonts.  But those fonts should still be available in the event that, for some reason, we want to rebuild our current fonts from the original sources.  For example, to fill in the Gacha glyphs from a BDF mono font before going to, say, Modern as we do now.  In that case we would want to start fresh.

rmkaplan:
I'd like to move forward on this, to clean up the active directories.  If someone could produce a new FONTSOURCE repository in the right place in the scheme of things, I will populate it with the current non-Medleyformat fonts on fonts/display/ fonts/postscript fonts/ipfonts/ etc. and remove those from {MEDLEY}/fonts/.

Those files won't be "obsolete", they'll be the sources for running IMPORTFONTS to create, recode, and deploy updated versions of the Medley fonts in fonts/medley[device]/fonts.  The pseudohost {FONTSOURCE} can be set to the local installation of that repository, so IMPORTFONTS can find those files by default.

 (We could also then simplify the names of the deployed-fonts directories to just fonts/ with subdirectories display/ postscript/ interpress/.)

PR #2524, opened 2026-03-02
PAIRS arguments of MOVEFONTCHARS can be a hashtable

rmkaplan:
To help in importing new font files.  MCCSMAPPAIRS returns the Unicode-to-MCCS hash table when asked for a mapping from a UNICODE-coded font to MCCS.  This fixes MOVEFONTCHARS so that it will map through that hash table to move glyphs around.

Also changes FONTDEVICEPROP so that it can be used to set a new value given as a third argument.


rmkaplan:
MEDLEYFONTFORMAT.READ.FONT was confused when given a filename and not also a font

pamoroso:
It builds and runs with no issues and fonts seem to work as usual.

pamoroso:
Still working fine at commit d324300.

pamoroso:
Still nothing unusual at commit b4aa468.

pamoroso:
Still looking good at commit a24ea25.

rmkaplan:
This is superceded by the #2555 , closing and deleting

PR #2532, opened 2026-03-09
Add maiko, notecards, loops, test as git submodules

masinter:
## Summary

- Adds `maiko`, `notecards`, `loops`, and `test` as git submodules, each pinned to the current HEAD of their default branch (`master` for maiko/test, `main` for notecards/loops)
- Removes `maiko/` from `.gitignore` so it can be tracked as a submodule
- Submodule paths match what the existing build scripts already expect (`$MEDLEYDIR/maiko`, `$MEDLEYDIR/notecards`, etc.), so no script changes are needed

## After merging

Contributors cloning fresh will need:
```sh
git submodule update --init --recursive
```

To advance submodules to latest upstream:
```sh
git submodule update --remote
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/buildLoadup.yml
+++ b/.github/workflows/buildLoadup.yml
@@ -124,6 +124,8 @@ jobs:
       # Checkout latest commit
       - name: Checkout Medley
         uses: actions/checkout@v4
+        with:
+          submodules: true
 
       # Setup release tag
       - name: Setup Release Tag
@@ -190,18 +192,6 @@ jobs:
           tar -xzf "${TARBALL_DIR}/${MAIKO_RELEASE_TAG}-linux.x86_64.tgz"
         env:
           GITHUB_TOKEN: ${{ secrets.MAIKO_TOKEN }}
-
-      # Checkout Notecards and tar it in the tarballsdir
-      - name: Checkout Notecards
-        uses: actions/checkout@v4
-        with:
-          repository: ${{ github.repository_owner }}/notecards
-          path: ./notecards
-      - name: Tar notecards into tarball dir
-        run: |
-          mv ./notecards ../notecards
-          cd ../notecards
-          git archive --format=tgz --output="${TARBALL_DIR}/notecards.tgz" --prefix=notecards/ main
 
       # Install vnc
       - name: Install vnc
--- a/.github/workflows/doHCFILES.yml
+++ b/.github/workflows/doHCFILES.yml
@@ -51,33 +51,8 @@ jobs:
 
       - name: Checkout Medley repo
         uses: actions/checkout@v4
-
-      - name: Checkout maiko
-        uses: actions/checkout@v4
         with:
-          repository: ${{ github.repository_owner }}/maiko
-          path: ./maiko
-
-      - name: Checkout notecards
-        uses: actions/checkout@v4
-        with:
-          repository: ${{ github.repository_owner }}/notecards
-          path: ./notecards
-
-      - name: Checkout loops
-        uses: actions/checkout@v4
-        with:
-          repository: ${{ github.repository_owner }}/loops
-          path: ./loops
-
-      - name: Checkout test
-        uses: actions/checkout@v4
-        with:
-          repository: ${{ github.repository_owner }}/test
-          path: ./test
-
-      - name: Cleanup .git for notecards, loops, test
-        run: rm -rf ./notecards/.git ./loops/.git ./test/.git
+          submodules: true
 
       - name: Download Maiko
         run: |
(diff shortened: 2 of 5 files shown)


masinter:
## How the loadup process changes

**Basic loadup** (`lisp.sysout`, `full.sysout`): nothing changes. These only need the maiko binaries (`lde`, `ldeinit`) on PATH or at `../maiko/` — the source submodule does not provide compiled binaries.

**`apps.sysout`** (requires notecards): previously you needed notecards cloned at `../notecards`. Now it is automatically at `$MEDLEYDIR/notecards` after `git submodule update --init`. The `loadup-apps-from-full.sh` script already checked that path first, so this just makes it automatic.

**HCFILES** (requires loops and test too): previously you had to manually clone those repos and copy them in. Now they are already there.

### What you need to do differently after cloning medley

```sh
# Before:
git clone https://github.com/Interlisp/medley
git clone https://github.com/Interlisp/notecards ../notecards  # if building apps
# ... etc

# After:
git clone --recurse-submodules https://github.com/Interlisp/medley
# or on existing clone:
git submodule update --init --recursive
```

### What has not changed

- You still need maiko **binaries** built or downloaded separately — the maiko submodule is source only
- The actual loadup commands (`./scripts/loadup-all.sh`, etc.) are unchanged
- The fallback to `../maiko`, `../notecards` etc. still works if you prefer sibling directories

nbriggs:
Tried twice to clone medley freshly -- both times it failed with:

```
 % git clone  --recurse-submodules https://github.com/Interlisp/medley medley-submods
Cloning into 'medley-submods'...
remote: Enumerating objects: 24488, done.
remote: Counting objects: 100% (369/369), done.
remote: Compressing objects: 100% (217/217), done.
error: RPC failed; curl 92 HTTP/2 stream 5 was not closed cleanly: CANCEL (err 8)
error: 6783 bytes of body are still expected
fetch-pack: unexpected disconnect while reading sideband packet
fatal: early EOF
fatal: fetch-pack: invalid index-pack output

```

masinter:
I'm a little baffled: I can no longer find a place where it was recommended using
```
 git clone  --recurse-submodules https://github.com/Interlisp/medley medley-submods
```
as those instructions are wrong. The correct instructions are here in the PR 
and in the email.  https://groups.google.com/g/lispcore/c/jaOhaJJn-s4

## After merging

Contributors cloning fresh will need:
```sh
git submodule update --init --recursive
```

To advance submodules to latest upstream:
```sh
git submodule update --remote
```

nbriggs:
I think that if the goal is to have reproducible builds, and to have known compatible maiko and medley code, then the loadup should use the lde/ldex/ldeinit as built in the maiko submodule -- otherwise you're building with something (from outside) that may or may not be related to the vetted lde/ldeinit/ldex source  in the submodule.

rmkaplan:
Does GITFNS have to change?  It has some code for finding various clones along ancestor paths.

pamoroso:
I freshly cloned the medley repo with `git clone --recurse-submodules` at `~/smedley/medley/` but under that tree I don't see the maiko subdirectory (and neither notecards, loops, and test) which I need to build Maiko. But probably I'm misunderstanding how submodules are supposed to work.

nbriggs:
@pamoroso - since the PR hasn't been merged yet, the main branch that you're on doesn't yet have submodules, so you'll need to switch to the PR branch lmm30-use-submodules before doing `git submodule init` (from in the medley directory you cloned).

pamoroso:
I did what you suggested and it worked.

pamoroso:
However, now maiko and the other submodule directories are empty.

nbriggs:
@pamoroso `git submodule update`

pamoroso:
Never mind, updating with `git submodule update --remote` did the trick.

pamoroso:
I successfully made the apps loadup from the submodules on Linux Mint 22.1 Cinnamon.

masinter:
> Does GITFNS have to change? It has some code for finding various clones along ancestor paths.

* GITFNS should look in the submodules subdirectories (medley/loops, medley/test etc) before looking for the repos for those in sister/aunt relations. It might already be set up that way? If you have aa/medley/notecards (subdirecdtory) and aa/medley aa/notecards (sister), which one does it find first?  It SHOULD pick up from subdirectories first.

I have some changes to the .gitignore for each submodule repository. I've proposed PRs for each of notecards, test, loops, maiko. I think those PRs will need to be merged BEFORE merging this one (sigh).

masinter:
too many incertainties...   the maiko as submodule seems to require different handling.
master was updated to use checkout@6 instead of checkout@4. Does it check out submodules?
how?   Anyway,

PR #2545, opened 2026-03-17
Add DISPLAY sanity check to EDITFONT

rmkaplan:
Bad things would happen if EDITFONT is called on a non-display font.


pamoroso:
`FONTCREATE` now breaks with an error as expected when called on a non-display font instead of crashing the session. Tested on Linux Mint 22.1 Cinnamon.

pamoroso:
I meant `EDITFONT` not `FONTCREATE`.

MattHeffron (approved):
LGTM

PR #2552, opened 2026-03-31
Rmk172  clean unicode mapping directory

rmkaplan:
As requested in 3/30/2026 technical meeting.  The composite mapping files were confusing and probably out of date, even though they were not actually being used.


pamoroso:
It builds and works with no issues on Linux Mint 22.1 Cinnamon.

MattHeffron:
Both of the README files still describe the multi-charset files.
Those were eliminated in this PR, right?
Maybe have a note there that combining charsets into single files is deprecated, and not recommended.

Also then, what about `XCCS-JIS.TXT` _vs._ `jis/XCCCS-##=JIS.TXT`?
Shouldn't that multi-charset file be removed also?

rmkaplan:
I only see one README file.

I think the one in unicode/xerox/ is still accurate, in that it is possible to create a mapping file that has multiple character sets.  The code for producing/reading such files hasn't been removed.

We just decided that we wouldn't have those laying around, confusingly. But maybe that would be an easy way for someone to collect a subset of mappings into a single file to send around, or for some other purpose.

I haven't yet verified that the XCCS-JIS.TXT file is redundant.

MattHeffron:
> I haven't yet verified that the XCCS-JIS.TXT file is redundant.

I'll approve this PR. If the XCCS-JIS.TXT file is redundant, that can be another PR.

MattHeffron (approved):
Approved. Potential further changes can be new PRs.

MattHeffron:
> I only see one README file.

There's `README.TEDIT` and `README.TXT`

rmkaplan:
I looked at the 2 README files.  The .TEDIT one is a slight revision, but it should have been saved as plain .txt.  I'll convert it back to txt in the next cleanup PR.

PR #2553, opened 2026-03-31
Fix copy/delete for {UNIX} directories

rmkaplan:
Fix glitches in recent strategy for preserving the Unix (e.g. git {MEDLEY}) file in {NODIRCORE} so that See and Compare work in the same CD browser even after the copy/delete action has changed the target directory.


pamoroso:
I'm not noticing anything unusual on Linux Mint 22.1 Cinnamon.

MattHeffron (approved):

PR #2556, opened 2026-04-06
Add deleted/* (from GITFNS) to .gitignore

rmkaplan:
Only contains local information, and maybe just temporary.   (Maybe GITFNS should put it in {CORE}, but for now...)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,9 @@ loadups/gitinfo
 *.sysout
 *.SYSOUT
 
+# GITFNS deleted subdirectory
+deleted/*
+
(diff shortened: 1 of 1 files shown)


nbriggs (dismissed):
OK.  Keeping in mind that this will match all files in a "deleted" directory at any level, but not files within a subdirectory of a "deleted" directory (that would need to be deleted/**)

rmkaplan:
I changed it to **

nbriggs (approved):
LGTM.

pamoroso:
It works as expected on Linux Mint 22.1 Cinnamon.

PR #2559, opened 2026-04-10
Add preliminary files with mappings from XCCS to Unihan set of Unicode.

MattHeffron:
These were generated, using scripting, from the data in the unihan folder of the Unicode Data Base. 
That info _claims_ to know the mapping from Unicode(unihan) to "Xerox" coding (2 bytes in octal). 
These were not validated **_at all_** for correctness/completeness. 
None of these files have **any** descriptive header.

(Unfortunately, this kind of mapping information to "Xerox" is _only_ for Unihan characters.)


pamoroso:
The nw files appear to be intact and readable.

rmkaplan:
The format seems good.  The only issue is that we need a predicate that is true for the range of Unihan charsetset numbers, like we have for Kanji and Chinese.  That would be used to make sure that the glyphs for the bold and italic versions of the display fonts, if we ever got them, would not be faked up.

MattHeffron:
Unihan **_is_** the unification of Kanji, Chinese, Japanese, etc. characters. See: [Han unification](https://en.wikipedia.org/wiki/Han_unification)
But "many characters have regional variants assigned to different code points"

The XCCS Standard Document map of character sets to languages (page 34, 2-8) appears quite incomplete.
In addition, some character sets include mixes of Latin, Symbols, Asian, Arabic, etc. characters, so _by character set_ management of _faking_ might be insufficient.

rmkaplan (approved):
Why not?  Turns out that the function CHINESECHARSETP was basically picking out just these character sets (except for 171, which KANJICHARSETP should cover), so I'll separately rename CHINESECHARSETP to UNIHANCHARSETP

PR #2566, opened 2026-04-16
Define {MEDLEY} as MEDLEYDIR, function MEDLEYDIR produces {MEDLEY} names

### LINKED ISSUE #2563: Should PSEUDOHOSTS automatically define {MEDLEY}?
**Is your feature request related to a problem? Please describe.**
Frequently there is a need to specify paths to directories that are relative to MEDLEYDIR, e.g. for `LISPUSERSDIRECTORIES`, or `DISPLAYFONTDIRECTORIES`.

**Describe the solution you'd like**
The PSEUDOHOSTS(.LCOM) should define the `{MEDLEY}` pseudohost to be the value of the `IL:MEDLEYDIR` variable.

**Describe alternatives you've considered**
None.

**Additional context**
The `PSEUDOHOSTS.TEDIT` documentation uses creating a `{MEDLEY}` pseudohost _exactly_ like this as the example of a useful definition.

rmkaplan:
This is a stab at defining the pseudohost {MEDLEY} as the (possibly new) value of MEDLEYDIR when a sysout or makesys starts up, as discussed in #2563.

It also wraps PSEUDOFILENAME around the body of MEDLEYDIR, so that (MEDLEYDIR ...) produces a {MEDLEY} file name.  The variable DISPLAYFONTDIRECTORIES is thus returned as {MEDLEY}/fonts/medleydisplayfonts, and so also for other MEDLEY-INIT-VARS (LISPUSERSDIRECTORIES etc.)

Marked as draft for now


pamoroso:
I tested the PR on Linux Mint 22.1 Cinnamon and it works as expected.

rmkaplan:
I worked a little more on this to get the {MEDLEY} pseudohost available early in the loadup, in particular before any fonts are loaded.  I moved PSEUDOHOSTS into 1LISPSET, so it can create {MEDLEY} from the variable MEDLEYDIR, and then MEDLEYDIR in front of that so that MEDLEYDIR is defined in the loadup.

The function MEDLEYDIR then systematically returns the PSEUDOFILENAME of its arguments, so that all of the MEDLEY-INIT-VARS are now {MEDLEY} paths.

Seems to work.

pamoroso:
Doing the apps loadup at commit 193e85e fails with an error and drops to URAID:

loadup-error

```
paolo@lispmachine:~/medley/medley$ ./scripts/loadup-all.sh -apps
>>>>> START loadup-init
"/home/paolo/bin/lde" "/home/paolo/medley/medley/internal/loadups/starter.sysout" -id "loadup_init_1" -title "Medley::loadup_init_1" -g 1024x768 -sc 1024x768 -noscroll
LOGINDIR: "/home/paolo/medley/medley/loadups/build/logindir"
GREET FILE: "/home/paolo/medley/medley/loadups/build/loadup-init.init"
REM.CM FILE: ""
VMEM FILE: "/home/paolo/medley/medley/loadups/build/logindir/vmem/lisp_loadup_init_1.virtualmem"
+++++ SUCCESS +++++
[...]
>>>>> START loadup-lisp-from-mid
"/home/paolo/bin/lde" "/home/paolo/medley/medley/loadups/build/init-mid.sysout" -id "loadup_lisp_from_mid_1" -title "Medley::loadup_lisp_from_mid_1" -g 1024x768 -sc 1024x768 -noscroll
LOGINDIR: "/home/paolo/medley/medley/loadups/build/logindir"
REM.CM FILE: "/home/paolo/medley/medley/loadups/build/loadup-lisp-from-mid.cm"
VMEM FILE: "/home/paolo/medley/medley/loadups/build/logindir/vmem/lisp_loadup_lisp_from_mid_1.virtualmem"

*Error* URAID Called:
Enter the URaid
IL:FIND-DEBUGGER-ENTRY-FRAME

< lu
  0 :    0x12db2 : IL:RAID
  1 :    0x12d9c : IL:FAULTAPPLY
  2 :    0x12d50 : IL:\INTERPRETER
  3 :    0x12d2e : SI:*UNWIND-PROTECT*
  4 :    0x12cf6 : CL:ERROR
  5 :    0x12cda : CL:CERRORA0002
  6 :    0x12ca2 : CL:CERROR
  7 :    0x12c62 : IL:DOFILESLOAD
  8 :    0x12c48 : IL:LOADUP
  9 :    0x12c18 : IL:LOADUP-LISP
 10 :    0x12bf6 : IL:\EVALFORM
 11 :    0x12bdc : CL:PROGN
 12 :    0x12bbc : IL:\EVALFORM
 13 :    0x12ba8 : IL:EVAL
 14 :    0x12b8e : IL:EVALQT
 15 :    0x12b78 : IL:\REPEATEDLYEVALQT
 16 :    0x12b58 : IL:\EVALFORM
 17 :    0x12b0a : IL:\MAKE.PROCESS0
 18 :    0x11802 : CL:T
```

pamoroso:
This is the loadups directory of the failed loadup: [loadups.zip](https://github.com/user-attachments/files/26794535/loadups.zip)

rmkaplan:
What does it show when you issue the commands

f7
f8

If you are seeing with the apps, I would be surprised if you don't also see it with the full.

rmkaplan:
Never mind.  I copied the wrong file from my working directory.  Try it now.

If this works, then the next test is to try to move a sysout from one machine/environment to another, see if FINDFILE for a file on sources/ is found when you are not connected there.  I only have one environment.

pamoroso:
Commit c29c3f0 fixes the loadup, which now completes successfully, and the pseudohost works as expected.

pamoroso:
Still working fine at commit c0bbc3d.

rmkaplan:
I upgraded this from Draft.

rmkaplan:
If/when this is merged, the next step would be to remove all of the entries in MEDLEY-INIT-VARS that call the function MEDLEYDIR, and to put those variables (with {MEDLEY} attached) back in their logical places (POSTSCRIPTFONTDIRECTORIES would appear only on POSTSCRIPTSTREAM, for example).

pamoroso:
Still no issues at commit 8c8b9f7.

rmkaplan:
I'm reverting this to Draft.  PSEUDOHOSTS has been stable for 3 or 4 years, but we have never beaten on the particular configuration of pseudohosts that this transformation entails.

The base pseudohost here is MEDLEY, which gets updated when a system is brought up with a new MEDLEYDIR.  All the (MEDLEYDIR) pseudohosts (e.g. the font directories and {IRM}) are then defined in terms of {MEDLEY}, with the intent that they also would implicitly get updated when MEDLEYDIR and thus {MEDLEY} changes.

A problem emerged as I was poking around with the HCFILES failure files in #2567.  The file {MEDLEY}/docs/dinfo/CHAPACK.TEDIT is specified there.  But the pseudohost IRM is defined as {MEDLEY}/docs/dinfo/, so that that  file also has a shorter synonym {IRM}CHAPACK.TEDIT.  This is OK with respect to the operations that operate on the stream--the truefilename of the stream is opened, but the stream is marked as having whatever name was specified in the calll.  The pseudohost set up is invisible to all of the open-file operations.

The problem is with operations that operate on the name itself, not the stream.  The generic code for GETFILEINFO, for example, calls the device operation for GETFILENAME which in turn calls CONTRACT.PH to return the shortest equivalent prefix, in this case {IRM}CHAPACK.TEDIT.  But the GETFILEINFO code has a handle on the original {MEDLEY} device and thinks that the contracted filename is still valid for that device.  But it isn't: the /docs/dinfo/ has disappeared.  The result is that GETFILEINFO.PH can't find the file, and returns NIL for, say, CREATIONDATE.

It may just be that contraction is in general a bad idea, except for an explicit call in PSEUDOFILENAME.  But it may take a little time to unwind all the pathways.

rmkaplan:
I modified PSEUDOHOSTS to deal with the host-ambiguity problem.  For operations on pseudohost files that return a filename, the return will include the file's original pseudohost even though there is a pseudohost designating the same file with a longer prefix and therefore shorter filename.  The test is to do operations on both {MEDLEY}/docs/info and {IRM}, which denote the same files.  The returned name should be on the particular starting host.

As a further test, I also removed the fontdirectory settings for postscript and interpress from MEDLEY-INIT-VARS, and put them in INTERPRESS and POSTSCRIPTSTREAM as {MEDLEY} variables.

I moved PSEUDOHOSTS in front of MEDLEYDIR in the loadup sequence, so that MEDLEYDIR can initialize the MEDLEY pseudohost.

I'll remove draft status after a little more testing.

pamoroso:
I made the apps loadup at commit 056ebd0 and ran Medley with `medley -a -e -n -v -r -`. I expected this command not to yield an error:

```
2/6_ DIR {IRM}*.*
File not found: {IRM}*.*;*
```

rmkaplan:
I don't get that error, a full.  I get the expected directory listing.

rmkaplan:
I added a new function PSEUDOFILENAMES to give a handle on all the ways that a given file can be referred to.

I looked at the problems that SAMEDIR was having, determined that its logic was screwed up even without the synonyms that pseudohosts may introduce.   So I cleaned it up.

(A separate glitch is that it never worked properly for CL:COMPILE-FILE.  Looking at the strategy, my bet is that there is a RESETLST in CLEANUP/MAKEFILE that is too low to cover the CL:COMPILE-FILE, so the directory reverts too soon.  But that's not in SAMEDIR itself.)

PR #2580, opened 2026-04-27
Fix Issue# 2562 - SEE-PDF fails if there is a space in the given filename

fghalasz:
Fixed Issue#2562 by preventing SLASHIT from backquoting spaces when called from ShellOpen.  To do so added an extra arg to SLASHIT called NO.QUOTE.SPACE and setting it to T when SLASHIT is called from ShellOpen.


pamoroso:
On Linux Mint 22.1 Cinnamon I made the apps loadup and from Firefox saved to `{LI}` as PDF the home page of Interlisp.org. `(SEE-PDF "Interlisp.org Home Page.pdf")` does nothing and the Exec prompt doesn't come back, requiring me to interrupt with `^D` or `^E` to regain control:

see-pdf

rmkaplan:
It works for my test case

rmkaplan:
However, when I click the See button on a pdf in the file browser, a TTY window pops to show the ShellOpen command.  Maybe some left over debugging?

pamoroso:
I can reproduce the TTY window of the File Browser issue.

fghalasz:
Fixed the TTY Window issue.  It indeed was a leftover PRINT statement added for debugging.

pamoroso:
The commits up to b1957a8 fix the [issue](https://github.com/Interlisp/medley/pull/2580#issuecomment-4325866897) for me.

fghalasz:
Fixed issue where SEE-PDF (and ShellOpen in general) goes into an infinite loop when there are two or more spaces in the filename.q

rmkaplan (approved):
LGTM

PR #2581, opened 2026-04-27
Fixing \DO.PARAMS.AT.OPEN take two

rmkaplan:
I took another shot.  I didn't bother with the FIND-FORMAT, if it isn't an EOL than go directly to \EXTERNALFORMAT, let it error.

(The reason that FIND-FORMAT doesn't deal with :DEFAULT is that that specifies the default for the file device, and FIND-FORMAT doesn't take a stream, unlike \EXTERNALFORMAT.)


pamoroso:
It seems to work as expected on Linux Mint 22.1 Cinnamon.

pamoroso (approved):
It works as described.

PR #2590, opened 2026-04-30
MCCS now includes XCCS external format, separate XCCS file now in Obsolete

rmkaplan:


pamoroso:
It seems to be working as intended on Linux Mint 22.1 Cinnamon.

masinter (approved):
I think we moved Obsolete outside of Medley and thus outside of Medley PRs. And to move a file to be part of another, it would be better to add the MCCS part to the XCCS part in the same PR.

But I think these comments can be addressed when merging so: approved

rmkaplan:
The vestiges ofXCCS were assimilated to MCCS long ago, it hasn't been part of the loadup for a long time.

PR #2593, opened 2026-05-07
Fix library>DATABASEFNS  DUMPDB to use get the FILEDATES correctly

MattHeffron:
`DUMPDB` in `library>DATABASEFNS` didn't get the `FILEDATES` correctly if the _actual_ code file name was not `U-CASE`.
Because of this, the attempt to match `.DATABASE` version with code file failed, as the `.DATABASE` file had `NIL` as the `FILEDATES` information.
Fixed to use `ROOTFILENAME` instead of `NAMEFIELD`.


rmkaplan (approved):
LGTM

pamoroso:
On Linux Mint 22.1 Cinnamon it builds and works with no apparent issues.

PR #2596, opened 2026-05-07
Minor fix in UNICODE-FORMATS

### LINKED ISSUE #2595: MAKESYS in FULL stage tries to write /tmp.SYSOUT-temp at filesystem    root (EACCES)
# `loadup-all.sh --target full` fails with EACCES: MAKESYS writes its temp sysout to `/tmp.SYSOUT-temp` at the filesystem root

**Describe the bug**

Building `full.sysout` from `lisp.sysout` (the `loadup-full-from-lisp` stage,
whether invoked directly or as part of `loadup-all.sh`) fails with
`System call error: open errno=13 Permission denied` on Linux when run as a
non-root user. The init/mid/lisp stages all succeed; only the FULL stage fails.
No `full.sysout` or `full.sysout-temp` is produced anywhere on disk.

`strace` shows Maiko's `vmem_save` (in `maiko/src/vmemsave.c:367`) trying to open
`/tmp.SYSOUT-temp` for writing — i.e. literally a file `tmp.SYSOUT-temp` in
the **filesystem root `/`**, not under `/tmp/`:

```
openat(AT_FDCWD, "/tmp.SYSOUT",      O_WRONLY)                          = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/tmp.SYSOUT-temp", O_WRONLY|O_CREAT|O_TRUNC, 0666)     = -1 EACCES (Permission denied)
```

`/` is owned by root, mode 0755, so the non-root user cannot create files there.
This is correct OS behavior; the bug is that MAKESYS is asking for that path at all.

The path comes from `\COPYSYS` (`sources/ADIR:228-232`), which hard-codes the
temp filename as `tmp.SYSOUT` with the directory taken from `TARGETFILE`:

```lisp
(DSK [SETQ TEMPNAME (PACKFILENAME.STRING 'HOST TARGETHOST 'NAME 'tmp 'EXTENSION
                           'SYSOUT
                           'BODY
                           (\UFS.RECOGNIZE.FILE TARGETFILE 'NON (\GETDEVICEFROMNAME
                                                                 TARGETHOST]
     (SETQ VAL (\FLUSHVM TEMPNAME)))
```

In the FULL stage, `TARGETFILE`'s directory ends up empty, so `TEMPNAME` is
`{DSK}tmp.SYSOUT`, which translates to `/tmp.SYSOUT` on the host.

The init/mid/lisp stages use the same construction (`(IL:CONCAT (QUOTE {DSK})
(IL:UNIX-GETENV (QUOTE LOADUP_WORKDIR)) (IL:L-CASE (QUOTE /<name>.sysout)))`)
and land their sysouts in the right place. The FULL stage does not; what
differs at runtime is the state set up by everything `LOADUP-FULL` loads
plus the `(IL:ENDLOADUP)` call in `loadup-full-from-lisp.cm` (which resets
`\CONNECTED.DIRECTORY` back to bare `{DSK}` per the `SYSTEMINITVARS` entry
in `sources/APUTDQ:170`).

**To Reproduce**

1. Files loaded:
   - `medley` repo at commit `6159c64b`, no local modifications.
   - `maiko` repo at commit `9259716`, built with the standard `lde` and
     `ldex` for `linux.x86_64`.
   - The starter sysout from `internal/loadups/starter.sysout`.

2. Forms to eval / commands to run (from a non-root shell, in the `medley`
   repo root):
   ```sh
   ./scripts/loadups/loadup-all.sh
   ```
   or, equivalently, after init/mid/lisp have already succeeded:
   ```sh
   ./scripts/loadups/loadup-all.sh --start lisp --target full
   ```

3. What happened: init, mid, and lisp stages succeed and produce
   `loadups/build/{init.dlinit,init-mid.sysout,lisp.sysout}` correctly.
   The FULL stage fails:
   ```
   >>>>> START loadup-full-from-lisp
   "/.../maiko.git/linux.x86_64/lde" "/.../loadups/build/lisp.sysout" -id "loadup_full_from_lisp_1" ...
   MEDLEYDIR: "/.../medley.git"
   LOGINDIR:  "/.../loadups/build/logindir"
   GREET FILE: "/.../greetfiles/NOGREET"
   REM.CM FILE: "/.../loadups/build/loadup-full-from-lisp.cm"
   VMEM FILE:  "/.../loadups/build/logindir/vmem/lisp_loadup_full_from_lisp_1.virtualmem"
   System call error: open errno=13 Permission denied
   ----- FAILURE loadup-full-from-lisp-----
   <<<<< END loadup-full-from-lisp
   ----- loadup-all: FAILURE -----
   ```
   `loadups/build/full.dribble` is written and shows the loadup completed
   normally up through `FULL fonts loaded`. No `full.sysout`, no
   `full.sysout-temp`, no `tmp.SYSOUT*` is created anywhere.

**Expected behavior**

The FULL stage should produce `loadups/build/full.sysout` (and `loadup-all.sh`
should then promote it to `loadups/full.sysout`), exactly the way the LISP
stage produces `loadups/build/lisp.sysout`. The MAKESYS call in
`loadup-full-from-lisp.cm` passes a fully-qualified path
(`{DSK}/.../loadups/build/full.sysout`); the temp file should be written
adjacent to that target, not at the filesystem root.

**Screenshots**

N/A — the failure is on stdout (shown in step 3 above). Full strace evidence:

```
$ strace -f -e trace=openat -o /tmp/full.strace \
        ./scripts/loadups/loadup-all.sh --start lisp --target full
$ grep -E 'EACCES|tmp\.SYSOUT|full\.sysout' /tmp/full.strace | grep -v /proc/
openat(AT_FDCWD, "/tmp.SYSOUT", O_WRONLY) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/tmp.SYSOUT-temp", O_WRONLY|O_CREAT|O_TRUNC, 0666) = -1 EACCES (Permission denied)
newfstatat(AT_FDCWD, "/.../loadups/build/full.sysout", ..., 0) = -1 ENOENT (No such file or directory)
```

 - Are you using online.interlisp.org? **no**
 - OS: **Linux**
 - OS Version: **Fedora 42 (kernel 6.17.11-200.fc42.x86_64)**
 - Display/window system: **X11**
 - Host arch: **x86_64**
 - Maiko version: **`9259716` ("Merge pull request #544 from Interlisp/fgh_redo-linux-makefiles")**
 - `IL:MAKESYSDATE`: not directly available, since we cannot complete the
   FULL build to inspect a `full.sysout`. The medley repo is at commit
   `6159c64b` ("man [filename] just opens the corresponding documentation
   file (#2572)"). The lisp.sysout that we build full from contains the
   sources at this commit; `LOADUP-FULL.LCOM` was last edited 28-Apr-2026
   per its `Edited` comment.

**Additional context**

* This is **not a permissions/SELinux/AppArmor/fapolicyd issue**. SELinux is
  disabled; AppArmor and fapolicyd are not installed. `loadups/build/` is
  writable (`drwxr-xr-x blake blake`); `touch
  loadups/build/full.sysout-temp` succeeds. The path `/tmp.SYSOUT-temp` is
  what's wrong, not the user's permissions.

* `LOADUP_WORKDIR` and `LDEDESTSYSOUT` are both correctly exported into the
  ldex environment (verified with `strace -v -e execve`).

* The same `(IL:CONCAT (QUOTE {DSK})(IL:UNIX-GETENV(QUOTE LOADUP_WORKDIR))
  (IL:L-CASE (QUOTE /full.dribble)))` form earlier in the same `.cm` script
  resolves correctly — `loadups/build/full.dribble` lands in the right
  place. So the path arg construction itself is fine.

* `MAKESYS` (`sources/ADIR:1094-1124`) does not consult `MAKESYSFILENAME`;
  it passes its `FILE` argument directly to `\COPYSYS`. The second argument
  (`:FULL`) is only used for the HERALD string, not for path resolution.

* `\COPYSYS`'s use of `tmp.SYSOUT` as an intermediate name appears to be
  redundant: Maiko's `vmem_save` (`maiko/src/vmemsave.c:354`) already does
  its own `${name}-temp`/rename dance, and the Lisp-side `(RENAMEFILE
  TEMPNAME TARGETFILE)` at `sources/ADIR:249` adds a second rename on top
  of that. The simplest fix is to drop the `tmp.SYSOUT` indirection on the
  DSK branch:

  ```lisp
  ;; sources/ADIR:228-232, replacement for the DSK clause:
  (DSK (SETQ TEMPNAME TARGETFILE)
       (SETQ VAL (\FLUSHVM TEMPNAME)))
  ```

  This lets `vmem_save` write `<dir>/full.sysout-temp` and rename to
  `<dir>/full.sysout` (the right place), and the Lisp `(RENAMEFILE TEMPNAME
  TARGETFILE)` becomes a same-path no-op. The pre-existing comment at
  `sources/ADIR:216` ("The reason for all this fooling around is because
  \\FLUSHVM doesn't like version numbers") suggests the original
  workaround was for a version-handling quirk that may no longer apply, or
  applies on the UNIX-host SELECTQ branch; that branch should be reviewed
  separately.

* We could not finish narrowing down *why* `TARGETFILE` loses its directory
  inside `\COPYSYS` only on the FULL stage. The `(SETQ FILE (OUTFILEP
  (PACKFILENAME.STRING 'BODY FILE 'BODY "WORK.SYSOUT" 'BODY
  \CONNECTED.DIRECTORY)))` line at `sources/ADIR:222` is the suspect:
  with `\CONNECTED.DIRECTORY` reset to bare `{DSK}` by `ENDLOADUP`, and
  some interaction between `'BODY FILE` parsing and the FULL stage's
  loaded packages, the resulting `FILE` ends up directory-less. Every
  attempt to inject diagnostics into `loadup-full-from-lisp.cm`
  (a `PRINTOUT` of the path right before MAKESYS, an `IL:CNDIR` to restore
  `\CONNECTED.DIRECTORY` between ENDLOADUP and MAKESYS, etc.) caused the
  Lisp environment to drop into BREAK at level 3 or wedge in what looked
  like a tight loop, even with `LOADUP_NOBREAK=1` and `IL:HELPFLAG NIL`.
  Interactive debugging from inside Medley would be needed to pin down
  the exact path-parsing failure; the suggested fix to `\COPYSYS` makes
  that further investigation unnecessary because the temp filename no
  longer depends on `TARGETFILE` carrying a directory.

rmkaplan:
Fixes #2595


nbriggs (approved):
Verified correct `*SYSTEM-EXTERNALFORMAT*` with `LANG=en_US.utf-8` and checked that the underscore is present as ASCII underscore in `LC_LANG` and `LC_ALL` for calls to `unix_getenv()` on the C side.

Issue #2588, opened 2026-04-29
TEDIT: opening the page layout menu on default small document gets NIL is not a number error.


nbriggs:
Open a TEdit window from the background menu - a tiny default-sized TEdit document is displayed.
Middle-button in the title bar to open the Expanded Menu > Page Layout menu.
The result is a "NIL is not a number" fault.

TEDITSYSTEMDATE:  2-Mar-2026 18:32:06

```
TEDIT.DEFAULT.MENUFN/4(debug)BT!
IDIFFERENCE
\TEDIT.FORMATLINE
\TEDIT.LINES.BELOW
\TEDIT.WINDOW.SETUP
\TEDIT.OPENTEXTSTREAM.WINDOW
SI::*UNWIND-PROTECT*
OPENTEXTSTREAM
TEDIT
\TEDIT.MENU.START
\TEDIT.PAGEMENU.START
SI::*UNWIND-PROTECT*
TEDIT.DEFAULT.MENUFNA0001
TEDIT.DEFAULT.MENUFN
```

https://github.com/user-attachments/assets/8c347343-cbd4-42fc-b977-8a9a831ef300

rmkaplan:
Fixed in next Tedit.  An Nway menu that doesn't fit horizontally

rmkaplan:
@nbriggs , can this now be closed?

nbriggs:
It's sort of fixed -- the default window is now larger, and on moderately small windows you can now open a Page Layout menu.   But in testing it,  I found that attempting to open a TEDIT on a shallow window takes a "NIL is not a number error" before it opens fully.  For an extreme test case try:  `(TEDIT NIL (CREATEW 100 100 200 10))` -- so I don't know whether you want to continue under this issue or open a new one.

```
CL::%%NOT-NONCOMPLEX-NUMBER-ERROR
IGREATERP
\TEDIT.SCROLLCH.TOP
\TEDIT.SCROLLCH.BOTTOM
\TEDIT.SCROLL.CARET
SI::*UNWIND-PROTECT*
OPENTEXTSTREAM
TEDIT
```
It's doing `(IGREATERP 1 NIL)`.

rmkaplan:
I get a different issue if I type in your TEDIT expression, a failure inside CREATEW itself.  

Did you mean CREATEREGION?  Or maybe putting the 100... in a quoted list?

But switching to CREATEREGION, I get a different error from TEDIT:

    "Region too small to use as a window"

with the offender showing as

[...]

Issue #2585, opened 2026-04-29
TEDIT: certain actions can allow editing in the Page Layout menu


nbriggs:
If you start with an empty (new) TEdit document, open the Page Layout menu, then click "SHOW" it will note "Please specify the page-type", but then you can position the caret in the Page Layout Menu and insert text within the menu.

rmkaplan:
So the problem is that the button isn't properly protected?  Doesn't seem to matter whether the document is empty or not, probably the button code isn't throwing the TTY away.

nbriggs:
It's protected before you hit the SHOW, it's not protected afterwards.  I noticed it when I was looking at (1) without selecting a page type ("Please specify a page-type"), and (2) when a page type is selected but the document is empty ("No characters are selected").

rmkaplan:
Fixed in next Tedit

rmkaplan:
@nbriggs , can this be closed?

nbriggs:
It's not doing what it used to do (at least I can't get the menu to be editable), but it's also getting confused sometimes about the caret in the Page Layout Menu -- in the attached screen shot see that the caret is in both the Right and Top items...

<img width="482" height="209" alt="Image" src="https://github.com/user-attachments/assets/981f21a8-093d-4261-947b-a2408e9ffcc4" />

nbriggs:
The only *active* cursor is in the Top item -- the bottom one is a remnant of where it previously was.  Clicking again somewhere else can cause it to redraw and remove the remnant.

rmkaplan:
I've struggled with this, I really don't understand why it doesn't take down the cursor before it moves to the new position.  But then again, I don't really understand Tedit's cursor handling in general.  It has its own special code, to deal with the fact that it can blink the cursor at different positions in different split windows.

But I've only seen this screen problem in the menu set up.  I've fiddled around to try to stop it, but it apparently is still there.  And I presume that you don't see the dirt every time you move the input focus, only if the focus shift happens while the bits of the old position are on the screen.

nbriggs:
I think perhaps it's not worth spending more time on right now.  The original problem is fixed.

Issue #2434, opened 2025-12-30
[TEDIT] scrolling interacts badly with shift-select


nbriggs:
Under old TEdit, you could start a shift-select operation, move the mouse into the scroll bar and reposition the source TEdit window, then extend the selection (right mouse) to a point in the scrolled window, and on letting up the shift key it would send the whole selection to the target window.

Under new TEdit, if you move the mouse into the scroll bar you lose the shift-select state -- the dotted underlining of the source selection goes away, and the selection is no longer extensible.

This makes it difficult to shift-select pieces of text that are larger than the source window in a single operation.

There is also some odd interaction that I can't characterize when you use scroll-wheel actions during a shift-selection and extension operation -- I think that the displayed state (dashed underlines etc.) is out-of-sync with the actual state.

rmkaplan:
I also had noticed this a month or so ago, but put it in the come-back-to file.

I think the old Tedit was also screwed up, just in a different way.  There you could end up with highlighting on the screen when in fact it didn't think there was a selection.

I'm not sure what should happen in all of the situations where you can leave the window when there is a secondary selection showing, and you either come back eventually or not, with the same keys down or not.

But I think the flaw in the design, probably from the beginning, is that the secondary selection is bound to a variable in the buttonevent fn, and not stored in the object.  So when it leaves and comes back, it doesn't actually know the selection state.

So probably the first move would be to store the secondary selection (and the associated operation) in the object when it detects that the mouse has left the window, for whatever reason.  Then it will at least have the data to decide what to do the next time there is a mouse click in the window--take it down, extend it, do the operation.

rmkaplan:
I've made some progress on this, but won't do a PR until the Hardcopy PR is approved--I don't want to get lost in a merger maze.

[...]

Issue #2286, opened 2025-09-10
HCFILES failures


rmkaplan:
I began to look at the failures in the dribble file.

I took the first file from the lines
Starting on {MEDLEY}<CLTL2>AERROR.;1 :
PDF printing to {MEDLEY}<CLTL2>AERROR.pdf...IL:FAIL
DONE

I did (TEDIT "{MEDLEY}<CLTL2>AERROR.;1"), meta-P to bring up the print-to-file, chose AERROR.PDF, and the pdf file was created with no error.

However, that's not the way that HCFILES deals with a Lisp source file.  If a file is a lisp source file and PRETTY is in PHASES, it executes

(NLSETQ (CL:WITH-OPEN-STREAM (STR (OPEN-PDF-STREAM DEST))
                                        (PRETTYFILEINDEX SRCPATH NIL STR)))

And that's where it fails.  If I take off the NLSETQ, I get the message
    "External symbol FLOATING-POINT-UNDERFLOW not found in package LISP"
 underneath PRETTYFILEINDEX.  And of course the same error shows up if you SEE that file from the exec.

For starters, it would be helpful if the FAIL message printed out the phase, so we know what to look for.  And then, if there is an error in the PRETTY phase, a fallback heuristic would be to try it without the PRETTYFILEINDEX, maybe by calling TEXTOIMAGEFILE or COPY.TEXT.TO.lMAGE.

rmkaplan:
Another suggestion re the dribble:  Print out the line with the destination file after it has succeeded, with the DONE.  Let the FAIL be associated with the input file, and if fails, don't print out the destination.

masinter:
I (re)discovered how to test branches without merging into master:
run Action run_HCFILES specifying the branch you want to test (lmm-diagnose-hcfiles) and set it to produce a draft release. The result of this will be (if you ssh oio) in
/srv/oio/files/development/medley 

there's a url for accessing draft hdfiles output but I don't remember and can't find.

the diagnnose branch adds a backtrace when there is an error.

masinter:
https://files.interlisp.org:8081/medley/loadups/hcfiles.dribble
no need to run the test HCFILES again -- the log from my run is here:

and attached

[hcfiles.dribble.txt](https://github.com/user-attachments/files/27447144/hcfiles.dribble.txt)

rmkaplan:
On internal/mesatypes.tedit...

[...]

PR #2597, opened 2026-05-08
Cleanup the piece datatype, address a few issues

rmkaplan:
This has a bunch of files (as usual) but mostly just propagating a cleanup of the PIECE datatype.  It had several fields that were simple combinations of other fields but required extra code to maintain consistency.

This includes minor fixes for  #2588 #2585 and also the change to TEDIT.TO.IMAGEFILE mentioned in #2286, for HCFILES, and the fix for line overflow mentioned there.

This also has some initial code changes to manage what happens when the mouse leaves the window while a secondary selection is up #2434 and then possibly returns.  The behavior here is still suboptimal, but in a different way than before


pamoroso:
It's working with no issues on Linux Mint 22.1 Cinnamon.

library/tedit/TEDIT-BUTTONS
```diff
@@ -922,11 +922,12 @@
...
 (MB.NWAY.SIZEFN
-  [LAMBDA (OBJ STREAM CURX RIGHTMARGIN)                      (* ; "Edited 20-Aug-2024 15:12 by rmk")
+  [LAMBDA (OBJ STREAM CURX RIGHTMARGIN)                      (* ; "Edited 29-Apr-2026 17:56 by rmk")
+                                                             (* ; "Edited 20-Aug-2024 15:12 by rmk")
                                                              (* ; "Edited 22-Jul-2024 11:31 by rmk")
                                                              (* jds " 6-Sep-84 14:19")
                                                              (* ; "Tell the size of an n-way menu")
-    (OR (IMAGEOBJPROP OBJ 'BOUNDBOX)
+    (OR (AND NIL (IMAGEOBJPROP OBJ 'BOUNDBOX))
         (LET ((OLDBOX (IMAGEOBJPROP OBJ 'BOUNDBOX))
```
masinter, line 932: the AND NIL comments out the code... worthy of a comment as to why and how it interacts with other parts. Just saying... what does "Tell the size of a n-way menu" comment apply to?

masinter (commented):
I dont know enough about TEdit internals to review the code sources.
In a finite amount of time.
How would you recommend we review this?

pamoroso:
Nothing unusual to report at commit f38e216.

masinter (approved):
@pamoroso tested and testing more unlikely so I will approve: LGTM

rmkaplan:
Tedit is tough because there are too many interacting components (model/view/controller), and also a lot of edge cases.  I've been working to modularize, put it's a slow process.

So random probes of the code may turn up some things (like the left-over AND NIL), but mostly does it work at all and do the claimed bug fixes appear to address the reporting issues.

Issue #2565, opened 2026-04-15
Delete unused sources files


masinter:
These files (and their LCOMs) were proposed to be moved to obsolete in PR #2056.

The task here is to double check that these files are really unused, and then delete (or move) to Interlisp/history repo.

CMLSEQ
CMLWALK
COMMENT
DFILE
DISKDLION
INSPECT-CLOSURE
JAPANESE
MACROAUX-OPTIMIZERS
MOD44IO
SEDIT-CONVERT
SEDIT-DEBUG

DEFPACKAGE-IMPORT

were the files originally proposed

rmkaplan:
What about sources/TCPHTE.  Also lispusers/GENERIC-INIT and VANILLA-INIT--undocumented, complex.

We should look more carefully at SEDIT-DEBUG.

nbriggs:
SEDIT-DEBUG should stay -- you don't use it normally, but if you're trying to debug something in SEDIT it provides a bunch of convenience functions.  SEDIT-CONVERT also contains a utility function that most people wouldn't use, but when you need it you need it.

nbriggs:
sources/TCPHTE is for reading HOSTS.TXT (aka /etc/hosts) files.  We probably want to keep that.
GENERIC-INIT and VANILLA-INIT are really good init files that Stan Lanning wrote - excellent examples that should be kept around.

masinter:
https://github.com/Interlisp/medley/issues/2565#issuecomment-4256346576
move SEDIT-DEBUG to Libary? Or Internal?

TCPHTE belongs in Library? 

GENERIC-INIT and VANILLA-INIT belong in greetfiles?

PR #2598, opened 2026-05-08
add BAKTRACE to HCFILES fails. Skip HC of CLTL2 (package errors) and dinfo (not really tedit).

masinter:


pamoroso:
On Linux Mint 22.1 Cinnamon HCFILES seems to run to completion and skip the CLTL2 directory as expected.

```
paolo@lispmachine:~/medley/medley/scripts$ ./do_hcfiles.sh 
"/home/paolo/bin/lde" "/home/paolo/medley/medley/loadups/apps.sysout" -id "hcfiles_1" -title "Medley::hcfiles_1" -g 1024x768 -sc 1024x768 -noscroll
LOGINDIR: "/tmp/hcfiles-75506"
REM.CM FILE: "/tmp/hcfiles-75506/hcfiles.cm"
VMEM FILE: "/tmp/hcfiles-75506/vmem/lisp_hcfiles_1.virtualmem"
Added hcfiles.dribble to /home/paolo/medley/medley/loadups
Added hcfiles-fails.txt to /home/paolo/medley/medley/loadups
```

Here are the dribble files:

* [hcfiles.dribble.txt](https://github.com/user-attachments/files/27533733/hcfiles.dribble.txt)
* [hcfiles-fails.txt](https://github.com/user-attachments/files/27533737/hcfiles-fails.txt)

rmkaplan:
sources/mesatypes is just the txt file.  Did you mean a lispsource file?

pamoroso:
It still works with no issues at commit 5e88d17.

masinter:
@rmkaplan I'm not sure what you're looking at but the result of this PR is that there is a MESATYPES, MESATYPES.TXT, MESATYPES.LCOM in the {MEDLEY}<sources>, where MESATYPES is the lisp source and MESATYPES.TXT is the documentation for it, and there are no files MESATYPES* (of either case) in {MEDLEY}<internal>.

rmkaplan:
I think it's a problem in GITFNS:  When I "See" MESATYPES it shows the .txt file instead.

So it isn't your PR.

rmkaplan:
But if MESATYPES is going to be moved from internal/ into sources/ and perhaps become part of a release, the documentation needs to be upgraded.  At least with one or two examples of what one of these new declarations might look like.  Otherwise it really is internal.

masinter:
MESATYPES is loaded by other files in sources which seems like a more decidable criterion.--LarryMasinter.netOn May 10, 2026, at 11:38 AM, rmkaplan ***@***.***> wrote:﻿rmkaplan left a comment (Interlisp/medley#2598)
But if MESATYPES is going to be moved from internal/ into sources/ and perhaps become part of a release, the documentation needs to be upgraded.  At least with one or two examples of what one of these new declarations might look like.  Otherwise it really is internal.

rmkaplan:
If there are examples somewhere, maybe some of them should be copied into the documentation as illustrations of what functionality this provides.

nbriggs:
MESATYPES is used by DOVEDECLS, DOVEETHERDECLS and LOCALFILE to access the mesa records (with mesa type declarations) in interfaces exported by the mesa code running on the I/O processor and the file system structures.

We include LOCALFILE in LOADUP-LISP, so indirectly we use MESATYPES.

There are examples of MESATYPE and MESARECORD declarations in the source files that use it.

masinter:
if we do #2565 then there's a good chance MESATYPES will also become unused and can be deleted or moved.

for now, though, without editing the content, the rename to sources seems best.

masinter:
https://files.interlisp.org:8081/medley/loadups/hcfiles.dribble'
I'd guess is the best so far, 90 FAILs for maybe 12 different reasons.

rmkaplan:
One of the Tedit ones was a bug in reading Tedit file-format 0 (from early 1985).  I'll put the fix in a separate mini-Tedit PR I'm about to push.

A bunch of others were font-not-found for size 13.  Those are probably 10/13 swaps.  Some are missing image object fns.  And then there's the rest

I think the dribble file should be opened as UTF-8.  You might be able to specify that by passing an open UTF-8 stream to DRIBBLE:
  (DRIBBLE (OPENSTREAM 'foo.dribble 'OUTPUT 'NEW '((FORMAT :UTF-8))))

pamoroso:
Still looking good at commit a08e4b8.

pamoroso (approved):
Looks good to me.

PR #2599, opened 2026-05-09
DATABASEFNS failed when writing .DATABASE file for the first time if the source file name wasn't U-CASE.

MattHeffron:
DATABASEFNS failed when writing .DATABASE file for the first time if the source file name wasn't U-CASE.
It didn't find any defined FNS or FUNCTIONS on the file, so it thought there was no need for a DATABASE file.


pamoroso:
I'm not sure it's working as intended or I may be misunderstanding how it's supposed to work. In a first Medley session with DATABASEFNS loaded I defined a square function `SQ` and assigned it to file `sq` which I saved:

```
4_ (FILESLOAD DATABASEFNS)

{DSK}<home>paolo>medley>medley>library>DATABASEFNS.LCOM;1
compiled on  8-May-2026 22:09:12
File created  8-May-2026 22:08:38

(LOAD redefined)
(LOADFROM redefined)
(MAKEFILE redefined)
DATABASEFNSCOMS
(LINKEDFNS reset)
(MAKEFILEFORMS reset)
({DSK}<home>paolo>medley>medley>library>DATABASEFNS.LCOM;1)
5_ (DF SQ)
SQ has no definition of any of these types:
  (FUNCTIONS FNS)
Select a type of dummy definition to install.
Select a definer to use for a dummy definition.
SQ
6_ (SQ 3)
9
7_ (FILES?)
the functions: SQ
...to be dumped.  want to say where the above go ? Yes
(functions)
SQ  File name: sq
create new file SQ ? yes
8_ (MAKEFILE 'sq)
{DSK}<home>paolo>il>sq.;1
```

In a second session I loaded the file, edited `SQ` to add a comment, and saved the file again:

```
4_ (FILESLOAD DATABASEFNS)

{DSK}<home>paolo>medley>medley>library>DATABASEFNS.LCOM;1
compiled on  8-May-2026 22:09:12
File created  8-May-2026 22:08:38

(LOAD redefined)
(LOADFROM redefined)
(MAKEFILE redefined)
DATABASEFNSCOMS
(LINKEDFNS reset)
(MAKEFILEFORMS reset)
({DSK}<home>paolo>medley>medley>library>DATABASEFNS.LCOM;1)
5_ (LOAD 'sq)

{DSK}<home>paolo>il>sq.;1
File created  9-May-2026 11:18:31
SQCOMS
{DSK}<home>paolo>il>sq.;1
6_ (DF 'SQ)
Editing FNS definition of SQ.
SQ
7_ (FILES?)
To be dumped:
SQ ...changes to FNS: SQ
8_ (MAKEFILE 'sq)
{DSK}<home>paolo>il>sq.;2
```

But I can't find any relevant .DATABASE file.

rmkaplan:
According to the documentation, this is controlled by the setting of some variables LOADDBFLG and SAVEDBFLG, which maybe you have to set in your init.  But it says that the default values are Ask, so it should ask you the first time you make the file whether you want a database to be maintained (saved, loaded, saved again) for that file.

pamoroso:
I tried setting `LOADDBFLG` and `SAVEDBFLG` but still no database is saved.

pamoroso:
At commit 26bba9e DATABASEFNS does ask whether to create a database, which I agree, but I can't find the relevant database anywhere on my system. The first session:

databasefns1

The second session:

databasefns2

rmkaplan:
You shouldn't have to give an explicit ANALYZE command once you say Yes to the question.  It should then write the file ...il>SQ, analyze the functions on SQ, and also write ...il>SQ.DATABASE.

And then reanalyze and remake the database everytime you remake the file.

pamoroso:
Skipping the call to `ANALYZE` doesn't make a difference. Although DATABASEFNS still asks whether to save the database no such database file is created.

rmkaplan:
I poked around.

It seems that the function MAKEDB is called under MAKEFILE, it is asking the question, and it is storing the answer on the property list of the rootfilename.

But when the MAKEFILE is complete, the property is gone.  And maybe it is gone somewhere before it would actually try to do the DUMPDB call.  the property is set by an undoable /PUT, maybe somewhere in the sequence it is getting undone?

rmkaplan:
There was a glitch in the function DUMPDB, it was using FL instead of FILE in a PROG instead of a PROG*.  I reverted that one to FILE, seems to work.

rmkaplan:
But maybe this is a case where the MAKEFILE etc. here should be renamed as MAKEFILE.DBFNS, say, with MOVD?

MattHeffron:
> I reverted that one to FILE, seems to work.

That works only if the `FILE` is uppercase. The better change is `PROG` => `PROG*`.
I just committed that.

nbriggs:
Watch out for files where the COMS variable is lowercase...
```
lispusers/ACE/RDHDEMO.ACE:(PRETTYCOMPRINT rdhdemoCOMS)
lispusers/ACE/RDHDEMO.ACE:(RPAQQ rdhdemoCOMS ((UGLYVARS ACE.CURRENT.SEQUENCE)))
```

MattHeffron:
> Watch out for files where the COMS variable is lowercase

@nbriggs  `DATABASEFNS` just uses the `ROOTFILENAME` (always uppercase) to get the `FILEDATES` property, and to set/check the hash table of loaded Masterscope `.DATABASE` files. All other handling of `COMS` names, etc., possibly mixed case, would be in `source/FILEPKG`.

masinter:
i would really like to make a medley lisp managed file source extension but it's a hard problem -- I think it is necessary to be more explicit about the destinction between file package name (fpkgname) and file system name.

To turn a file system name into a file package name you can use ROOTFILENAME. 
The standard file system name has an extension -- .lisp for common lisp files and .ilsp or .medley or .mlsp or ... that gets added to turn a file package name into a file system name.

do we eliminiate the lcom dfasl two file extensions for compiled -- just have one?

maybe this is a different discussion

rmkaplan:
It's a different discussion, but...

Suppose we

* set a variable LISPSOURCEEXTENSIONS to (ILISP NIL).
* rename (e.g. shell mv) some/all source files FOO to FOO.ILISP.  Don't load/make, just rename.
* replace the FINDFILE's in LOAD and related functions with FINDFILE-WITH-EXTENSIONS
* fix MAKEFILE to pack on (CAR LISPSOURCEEXTENSIONS).

What could go wrong?

nbriggs:
@rmkaplan -- I'm good with that up to the "fix MAKEFILE to pack on (CAR LISPSOURCEEXTENSIONS)".

If I say to `(MAKEFILE 'FOO)` with no extension I don't want it "helping" me do something I didn't want.  If I think the extension should be there then I'll `(MAKEFILE 'FOO.LISP)`, with the extension I want (which, BTW,  isn't "ILISP").  This is what I do for my INIT.LISP file.  It all works fine.

rmkaplan:
That's a variation of the same dot problem I raised with respect to Tedit.

In the world we would hypothetically be migrating to, the default when an extension isn't specified for a source file would be ILISP (or MLISP or whatever we standardize on).  When you are operating on a file as a source file and you don't specify the extension, that's what it should try first.  Just like when you are opening a file with TEDIT and you don't specify an extension, it should guess one of the Tedit extensions first.

On this view, there would be a difference between (MAKEFILE 'FOO) and (MAKEFILE  'FOO.).  The intent expressed in the second one is that you want the empty extension.  The intent of the first one is that you want the (new) default extension for source files (because MAKEFILE is a sourcefile operation).  (LOAD 'FOO.) would explicitly look for the file without an extension, (LOAD 'FOO) would look first for FOO.MLISP and then fall back to FOO (if NIL is in the extension list).

pamoroso:
With the commits up to 8365cac DATABASEFNS now creates and loads the database file.

rmkaplan (approved):
lgtm

PR #2600, opened 2026-05-09
Make (ARGLIST fn T) and (SMARTARGLIST fn T) a bit "smarter"

MattHeffron:
Make `(ARGLIST fn T)` and `(SMARTARGLIST fn T)` a bit "smarter" with the atomic name of the arg for a `LAMBDA` nospread with a `CCODEP` definition.

_E.g._, before `(ARGLIST 'FONTCOPY T)` returned `U`; now it returns `FONTSPECS` as in the `CCODEP`. Likewise for `SMARTARGLIST`.

I didn't test if `NLAMBDA` nospread would follow the same conditions to be able to extend this edit to handle that case as well.


pamoroso:
On Linux Mint 22.1 Cinnamon it's working as intended.

MattHeffron:
I just checked, and this works "correctly" for NLAMBDA no-spread FNS.

masinter (approved):
i wish i understood all the uses of ARGLIST and SMARTARGLIST but I don/t. The current behavior seems wrong and this likely looks like a good fix, so... "looks good to me"

PR #2607, opened 2026-05-11
Align HCFILES version of DOC-OBJECTS with minor changes I had stashed away

rmkaplan:


pamoroso:
It still works with no issues at commit d441397.

MattHeffron:
How is this related to HCFILES?
It's just a LispUsers package.

rmkaplan:
Larry made some changes in the HCFILES PR, I copied those changes into the version that I was working on.  This just eliminates future incompatibility.

nbriggs (approved):
LGTM.  I would suggest a better comment on merge -- I think this is what you're doing here:

  - DOCOBJ-GET-LOOKS: update to use new TEdit internal functions for extracting "looks"
  - DOCOBJ-INCLUDE-EDIT: restructure EDIT.FILE case "for" expression to be clearer

PR #2621, opened 2026-05-20
Update buildLoadup.yml workflow to account for deprecation of node20 in github runners.

fghalasz:
Routine maintenance on github action


pamoroso:
Nothing unusual to report at commit e6567ab on Linux Mint 22.1 Cinnamon.

nbriggs (changes requested):
Looks good as far as it goes, but doHCFILES.yml and buildDocker.yml also use action/checkout@v4 and they might as well be updated in this PR.

nbriggs (approved):
LGTM.

Issue #2622, opened 2026-05-21
Font not found error when opening DInfo in Medley Online


pamoroso:
When I open DInfo in the latest Medley Online (20-May-2026 03:25:22) I get the error:

```
INTERLISP-ERROR
In ERROR:
FONT-NOT-FOUND
(MODERN 18 (BOLD REGULAR REGULAR) 0 DISPLAY)
```

<img width="1920" height="1040" alt="Image" src="https://github.com/user-attachments/assets/dade0f12-f0f8-4232-8916-8accb1c9558f" />

The backtrace:

```
OLDMOUSE/6(debug)BTV
FONTCREATE
   FILE 
#<Input Stream on {DSK}<usr>local>interlisp>medley>docs>dinfo>IRMTOP.TEDIT;1/166,101100>
   TEXTOBJ {TEXTOBJ}#121,140600
   LOOKS {CL101/16254:Gacha10}
   FILEPOS 1826
   LOOKSLEN 20
   FONT NIL
   NAME MODERN
   SIZE 18
   SUPER 0
   STYLESTR NIL
   BOLD T
   ITALIC NIL
   BITS 512
   PROPS NIL
   PROPS ((USERINFO) (COLOR . BLACK))
\TEDIT.GET.SINGLE.CHARLOOKS
   FILE 
#<Input Stream on {DSK}<usr>local>interlisp>medley>docs>dinfo>IRMTOP.TEDIT;1/166,101100>
   TEXTOBJ {TEXTOBJ}#121,140600
   I 4
\TEDIT.GET.CHARLOOKS.LIST
   TEXT 
#<Input Stream on {DSK}<usr>local>interlisp>medley>docs>dinfo>IRMTOP.TEDIT;1/166,101100>
   TSTREAM #<IO Tedit Stream/127,65100>
   PCCOUNT 77
   CURTEXTBYTE# 0
   END 2635
   PCNO 0
   TEXTOBJ {TEXTOBJ}#121,140600
   ORIGBYTE# 0
   PC NIL
   BYTELEN 0
   PREVPC {PIECE}#121,134660
   FIRSTPC {PIECE}#121,134660
   PARALOOKSMAP {ARRAYP}#136,76220
   CHARLOOKSMAP NIL
   DEFAULTCHARLOOKS {CL101/16326:Gacha10}
   OLDPARALOOKS {PL81/48584:LE-0-0}
\TEDIT.GET.PIECES3
   TEXT 
[...]

PR #2623, opened 2026-05-21
Reset {MEDLEY} pseudohost on restart

rmkaplan:
(PSEUDHOST 'MEDLEY (UNIX-GETENV "MEDLEYDIR")) updated entry on AROUNDEXITFNS


nbriggs:
Tested by copying a `lisp.virtualmem` between hosts and observing that the `{MEDLEY}` pseudohost gets reset to the correct `MEDLEYDIR` location.

nbriggs (approved):
Yes, this solves the underlying problem that I believe caused #2622

rmkaplan:
You might also check to make sure that the lisp variable MEDLEYDIR is also being set.  That may also be mixed up because of the preconditions in MEDLEY-INIT-VARS,  independent of the new pseudohost definition.

nbriggs:
Before any changes: MEDLEYDIR is correct in online.interlisp.org, though the MEDLEY pseudohost is not.

sources/MEDLEYDIR
```diff
@@ -99,7 +100,7 @@
                                                                 (* ;; 
                                        "Any old values, restore them, substituting the new MEDLEYDIR")
 
-             (PSEUDOHOST 'MEDLEY MEDLEYDIR)
+             (PSEUDOHOST 'MEDLEY (UNIX-GETENV "MEDLEYDIR"))
              (PROG (OLDMD NEWMD SAME TMP)
```
masinter, line 104: PSEUDOHOST takes a Medley-style path string, can be called with {DSK} or {UNIX} or an XNS server name as the HOST/DEVICE.

But UNIX-GETENV("MEDLEYDIR") is (if set) a SLASHIT T string. It defaults to {DSK} but this is a type mismatch.

nbriggs: Yes, that's a concern - though as it stands it will work on online.interlisp.org, so worth replacing the broken current system.

PR #2634, opened 2026-06-02
Consistent treatment of Tedit secondary selections--take 2

rmkaplan:
The secondary selection is contained in the textobj to enable consistency between the data and the screen.  Selections now also are marked with their "operation", and operation PENDINGDEL of the selection replaces BLUEPENDINGDELETE as a field of the textobj.


rmkaplan:
This is marked as a draft.  There are many possible interactions between the mouse location (in or out of the window), mouse buttons, modekeys, processes, and the display.  I've tested the obvious pathways, but I have probably missed some

pamoroso:
I haven't noticed anything unusual in my initial tests on Linux Mint 22.1 Cinnamon.

pamoroso:
Still looking good at commit d06bec7.

nbriggs:
I still seem to be able to get it into a weird state:

image

nbriggs:
You can't see it in the snap, but the caret was blinking at the beginning of the line below the blue-pending-delete line. No mouse buttons were down at that point.

rmkaplan:
Do you know/remember the sequence?

pamoroso:
The only unusual thing I've noticed at commit 4439470 is deleting a line with selecting the line in the left area of TEdit with `Ctrl` pressed, moving outside the window, moving back inside, and releasing `Ctrl`: the text remained selected in reverse and didn't disappear. I can't reproduce it though.

nbriggs:
No, but I've got it sitting here and can inspect things or save the sysout and give it to you if it will retain enough state.

rmkaplan:
Or, more to the point, what do you expect to see?  The bluependingdelete would be the state of the primary selection (because of a previous right-button extension), the underline would be the source of a copy.  If you have the copy key down in the window (or have moved it outside while the copy was down) when you took the snap, that would be OK.  If you release the copy key with the mouse in the window, then the black should be replaced by the underline.

nbriggs:
I can move the caret around, no keys are down (no shift, no mouse), and I expect the secondary selection to have gone away because I believe I have no operation in progress.

rmkaplan:
OK, I think I see the same or related problem:  after moving out with the ctl down (which only applies to the primary selection), it doesn't actually do the delete when the mouse comes back in with the ctrl down and then the ctrl is released.

This is not the same as bluependingdelete (by rightbutton extension), which sets up a long-term state.  This is in the middle of the delete operation.

nbriggs:
Do you want my gzipped lisp.virtualmem ?

rmkaplan:
When you say you move the caret around with no keys down, is that inside the window or outside?  If inside the window with nokeys down, then you shouldn't see the black if it was the result of CTRL selecting vs right-button extension?  The deletion should already have happened.

rmkaplan:
Sure, might as well send the sysout, although it probably doesn't represent the event sequence

nbriggs:
[tedit-lisp.virtualmem.gz](https://github.com/user-attachments/files/28724684/tedit-lisp.virtualmem.gz)
I was selecting a point and turning it inverse video with a two-finger click on the trackpad at a different point.  That's a right-button, I guess.

rmkaplan:
This latest commit deals with what happens with a simple on-the-fly DELETE/CTRL selection, when the mouse moves in and out.  That's the main odd behavior that I have seen, and I think that is what @pamoroso reported.  A delete selection is not a secondary selection, it is the primary.  And it is also not a bluependingdelete selection, the difference being that the delete operation takes place when you left the CTRL key, while the PENDINGDEL  doesn't take effect until you type something.

The behavior I am trying for is that the selection turns black when you click with any button while the CTL is down.  When you lift CTRL with the mouse in the window, the selection is deleted**.  When you move the mouse out, the selection stays black.  If you move it back in with the CTL down (even if it was up for awhile outside the window), then you are still in the delete mode, and lifting will perform the operation.  If you move back in with the CTL up, then the highlighting goes to underline to reflect the fact that the operation is canceled.

(**I think the operation is also supposed to cancel if CTL goes up when the caret is in the window but at least one mouse button is still down.  That's not yet there.)

The PENDINGDEL hangs off the right button (2 fingers on a pad).  My sense is that the behavior that has been reported in that mode is OK.

Please fiddle with this, I hope it's better.

rmkaplan:
Another commit:  An attempt at getting the within-window cancel to work

pamoroso:
So far nothing unexpected at commit a550260.

pamoroso:
Still looking good at commit e9ca197.

pamoroso:
At commit 771e739 the caret non longer flashes in `READONLY` documents.

pamoroso:
Nothing unusual to report at commit 6b1e623.

pamoroso:
Nothing unusual to report at commit a391630.

rmkaplan:
Quantizing the marginbar display revealed some inconsistencies in the coordinate system of the image object versus the pica scale for margins and tabs.  Now better aligned, and quantizing at .5 picas.

pamoroso:
Still looking good at commit ef348e5.

pamoroso (approved):
It works with no issues in my testing.

Issue #2635, opened 2026-06-03
Tedit failures in HCFILES


rmkaplan:
I have done some initial probing to try to see what's going on.

In the HC-FAILS file, there are 64 entries for TEDIT.TO.IMAGEFILE failures.

Of those, 41 were font failures:

     11 were failures in trying to find size 1 postscript fonts for Modern and Gacha.

     30 were failures in trying to find a display font.
            25 of those were  in Notecards>docs/user-guide-v1.2>  files looking for nonexistent size 13.
              5 were not notecards and not size 13 (Helvetica 6 and Timesroman 24).

The 13's are likely 10/13 smashes, but it is still surprising that it needs display fonts for hardcopy.
(The Helvetica 6 and Timesroman 24 also don't exist--probably there should be a table entry that allows coercion to something close)

nbriggs:
```
./medleydisplayfonts/TIMESROMAND30-MRR.MEDLEYDISPLAYFONT
./medleydisplayfonts/TIMESROMAND72-MRR.MEDLEYDISPLAYFONT
./medleydisplayfonts/TIMESROMAND24-MRR.MEDLEYDISPLAYFONT
./medleydisplayfonts/TIMESROMAND36-MRR.MEDLEYDISPLAYFONT
./displayfonts/c0/TIMESROMAND36-MRR-C0.DISPLAYFONT
./displayfonts/c0/TIMESROMAND30-MRR-C0.DISPLAYFONT
./displayfonts/c0/TIMESROMAND72-MRR-C0.DISPLAYFONT
./displayfonts/c0/TIMESROMAND24-MRR-C0.DISPLAYFONT
```
You use "display" style fonts instead of body fonts for headlines, posters, etc. at large sizes.  24pts and up here are TimesRomanDisplay.

nbriggs:
I guess the Alto fonts, and by inheritance the Interlisp-D fonts, didn't have a separate field in the face for "display" so it got folded into the name instead

 The NY Public Library tells us that:
> Times Roman is an everyday text font designed for small, narrow-column body text (10–12 points), while Times Display is designed for large headlines (usually 18+ points). Display fonts feature exaggerated stroke contrast and tighter letter spacing, whereas text fonts have thicker, sturdier strokes to remain legible at small sizes.

nbriggs:
Helvetica 6 exists as an Interpress font, but not as a screen font.  It should probably go to Modern 6.

rmkaplan:
[...]

PR #2637, opened 2026-06-04
Rmk193  fix some hcfiles font errors

rmkaplan:
The postscript failures in #2635 were the result of an argument mismatch between the POSTSCRIPTSTREAM call \SEARCHFONTFILES and the expectations of that function.

This PR also provides  a brute-force patch for the DISPLAY failures for HELVETICA 6 and TIMESROMAN 24.  Basicallly, asserting that TIMESROMAN can be coerced to CLASSIC, and that HELVETICA 6 can be coerced to HELVETICA 7.

Still don't know why it needs to create those display fonts when it is creating the image file, but if it asks for them, it will get a good approximation.

I tried coercing the size 13 fonts to 12, but that didn't help--it got an array error saying that an index 13 was out of bounds in a situation where 10 was perfectly fine.  That reinforces the belief that there was a 10/13 swap in those notecards files--maybe byte swapping would fix those.


pamoroso:
On Linux Mint 22.1 Cinnamon I ran do_hcfiles.sh locally and, unlike files.interlisp.org, the loops and notecards directories are missing altogether. This is the log file: [hcfiles-fails.txt](https://github.com/user-attachments/files/28589614/hcfiles-fails.txt)

pamoroso:
At commit 7ee6c72 the directories loops and notecards are still missing. The fails file: [hcfiles-fails.txt](https://github.com/user-attachments/files/28604458/hcfiles-fails.txt)

nbriggs:
notecards and loops are separate git repositories that you can check out so their directories are siblings of the medley directory - then the hcfiles script should find them.

rmkaplan:
In the failure file produced by the script, the notecards and loops files show up as subdirectories of {MEDLEY}.  But they aren't in the same repo, and at least in my set up I have them as sister directories of {MEDLEY} (with pseudohosts {GLOOPS} and {GNOTECARDS} that GITFNS creates if it can find them).

pamoroso:
The directories loops and notecards are actually already siblings of the medley directory, so I wonder why do_hcfiles.sh doesn't detect them:

```
paolo@lispmachine:~/medley$ ls
DOS  envos  loops  maiko  medley  new-gabriel  notecards
```

pamoroso:
At commit a617708 I still don't get the directories loops and notecards. The fails file: [hcfiles-fails.txt](https://github.com/user-attachments/files/28630978/hcfiles-fails.txt)

rmkaplan:
I assume you also don't get those in master

masinter:
PR #2532 was an attempt to reduce some of the confusion around the sub-module / sister repository.  Right now, we have things that are created by scripts in the medley repository that refer to sources _not_ in the medley repository.  in particular, the HCFILES script copies these "ordinarily sister" repositories to be under {medley}.

I marked #2532 as draft with the idea of doing one sub-module repository at a time.
This will lead to an interim state where the module structure is half-done. Is that OK with you? @rmkaplan @pamoroso

pamoroso:
@rmkaplan Yes, the loops and notecards directories are missing also in master. The fail file of master: [hcfiles-fails.txt](https://github.com/user-attachments/files/28646179/hcfiles-fails.txt)

@masinter I don't know, so far I checked out submodules only lightly and find the feature confusing.

rmkaplan:
I'm confused.  It appears that the font 13 files in the bigger hcfiles-fails.txt are in the user-guide-v1.2> from_envos subdirectory, but there are versions of those files above in the user-guide-v1.2.  And the versions higher up have already had the CR/LF swap (with a little fixing shell script that is stored at the docs/ level).

So, why is HCFILES looking at the from_envos subdirectory?  Maybe that directory should be moved to obsolete so we don't waste time on it.

rmkaplan:
If GITFNS is loaded, it looks around for loops and notecards and tries to create pseudohosts {GNOTECARDS} and {GLOOPS} that point to them.

I don't know if its current search covers both the sister and subdirectory possibiliites.  I have the sister configuration, and that works.  The subdirectory needs to be tested.

If we verify that it covers both configurations, then HCFILES could use {GNOTECARDS} and {GLOOPS} and the paths listed in the HCFAILS file would be good in both environments.

nbriggs (approved):
Looks good.

rmkaplan:
Did you catch it after the second commit that removed the MEDLEY-INIT-VARS entries?

nbriggs:
Yes.  It would have shown as has having been changed since I approved it and would require reapproval if I'd jumped the gun on that.

PR #2638, opened 2026-06-05
Finish updating all Medley Github actions to account for deprecation of Node 20 on github runners

fghalasz:
Updated or replaced remainder of calls to marketplace github actions to account for deprecation of Node 20 on github runners.  For most just updated version to latest released version.  For cb80/delrel, replaced with a shell script that deleted a tagged release using the gh cli (since cb80/delrel appears no longer active and is not being upgraded past Node 20).


nbriggs (approved):
LGTM.

PR #2639, opened 2026-06-05
Prevent BITMAPCREATE.BIGBM from causing stack overflow due to improper handling of BITSPERPIXEL.

### LINKED ISSUE #2613: BITMAPCREATE can stack overflow for large bitmaps with BITSPERPIXEL > 1
**Describe the bug**
I tried `BITMAPCREATE` of a large bitmap with `BITSPERPIXEL` of 24. It entered RAID with a stack overflow.

**To Reproduce**
1. Running a full.sysout. 
2. In an Interlisp EXEC, enter `(BITMAPCREATE 1920 1080 24)`
3. Stack overflow, entered RAID

**Expected behavior**
An error that the bitmap was too large, or return a BIGBM.

 - Are you using online.interlisp.org? no
 - OS: [e.g. macOS/Linux/Cygwin] Linux
 - OS Version: Ubuntu 24.04 (WSL)
 - Display/window system: X11
 - Host arch: x86_64
 - Maiko version: [e.g. commit ID from `git log | head`]
 - MEDLEY commit id: b99630d6

**Additional context**
This depends on BIGBITMAPS being loaded (part of full.sysout).
The problem _appears_ to be that `BITMAPCREATE.BIGBM` doesn't account for `BITSPERPIXEL` when calculating the height of a _slice_. It calls `BITMAPCREATE` to make the slice, but _that_ function says the slice is still too big, so it calls `BITMAPCREATE.BIGBM` again to create the `BIGBM`, which believes (incorrectly) that it isn't too big, and recurses to stack overflow.

MattHeffron:
Prevent BITMAPCREATE.BIGBM from causing stack overflow due to improper handling of BITSPERPIXEL.

`(IL:BITMAPCREATE 1920 1080 24)` no longer fails.

Resolves #2613

Also added MAKEFILE-ENVIRONMENT and FILETYPE PROPS to COMS.


pamoroso:
I tested it on on Linux Mint 22.1 Cinnamon and confirm it resolves the issue.

nbriggs:
@pamoroso try testing it with the widest bitmap that it'll accept too.

pamoroso:
I get an array full error at 6000x4000@24 but it works fine at 4000x2000@24:

bitmapcreate

nbriggs:
@pamoroso That's expected -- but try widest, and not particularly tall.  Say, just 34 rows or so - things will go subtly wrong if it can't get at least 16 rows into a strip.  You won't get an error, it's just unlikely that textures will appear correctly in the bitmap.

pamoroso:
I get no errors for `(BITMAPCREATE 6000 34 24)` and `(BITMAPCREATE 6000 34 15)` but, as you noted, the textures may not be as expected.

nbriggs:
Hmmm... the implementation is quite incomplete.  It should follow the IRM documentation for BITBLT, and it doesn't.

nbriggs:
If the fix here can result in component bitmaps that are less than 16 rows then there are *probably* fixes that also need to be made to the BITBLT family functions to ensure that they produce the correct result.  I haven't checked that.

MattHeffron:
> (BITMAPCREATE 6000 34 15)

Just a typo in the comment? 15 isn't a valid BITSPERPIXEL value. I get an "Invalid argument: 15" error.

MattHeffron:
> If the fix here can result in component bitmaps that are less than 16 rows

The original `BITMAPCREATE.BIGBM` used the call to the _undocumented_ macro `IL:FLOOR` (from `sources/MODARITH`) in `(FLOOR x BITSPERWORD)` . It didn't check if the result was zero! There was just a comment that the slices **_should_** be multiples of 16 rows, which **_could_** be interpreted as trying to explain what was happening, and assuming that one knew that `BITSPERWORD` was 16.

It seems ill-advised to add a check and error for slices not 16n pixel rows tall, as that would be a substantial change in behavior. Likewise, adding a new parameter to `BITMAPCREATE` to indicate that the constraint should be enforced seems wrong.
It _is_ verifiable by checking that the `BITMAPHEIGHT` of the first entry of the `(BIGBM BIGBMLIST)` is 16n, if it's important in a particular case.

(That `FLOOR` macro will give strange results if the second arg isn't an integer power of 2. It doesn't do any checking!)
```common-lisp
(PUTPROPS FLOOR MACRO ((X N)
   (LOGAND X (CONSTANT (LOGXOR (SUB1 N) -1)))))
```

nbriggs:
textures are typically 16x16 bitmaps, and if the new code can generate bitmap strips that are not multiples of 16 rows high (before the last one) then the code also needs to align the starting row of the texture appropriately for the smaller strip or the resulting texture will not flow correct down the strips.

rmkaplan:
FLOOR, CEIL, FOLDHI, FOLDLO, UNFOLD and others on MODARITH were set up as internal macros to make calculations with power-of-two constants more transparent. It looks like some of them check their argument at compile-time, FLOOR and CEIL don't but probably should.

MattHeffron:
`BITMAPCREATE.BIGBM` now gives warning (using `CL:WARN`) if the slice rows are not a multiple of 16 pixels tall.
Added `BIGBM.ShortSliceHeightP` function to check a `BIGBM` for failing the 16 pixel tall slices condition.
(Checking for failing was easier to name the function than checking for meeting the condition.)

To suppress the warning message, use the construct:
```common-lisp
(HANDLER-BIND ((WARNING #'CONDITION:MUFFLE-WARNING))
  (BITMAPCREATE width height bitsperpixel))
```

pamoroso:
At commit a8eccb2 I get the warning also for pixel rows that *are* multiples of 16:

bitmap-warning

MattHeffron:
> I get the warning also for pixel rows that _are_ multiples of 16

The pixel rows in slices is unrelated to the `HEIGHT` of the original `BITMAP`.
When `BITMAPCREATE` is called to create a bitmap that would be too large for a single bitmap (> 131066 words), it calls `BITMAPCREATE.BIGBM` which creates a list of smaller `BITMAP` instances, each of which corresponds to a horizontal slice of the whole. The height of the slices are determined by how tall each **full width** `BITMAP` can be without exceeding the overall size limit. If that height is > 16, then it is truncated back to a multiple of 16. The design/implementation doesn't allow for vertical slices, or a 2D array of pieces of the whole.

So, in your cases, the `WIDTH` of 6000 and `BITSPERPIXEL` of 24 forced the slice height to 14.

pamoroso (approved):
It works as expected.

Issue #2646, opened 2026-06-09
MEDLEYDIR and {MEDLEY} are not updated when a sysout comes up on a new system


rmkaplan:
This problem was discussed at the technical meeting on 6/8/2026.

It first showed up when fonts could not be created when online was starting up.  It seems that MEDLEYDIR still had the value that was appropriate when the system was building (or whenever the runner setting was correct), but that it did not get changed to the proper value when the online system starts up for users.  The symptom is that fonts not already included in the loadup cannot be found, because the pseudohost path {MEDLEY}/fonts/ has not been updated.

This may also be the problem that Ryan reported in #2644, which also seems to have been in the online system, and it may be limited to the way that system starts up.  All the released system seem to work although they are clearly not running in the environment where they were constructed.

In the case of online, it appears that the system greet fails because it is requesting a font CLASSIC 14 BRR during the startup, and that request is happening before MEDLEY-INIT-VARS has run to update the pseudohost.

@fghalasz is going to switch to Helvetica 14 BRR, which does exist in the sysout--with luck, that will hide this problem.

But it is still not clear why MEDLEY-INIT-VARS hasn't run earlier in the startup.  It appears to be the first function on AROUNDEXITFNS to be applied by \USEREVENT when a system starts up (\USEREVENT runs the functions in reverse order).  Does the system greeting happen before the \USEREVENT?  Should MEDLEY-INIT-VARS be installed somewhere else, so that it comes before the system greeting?

rmkaplan:
Poking a little  more, in greetfiles/ there is a file MEDLEYDIR-INIT, and that has a call to MEDLEY-INIT-VARS.  If it is expected that that file comes early in the Greet sequence and that its call to MEDLEY-INIT-VARS is how the variable MEDLEYDIR and {MEDLEYDIR} are to be updated, then that may be one possible source of the problem.

It calls (MEDLEY-INIT-VARS 'GREET), where GREET is the event type.  But the call to set up the pseudohost is executed for the AFTERxxx events, not for GREET.  Would adding it also under GREET make a difference?

pamoroso:
Related to #2645.

masinter:
Two random LOADUP thoughts:

[...]

PR #2648, opened 2026-06-09
Added PSEUDOHOST MEDLEY under the GREET branch of MEDLEY-INIT-VARS

rmkaplan:
See the discussion in #2645.  Maybe this will ease the pain.

PR #2649, opened 2026-06-09
Fixes up the calling of (MEDLEY-INIT-VARS) when starting a sysout in an environment in which it was not created

fghalasz:
This PR supercedes and extends the PR#2648 from @rmkaplan .

This PR fixes up the calling of (MEDLEY-INIT-VARS) when starting a sysout (created by MAKESYS) in an environment in which it was not created.  It does 3 things:

1.  added AFTERMAKESYS event to the MEDLEY-INIT-VARS function.  Does the same as the GREET event.  

2.  in MEDLEY-INIT-VARS function added to the GREET/AFTERMAKESYS event processing the setting of the {MEDLEY} pseudoname host.  This is the same as PR#2648.

3. Third, changed the MEDLEYDIR-INIT greetfile so that it skips the LOADing of MEDLEYDIR and the calling of (MEDLEY-INIT-VARS) iff MEDLEYDIR is already loaded in the sysout.  That way (MEDLEY-INIT-VARS) is not called twice, once from the INIT file and once from the AROUNDEXITFNS.  And this way the MEDLEYDIR environment variables are properly set for all of the loadup sysouts when they are called with nogreet.


nbriggs:
Looks OK.  Is there any particular reason that the older code (still) does
```
(PSEUDOHOST 'MEDLEY (PACKFILENAME 'BODY (UNIX-GETENV "MEDLEYDIR")
                           'HOST
                           'DSK))
```
vs the newer code doing it with a call to `(MEDLEYDIR)`.  And why does it `(SETQ MEDLEYDIR)` then immediately `(SETQ MEDLEYDIR (MEDLEYDIR))` ?

nbriggs:
Can you push this to the online.interlisp.org development instance?

masinter:
> And why does it (SETQ MEDLEYDIR) then immediately (SETQ MEDLEYDIR (MEDLEYDIR)) ?

the variable MEDLEYDIR acts as a cache for the value of the function MEDLEYDIR.
So the first SETQ clears the cache and the second one installs a new value.

nbriggs:
@masinter ah, thanks.  Hadn't noticed that.

rmkaplan:
When I did this, I separated the pseudohost expression so that it didn't get mixed up with the various conditions in MEDLEYDIR, whether the variable is already bound or set or has a slash.  I think the host should be set up no matter what.  Maybe we want a separate subfunction that always sets the MEDLEYDIR variable and does the {MEDLEY} pseudohost, and nothing else.  (There is still some script where I had to reset the variable to NOBIND to get it to take effect.)

Note that MEDLEY-INIT-VARS also has RESET entries for ShellBrowser and ShellOpener.  Should those be moved to \SYSTEMCACHEVARS instead?

rmkaplan:
My bet is that that's to get around the NIL test in MEDLEYDIR, like setting it back to NOBIND in the script.  Those preconditions were suspicious, but I didn't want to mess with them.

fghalasz:
@nbriggs  Just finished building a draft Medley release that includes this PR, building the corresponding medley docker image for online and pushing it to the online development server.

fghalasz:
@rmkaplan  Seems to me that MEDLEYDIR and (TRUEFILENAME '{MEDLEY}) should always return the same value.  That's why I set {MEDLEY} using (a just reset) MEDLEYDIR.    If (MEDLEYDIR) sometimes returns something that is not compatible with pseudo hosts, we should update (MEDLEYDIR) so it always returns a pseudohost compatible value.

nbriggs (dismissed):
LGTM.  I checked it out on online.interlisp.org:8081 and it works great (except for DInfo, but that's not related to MEDLEYDIR and PSEUDOHOST)

rmkaplan:
What's wrong with dinfo?  Seems to work for me.

I think the dinfo problem before was also a MEDLEYDIR problem in that it couldn't find the fonts.

rmkaplan:
However, I think there is a separate unrelated problem: when I click Notecards Documentation there is pop up that says

sh: 1 /usr/local/interlisp/online/bin/request←new←table:  not found

Somewhere there is a Medley string with MCCS leftarrows (shift-hyphen) that need to be replaced by MCCS underscore (function 10).  Same for the other documentation buttons.

nbriggs:
If you open DInfo from the background menu, and then two-finger tap on the introduction window, it gets many of these:
image

rmkaplan:
OK, that doesn't appear to happen in PR #2634, which has all the Tedit selection updates.

pamoroso:
Supersedes #2648 and resolves #2646.

pamoroso:

rmkaplan:
The change to MEDLEYDIR-INITCOMS looks a little odd.  It seems to be testing for the **absence** of the namefield of a filename, not for a file with a **matching name**.

MattHeffron:
> It seems to be testing for the **absence** of the namefield of a filename, not for a file with a **matching name**.

Correct, but that prevents the LOAD from loading an already loaded file.
**HOWEVER**, the `(MEDLEY-INIT-VARS 'GREET)` probably shouldn't be inside the `(AND ...)`. 
As it is now, it executes **only** if the LOAD is executed, which is **not** the same behavior as the original that used `LOAD?`

And, shouldn't there be a `PACKFILENAME` that ensures there's a `HOST` of `{DSK}`?

rmkaplan:
Every file on LOADEDFILELST has a name field, in particular the first file.  So I think the NEVER condition is always false and nothing happens.  Maybe it wants to test (EQ 'MEDLEYDIR (FILENAMEFIELD F 'NAME)) ?

But, if some version of MEDLEYDIR has been loaded, then it would have a FILEDATES property, so is the desired set up something like
    (CL:UNLESS (GETP 'MEDLEYDIR 'FILEDATES)
               (LOAD (CONCAT  (OR (UNIX-GETENV ..) "") "/sources/MEDLEYDIR.LCOM"))
               (MEDLEY-INITVARS 'GREET))

fghalasz:
@rmkaplan. You are right about the LOADEDFILELST search.  My bad. Somewhere along a sequence of copies I lost an (EQ ... 'MEDLEYDIR) expression.  And my testing didn't catch that.  Just uploaded a fix into this PR.

pamoroso:
Nothing unusual to report at commit 122f8d4.

rmkaplan:
I wonder, is there a difference between running thru the LOADEDFILELST and testing the FILEDATES?  Not sure how all this is wired together.

masinter:
how would I go about creating a Medley distribution which _didn't_ iinclude the sources or the lcoms and dfasls in the distribution itself.  You'd probably want at least one fallback font? Or no?     
I was thinking files.interlisp.org could be used to serve sources if you want to look.
But it might help restore the distinction between "Medley as a development environment" and "Medley as a delivery platform".

masinter:
An important use case for the 'just the sysout' delivery is WASM... don't want to have to slog around megabytes just to run a short demo.

fghalasz:
@rmkaplan  I chose to go thru the FILESLOADLST and just check for the NAME field because I wanted to ensure that the INIT file would never overload MEDLEYDIR if there was already any version of MEDLEYDIR already in the sysout.   The principle being that a system-level INIT file should never overload files "behind the user's back" if the file has already been loaded.  If the user has a need for a later version of MEDLEYDIR they should load it explicitly (or in their personal INIT file.)

(I will admit that this was probably not practical when loadups took all day.  But now that loadups are a relatively lightweight thing this seems like the right way to go.)

LOAD? uses FULLNAME of files on LOADEDFILELST and FILEDATES properties to determine whether a file has been LOADed - which will kick off all kinds of false negatives if the question is ""has any version/date of file MEEDLEYDIR already been loaded".

rmkaplan (approved):
LGTM

RyanBurnside:
Did the issue by @nbriggs get a ticket/resolved?
I pulled the latest from Master and was still getting the 5 "Bad selection operation"  popup boxes.
I blew away my old vmem file first to ensure it wasn't loading that.

The website also currently has it if you open Dinfo and click anyware on the first page.

rmkaplan:
What are the 4 pop ups?  Can is there a screenshot?

RyanBurnside:
EDIT, must have gotten an update very recently - it's just fine now online.

I found simply single clicking does similar.

> If you open DInfo from the background menu, and then two-finger tap on the introduction window, it gets many of these: image

pamoroso:
I can't reproduce the issue on the latest Medley.

RyanBurnside:
Fix confirmed after a quick build. Working for me too.

PR #2657, opened 2026-06-15
Fix up the URL for the Notecards documentation file in APPS-INIT. 

fghalasz:
Somehow one character in the URL for the Notecards documentation file in APPS-INIT got changed from an underscore to a dash.  Fixed by changed back to a (true) underscore.


pamoroso:
The link to the NoteCards documentation is no longer broken but still points to version 1.2 of the manual. What about linking to version 2.0 https://files.interlisp.org/medley/notecards/docs/user-guide-v2.0 instead?

pamoroso (approved):
Looks good to me.

rmkaplan (approved):
LGTM

PR #2662, opened 2026-06-17
eliminate NLSETQ for handling null selection in SEDIT

masinter:


rmkaplan (approved):
I tested the various instances with or without an empty selection.  All seem to work.

rmkaplan (approved):
I tested the various instances with and without an empty selection

fghalasz (approved):
As per meeting on 6/17

PR #2671, opened 2026-06-20
Fix bug in FONTSAVAILABLE

rmkaplan:
Referenced an incorrect variable


pamoroso:
I've tested on Linux Mint 22.1 Cinnamon and everything seems to work fine except for this unexpected result:

```
8_ (FONTSAVAILABLE 'MODERN * 'MRR 0 'DISPLAY)
MODERN is not a NUMBER.
```

Am I supposed to pass `*` as the size?

rmkaplan:
That's supposed to work, I'll take a look

rmkaplan:
On further examination, I think there is an error in your input:  you should quote the *.

When the * is quoted, it works for me.

But the error message doesn't seem right, I would have expected to see a message that * is an unbound atom.  If I replace * with (unquoted) zzz, I get the expected error.  Is there something special about * ?

In any event, I don't think this is a problem inside FONTSAVAILABLE.

pamoroso:
You're right, quoting `*` works as expected.

masinter:
> Is there something special about `*` ?

yes, `+`, `-`, `*`, `,/,` all have special behavior when used as an unquoted IL symbol or as the first or last part of a unquoted IL symbol -- namely CLISP.
`(FONTSAVAILABLE 'MODERN * 'MRR 0 'DISPLAY)`

turns into (FONTSAVAILABLE (TIMES 'MODERN 'MRR) 0 'DISPLAY)

masinter (approved):
LGTM

PR #2676, opened 2026-06-23
Fixed loadup -a (aka loadup-apps-from-full.sh) to use the value of env variable NOTECARDSDIR as the Notecards directory

fghalasz:
Fixed loadup -a (aka loadup-apps-from-full.sh) to use the value of env variable NOTECARDSDIR as the Notecards directory.  Only if NOTECARDSDIR doesn't exist does it look for the Notecards directory in MEDLEYDIR/notecards and MEDLEYDIR/../notecards.  Similar fixes for ROOMSDIR and CLOSDIR (except default is simply MEDLEYDIR/rooms and MEDLEDIR/clos, respectively).

### DIFF
--- a/scripts/loadups/loadup-apps-from-full.sh
+++ b/scripts/loadups/loadup-apps-from-full.sh
@@ -6,31 +6,48 @@ main() {
 
 	loadup_start
 
-	export ROOMSDIR="${MEDLEYDIR}/rooms"
-	export CLOSDIR="${MEDLEYDIR}/clos"
-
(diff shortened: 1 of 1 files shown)


rmkaplan (approved):
I set up the variable to be alternatively working-notecards and git-notecards, was able to successfully build the app sysout both ways.

pamoroso:
Doing the apps loadup on Linux Mint 22.1 Cinnamon works fine with and without `NOTECARDSDIR` set.

PR #2677, opened 2026-06-23
BuildLoadup.yml:  Prevent the creation of a spurious (empty) release by fixing a one letter typo

fghalasz:
In buildLoadup.yml in the downloads-page job there was an incorrect setting for github_release_tag (caused by a one letter typo), which was causing a spurious empty release to be created on the releases page of the medley repo.  This fixes the github_release_tag setting.  One character change.


masinter (approved):
LGTM

PR #2678, opened 2026-06-29
Fix Tedit tabs, marginbar, object selection, bravo page numbers

rmkaplan:
This addresses the following issues:
*  A longstanding bug in the implementation of default tabs
* A bug in the interpretation of image object buttoneventinfns that was introduced in the previous revision dealing with mouse scrolling in and out of a window
* Display glitches in the layout of the margin bar in the paragraph looks menu

This also includes code for doing page numbers by means of a page number image object in an ordinary heading paragraph.  As a first step, I am trying this more flexible strategy for page numbers created by the bravo converter, so as not to muck up any existing files--that comes later.

There is one menu glitch that I have not been able to figure out.  If the caret is blinking in a menu input field (say the name of a font family) and you click on one of the other buttons at just the wrong time, you will see the caret stuck on the screen.  The button needs to wait until the caret is down before it grabs control, or it needs to force it down--any suggestions on how to do that?


pamoroso:
I haven't noticed anything unusual in my initial testing on Linux Mint 22.1 Cinnamon.

nbriggs:
I have an idea that the caret blinks via a TTYBACKGROUNDFNS thing- I’d start looking there for any ideas about how to synchronize with it coming down

rmkaplan:
TTYBACKGROUNDFNS does have the \CARET.FLASH?, but Tedit seems to run its own (in \TEDIT.COMMAND.LOOP).  I think that's because it is causing the caret to flash where the typein is visible in all the panes of a split window (even though menus are presumably never split)

The toggle buttons are interpreted in the mouse process (\TEDIT.BUTTONEVENTFN) while the caret is being flashed in the Tedit process.  I think I tried having the button give the TTY back to the main window before it did anything else, but that didn't help.

rmkaplan:
I pushed new versions for the files in this PR, most with minor changes for compatibility and for fixes to little glitches.

The main work is in the Bravo converter.  There is now machinery for representing named tabs as image objects, and  page numbers now move from one side to the other, according to the Bravo convention that the position is different for left and right pages. (The Binding parameter is another Bravo page layout convention--that one I haven't yet figure out.)

But the fact that there is now an imageobject platform for implementing named tabs doesn't mean that tabbed material now shows up in the right place on the display or hardcopy.  The only specification I have is the Mayberry documentation in docs/bravo, and its description doesn't cover everything I see in the one heavily tabbed file I have been working on, from the IDL manual.  Mayberry cites a another document by Greg Kusnick that presumably has more technical detail, but I haven't been able to find it.  

So, does anyone have a better understanding of how named-tab definitions should propagate from run to run, what should happen with unnamed tabs in runs with name definitions , or what should happen when tab 0 is specified?  That would help a lot.

One other confusion in all of this is that the metrics for the Helvetica fonts that we now have seem to be different from the ones in the Alto/Press days--our characters are a little less wide.  That can also throw off some tab alignments.

pamoroso:
Still looking good at commit 3b54dc0.

pamoroso:
Still looking good at commit bc9e337.

rmkaplan:
Bravo's vertical tabs define the top of the line of text, Tedit's SPECIALY specifies the font-dependent baseline.  This now does the proper transformation.

This work revealed an off-by-one bug when a page has multiple headings, now fixed.

Also, some dead code and unused arguments in the hardcopy routines have been removed.

pamoroso:
At commit 3b3c39a, after loading the LispUsers module EQUATIONS and opening the example file EQUATIONEXAMPLES.TEDIT in TEdit I get black boxes instead of the summation and integral symbols:

equation-error

TEdit has always displayed this document correctly and I don't remember when I last opened it. The issue, which occurs also in Medley Online 18-Jun-2026 02:17:24, may have been introduced recently but I've noticed it only when testing the latest commit in this branch.

rmkaplan:
This is a separate issue, a consequence of the transformation to fonts that are consistently recoded to MCCS.

The EQUATION package uses SIGMA 20 for summation and integral, and specifies that the summation glyph in that font is the character M.  But sigma has been converted to MCCS, and the summation glyph is now at 357,172 in this and every other transformed font.

So, this is one of the downside edge cases of our font normalization, an existing data structure that has legacy font-specific codes.  In this case, I think a brute force fix is for EQUATION to do the transformation of M in SIGMA to 357,172 as part of the getfn.

rmkaplan:
The problem with EQUATIONS is addressed in PR #2686

pamoroso:
Nothing unusual at commit 05649f9.

pamoroso (approved):
Looks good to me.

PR #2686, opened 2026-07-08
EQUATIONS summation and integral use MCCS codes (and NS fonts instead of SIGMA)

rmkaplan:
This fixed the problem that @pamoroso reported in PR #2678 .

Turns out that EQUATIONFORMS already had a switch that controlled whether to use MCCS (nee NS) fonts and codes or to use SIGMA instead.

I changed the default so that it always uses MCCS codes.  We get Modern instead of Sigma, so the look is a little different.  That switch could be interpreted instead as just switching the fonts instead of the encoding.

(I also renamed the involved variables and functions to replace NS in their names with MCCS, for less confusion going forward).


pamoroso (dismissed):
I tested this on Linux Mint 22.1 Cinnamon and it fixes the [issue](https://github.com/Interlisp/medley/pull/2678#issuecomment-4912704720) reported in PR #2678.

rmkaplan:
I just added a revised documentation file, probaby needs to be approved again

pamoroso (approved):
Looks good to me.

pamoroso:
Just for comparison this is the sample document EQUATIONEXAMPLES.TEDIT with the Sigma font in December of 2024:

equations-dec2024

The same document with the Modern font in July of 2026:

equations-jul2026

rmkaplan:
Are you pointing to the difference in the glyphs for the summation and integral?

I think that's the difference between those characters in SIGMA 20 and CLASSIC 24 (for Sum) and MODERN 30 (for Integral).

Previously, the default setting of the switch was to say don't use NS fonts, in which case it defaulted to SIGMA 20 for the display.  I changed that default setting, and in that case it takes the fonts from the list EQ.MCCSChars, which is where it gets the CLASSIC 24 and MODERN 30 for Sum and Integral, etc.  Those could be changed to SIGMA 20 in the INITVARS. 

But I think that list also specifies what happens on hardcopy, so we would want to make sure that SIGMA 20 for postscript/pdf looks good too.

pamoroso:
Yes, I'm pointing to the difference in the glyphs. Since I had the old screenshot I thought it would have been interesting to compare it with what the symbols look like now.

rmkaplan:
I looked at the original font list again.  It seems a little goofy, it's a property list but it has 2 entries for SUM.  The first one (the one that it sees) is for Sigma in CLASSIC 24, the second is for Summation in Modern 30.  PRODUCT also as 2 entries.

I think Tadd was playing around to get the best display appearance, from Greek vs symbol code against Sigma 20, Classic 24, Modern 30.

I suppose we should also play around to see which looks best.

rmkaplan:
Something else that is a little odd:  the right arrow in the limit example shows up as a left arrow in the recent display.  That's a slittle surprising, because I don't think we fooled around with the right arrow, just the left arrow. Maybe he built in a transformation for that too.

PR #2687, opened 2026-07-10
Rmk206  offline enlargement of small ns display fonts

rmkaplan:
This includes code for producing enlarged versions of NS display fonts as part of the font import process--the step of reading strike and ac files and producing the first set of equivalent medleyfont files.  Those files then feed the rest of the offline process of MCCS code conversion, completion for missing glyphs and character sets, and faking bold/italic.

The effect is that Classic 10 display fonts, for example, have the larger glyphs taken from the Classic 12 AC fontfile source.  These are the glyphs that would have been produced by loading the lispusers/NSDISPLAYSIZES package, but that would not have fixed all of the glyphs that had been inherited or faked from other sources as part of the offline font construction process.  This conceptualizes the enlargement as just part of the importation process from funky external source files.  This should obsolete the NSDISPLAYSIZES package.

The upside of this is that glyphs in small (less than nominal size 12) NS fonts will be more readable on the display, and they will be closer in size to the glyphs of the same nominal size for non NS fonts.  Previously (without loading NSDISPLAYSIZES), there would be a visible size difference in a line that had Classic 10 characters next to Timesroman 10 characters.  Now they are much closer.

The downside of this (as with the previous NSDISPLAYSIZES strategy) is that some characters or images that were carefully formatted with small NS fonts on the display may now not have the desired appearance--lines may overflow, characters may extend beyond intended boxes.  But the hardcopy appearance is not affected.

This PR also includes new versions of all the medley display fonts (306 font files).  I thought I could just include new versions of the small NS fonts, but I realized that Timesroman and other fonts were extended/completed by including glyphs from Classic/Modern of the same nominal size.  So in the end I just re-imported everything.


pamoroso:
In my initial tests on Linux Mint 22.1 Cinnamon at 1080p I notice text is slightly larger in Execs, menus, and elsewhere. For example, text in DInfo is more readable.

I assume the old font behavior is no longer available not even as an option, right?

pamoroso:
There are now text overflow issues with VIRTUALKEYBOARDS, see for example the function key row of the DEFAULT keyboard:

virtualkeyboard

This is what it looked like in April of 2023:

virtualkeyboard-apr2023

Also, the right and down arrow symbols in the numeric keypad are replaced by black boxes and the left and up arrows are rendered as `_` and `^`. Medley Online shows the right and down arrows correctly but not the left and up ones, so this may be a font encoding rather than enlargment issue.

pamoroso:
In CALENDAR the top of the year is slightly cropped in the year view.

calendar

pamoroso:
In MISSILE the text `MISSILES>` partially overlaps the graphics next to it. Here's how it was in April of 2023:

missilecommand-apr2023

Now:

missilecommand

pamoroso:
There are text overflow issues in the FM-CREATOR LispUsers module. Now:

fmcreator

In March of 2023:

fmcreator-mar2023

rmkaplan:
The display problem with virtual keyboards was noted in the original NSDISPLAYSIZES, I have to upload the fix.

But I also have to make sure that the keyboard codes are now MCCS compliant, a separate issue.

rmkaplan:
The old behavior can be made available by also keeping a version of the old fonts with their old nominal sizes, perhaps in an overlay fonts directory.  Just like the current NSDISPLAYSIZES package, the incore cache of fonts would be cleared out so that switching back  would catch the creation of any new font descriptors.  But descriptors already created and stashed in other data structures or variables would not be updated.

Or the loadup could be built with the old fonts, and NSDISPLAYSIZES could be fixed just to  smash the cache and then point to the directory with the new descriptors (again, as now, missing fonts that are stored in other places).  

The main improvement here is that enlarged fonts that nherit from other enlarged fonts and so have more consistent glyph sizes.

pamoroso:
I was thinking not so much of dynamically switching font behavior but rather a way of starting a Medley session with the old behavior, or doing a loadup with the old behavior if it's not practical to do it per session. Such a "classic mode" would provide maximum compatibility and an experience closer to the original Medley.

rmkaplan:
Switching at loadup is easy:  just have the loadup's DISPLAYFONTDIRECTORIES point to either of two of offline-constructed fonts.

On further thought, though, it would be pretty easy to switch all the fonts in memory, no matter where they may be stashed--so long as they were all created by FONTCREATE and thus stored in the \FONTSINCORE cache.  We could whip through the cache, smashing the existing font descriptors with field values taken from the alternative set of fonts.  Probably just a few lines of code.  The fonts in memory would all be in a consistent state.

pamoroso:
Do you mean something like `export DISPLAYFONTDIRECTORIES=/home/paolo/medley/medley/fontsold; ./scripts/loadup-all.sh -apps` for the loadup?

nbriggs:
I have never used NSDISPLAYSIZES, and it appears that doing so breaks other expectations -- because the glyph sizes now don't match the widths?  I would prefer to NOT have the NSDISPLAYSIZES effects by default.  If someone feels the benefits outweigh the problems then they should be able to optionally load NSDISPLAYSIZES.

rmkaplan:
I would do it on just a switch in the scripts:  NSDISPLAYSIZES = YES, NO.  If NIL, then whatever default we pick

rmkaplan:
Looking at the virtualkeyboards revealed another separate issue, having to do with different codes with the same glyphs (I forget what the term is for that).  Character set 342 of our Unicode mapping tables defines the International Phonetic Alphabet.  Some of the codes there have the same glyphs as codes in other character sets.

For example, the XCCS IPA code for c-cedilla xE2DA maps to Unicode x00E7, IPA code for k xE2DE maps to x006B.  

The problem appears when we read a UTF8 file that has a Unicode c-cedilla code 00E7: what MCCS/XCCS code should it show up as?  The phonetic c-cedilla 342,332  or the extended Latin c-cedilla 361,255.

We are currently getting the phonetic code in the UTOMCODE direction, and that is showing up as a black box in the virtual keyboard displays when that now-UTF8 file is read, because we haven't moved the c-cedilla glyph over to the phonetic code.  That's how the problem emerged:  the file has the unicode c-cedilla, we read it into the unintended MCCS code, and it's black.

Not sure what to do.  XCCS is making more distinctions here than Unicode.  Simplest thing would be to exclude the IPA character set when we build the mapping tables.

Other ideas?

rmkaplan:
A further thought on the IPA homoglyph problem...

I think we should just assert that the confusing homoglyphs in XCCS IPA don't exist in MCCS, that MCCS is like Unicode in this particular regard.  That is, we remove those confusing correspondences from the UTOMCCS mapping tables:  Unicode c-cedilla will map always to MCCS (XCCS) c-cedilla and not to IPA.  In the extremely unlikely event that we encounter an XCCS file with an IPA c-cedilla code, we would read that to the normal MCCS normal c-cedilla.  If we were then to write that file out, to MCCS or even again to XCCS, we would have lost the more specific XCCS encoding.  Oh well.

Michele31415:
On 7/10/2026 8:45 AM, Paolo Amoroso wrote:
> *pamoroso* left a comment (Interlisp/medley#2687) 
> <https://github.com/Interlisp/medley/pull/2687#issuecomment-4935397231>
>
> In CALENDAR the top of the year is slightly cropped in the year view.
>
That's odd.  It looks normal to me, in a sysout with a MAKESYSDATE of 
8-Apr-2026.

             - Michele

pamoroso:
The 8-Apr-2026 sysout didn't do the text enlargement of this branch my comment referred to. Se the text of the year, which is little bit larger in this branch, is slightly cropped at the top.

rmkaplan:
> On 7/10/2026 8:45 AM, Paolo Amoroso wrote: *pamoroso* left a comment ([Interlisp/medley#2687](https://github.com/Interlisp/medley/pull/2687)) <[#2687 (comment)](https://github.com/Interlisp/medley/pull/2687#issuecomment-4935397231)> In CALENDAR the top of the year is slightly cropped in the year view.
> That's odd.  It looks normal to me, in a sysout with a MAKESYSDATE of 8-Apr-2026.             - Michele

@pamoroso is reporting the glitches that happen if the small NS fonts are bumped in size (NSDISPLAYSIZES), for better consistenty of nominal sizes across fonts.

pamoroso:
Still looking good at commit f2894c0.

pamoroso:
Still looking good at commit ca36cf1.

PR #2688, opened 2026-07-11
Rmk207  MCCS codes for virtualkeyboards plus eliminate IPA homoglyphs

rmkaplan:
This adjusts the display and keyboard mappings to take account of the arrow/caret interchanges, and to eliminate the confusion in the Unicode mappings because XCCS associates the same glyphs with IPA codes and other codes (as discussed in PR #2687).

This is independent of the NSDISPLAYSIZE issue and fonts, in that it chooses the default keyboard font based on its height.  However, for reasons that I don't yet understand, the labels on the keyboard (NEXT, EXPAND) seem to be in a bigger font, even though as best as I can tell, those are specified as Helvetic 5, which is not involved in any of this font stuff.

As a separate point:  The French keyboard seems a little goofy, in that shift d does produces a circumflex-o and not capital D.  And shift e is also odd.

There is still a black box in the logic keyboard.  Is there an image anywhere of what character is supposed to be there?

This also modernizes the keyboard display windows, so they can be dragged around by their title bars.


rmkaplan:
Actually, the problem with the labels appears only with the enlarged NS fonts, even though I don't think NS is involved there.  This is compatible with current master in that regard.

One other virtualkeyboard issue:  the keyboard displays have the left panel taken over from the dandelion or daybreak, with a bunch of labeled buttons that are not connected to any actions.  Some of them could make sense and could be armed for Tedit or Sedit (e.g. Undo, Find, Cut, Open), but I think it would be better just to eliminate those buttons.  A lot of the labels on the function key and number pad also aren't meaningful.

(BTW:  What does "RDMND" stand for?)

pamoroso:
On Linux Mint 22.1 Cinnamon the arrows are now displayed correctly and the labels fit within the function keys. However, `F10` is missing the label `LOOKS`.

The logic keyboard is displayed in [this screenshot](https://github.com/Interlisp/medley/pull/2687#issuecomment-4935188100). Now the left arrow just to the left of `RSHIFT` is displayed as `_`.

nbriggs:
The Daybreak/Dandelion left keys should all be tied to the keyboard bits that were originally attached to the physical keys on those machines.  Likewise the top function keys and the right keypad.  The Lyric/Medley (?) release notes should have the key number mappings.  If you wanted to change anything, making the display of the additional keyboard sections optional on a FLG would make sense.  Removing them would not be helpful, especially if you have code that was written to use those keys.

nbriggs:
"RDMND" is Right DiaMoND.

pamoroso:
I wasn't clear, I meant that left arrow just to the left of `RSHIFT` *in the DEFAULT keyboard* is displayed as `_`.

pamoroso:
Apologies again for the confusion, it's the **LOGIC** keyboard I meant, not the default one.

rmkaplan:
> "RDMND" is Right DiaMoND.

OK, then, what is/was Right diamond?

> Apologies again for the confusion, it's the **LOGIC** keyboard I meant, not the default one.

That's another MCCS update, I'll do that.  

On your earlier image, the key to the left of that arrow key has a small empty-set symbol, I now see a black box there.  What do you see now?

On the F10-LOOKS, we have now defined it so that F10 is the way of typing underscore.  We can change that back to LOOKS, for old times sake.  But then you wouldn't be able to type an underscore when this is switched in.

pamoroso:
At commit 91a12c7, I see the DEFAULT and LOGIC keyboards like this:

vkeyboards

The black box is still there and the left arrow is fixed. As for `F10` and `LOOKS` it makes sense, thanks.

rmkaplan:
By the way, there is the same issue with the caret that is replacing Underline on F6.  (By the same argument, Currency should replace UCASE on F4, but nobody would miss Currency)

The Dvorak keyboard is broken in a different way, causes an error if you click to display.  Something about inconsistent assignment of key numbers. I think we can let that go, maybe just take Dvorak off the menu.

rmkaplan:
At least on the Mac, with XKEYBOARDS, the only labeled keys that transmit anything are the formatting function keys.  Those seem to work, at least in Tedit, transmitting the same codes as Tedit's separate button panel.

The left-panel keys don't do anything, nor do any of the unshifted labeled keys in the number panel.

Maybe more would happen if we loaded the dandelion keyboards instead of the x keyboards.   Or maybe for those that make sense, at least for Tedit, we could add the appropriate keybindings. But mostly these buttons now are false advertisements.

nbriggs:
I'm just at a place where I can see the images -- Those are Sun keyboards (type-4, I think) -- check this against the Lyric/Medley release notes.  In theory, I suppose the virtual keyboards should adapt to the actual physical keyboard you have, but we haven't got that information coming through in a way that it can easily be used (there's a 3(?) bit field that indicates the keyboard type).  Sun keyboards had a diamond key to the right of the space bar.  Please don't remove bits of  the keyboard layouts until we can work out a proper plan for how it should work.

nbriggs:
This is the Sun type-4 keyboard that VirtualKeyboards assumes you're working with.   image  <br>  Maiko/Medley never accounted for international versions of the Sun keyboard.

rmkaplan:
On the black box in the logic keyboard...

That key/button sends the empty-set character when unshifted, which makes sense.

When shifted, it sends the upper-case Norwegian slashed O, which doesn't make sense.  In an earlier version of the fonts, that showed up as something that looked like the empty set, perhaps intended to be a glyph of a different size.  But switching around the fonts probably landed on one that didn't have a glyph for that, hence the black box.

But this set up was probably designed more by a typographer than a logician:  there is no mathematic notion of a larger or smaller empty set.  And the proper way to get a size difference, if that was the intent, is for the client to switch fonts--the keyboard can't do that.

So that assignment in the logic keyboard should be replaced by some other useful symbol.  Any suggestions?

Michele31415:
On 7/11/2026 3:25 PM, Nick Briggs wrote:
> *nbriggs* left a comment (Interlisp/medley#2688) 
> <https://github.com/Interlisp/medley/pull/2688#issuecomment-4948459393>
>
> I'm just at a place where I can see the images -- Those are Sun 
> keyboards (type-4, I think)
>
The Sun Type 4 keyboard has little open diamonds to the left and right 
of the space bar.  The Type 5 bottom row goes: Control, little black 
diamond, Alt, space, Alt Graph, little black diamond, Compose.  I don't 
think I've ever used any of those except for control and space.

             - Michele

pamoroso:
At commit b18f515 I get many more black boxes on the LOGIC keyboard.

virtualkeyb

rmkaplan:
So far, it appears that this is not a problem with the keyboard file, but a much older and more basic bug in the utf-8 reader that seems to affect 3-byte codes.  Maybe an incorrect shift.

rmkaplan:
Actually, the utf-8 code is good.  I think it may be another homoglyph problem in the tables.  

(UTOMCODE (MTOUCODE  X)) does not always return X.

pamoroso:
A couple of commits earlier in this PR there were no black boxes in the LOGIC keyboard, so it may be some very recent change.

pamoroso:
Nothing new to report at commit f94cc32.

rmkaplan:
The Logic keyboard black boxes arise because I made the XKEYBOARDS file be a UTF8 file, and at the same time I replaced the key-names  (357,102) in the keyboard data with their actual characters so the file is interpretable outside of Medley.  It was also a good test case for the mappings--most of our files are really only Ascii, except for the arrows.

Some of the black boxes in the Logic keyboard came from randomly picking the unintended  instance of a class of homoglyph-characters for the Unicode in the file, when mapping backwards from the Unicodes to MCCS.  We saw this first in the IPA confusions, but this is a more general problem.

With the new code, the inverse mappings for homoglyph unicodes are resolved according to an ordered list of charset-preferences (HOMOGLYPH-PREFERENCES)--choose the code in the set whose charset is earliest in the list.  So, LATIN, then SYMBOLS1...  GEORGIAN is not in the list.  If none of the codes in a homoglyph equivalence class are in the list, arbitrarily choose the smallest code. 

A second problem was that there were 2 entries missing in the 357 mapping file, so 2 MCCS codes in the logic keyboard did not have Unicode mappings, and were put out as slugs in the XKEYBOARDS file and read back into black boxes.

This commit also adds to CHARACTERSETNAMES the names of all the (non-JIS) character sets in the mapping files.  So (CHARCODE DEVANAGRI,101) will give you a code.

(Separately, it isn't clear to me why there are separate files for the code assignments for DORADOKEYBOARD, DANDELIONKEYBOARDS.  The assignment of codes to keys ought to be common for all the keyboards (key labeled P gets the same codes in all keyboards), the difference is the physical layout of those labeled keys.)

(There are still 2 glitches in the XKEYBOARDS:

pamoroso:
Nothing unusual to report at commit 5873b61.

What additional glitches did you mean? Your [post](https://github.com/Interlisp/medley/pull/2688#issuecomment-4983583493) seems truncated.

rmkaplan:
(Ahh, I got interrupted by the meeting)

The 2 glitches are the French D and E, and a better assignment to the lower right key on the Logic keyboard

rmkaplan:
I did some more work on the virtual keyboards, I made a new XKEYBOARDS, and I also made new UTF-8 files for Dandelion, Dorado, and Dove keyboards.

On XKEYBOARDS, I fixed the French glitches by comparing with the version in the Venue sysout.  The accented characters are now on the shifts of the numbers, and the D and E are now there.  I saw a few glitches in other keyboards (currency?), fixed.

Note that the X keyboard layout in the Medley sysout is slightly different than what we currently have (e.g. no right diamond), something must have switched along the way.

I haven't been able to get the settings right to look at the Dorado/Dandelion/Dove, but basically I fiddled the arrows in the usual way.

pamoroso:
Still looking fine at commit 583b5b8.

pamoroso (approved):
Looks good to me.

PR #2690, opened 2026-07-11
Change limit arrow to MCCS right arrow in EQUATIONEXAMPLES.TEDIT

rmkaplan:


pamoroso:
On Linux Mint 22.1 Cinnamon I still get the left arrow in the limit.

limit

rmkaplan:
What do you see when you execute

(GDATE (CAR (NTH (TEDIT.FORMATTEDFILEP (QUOTE lispusers/EQUATIONEXAMPLES.TEDIT)) 5)))

The time component should be 13:49:01

pamoroso:
I see `"11-Jul-2026 22:49:01"`.

rmkaplan:
I think that's the same--GDATE must take into account the time zone.

So, if that's the version you are looking at, it's the same as the one I have, and I see the right arrow.  So, a mystery.

pamoroso:
I do the apps loadup with Medley/rmk208--Fix-limit-arrow-character-in-EQUATIONEXAMPLES at commit 67569c0 and Maiko/nhb-pcap-ethernet built for X11 at commit 4f7be29. I run Medley under VNC.

rmkaplan:
That all makes sense.  Another check:  the length of the file should be 13285.

Try 

(BREAK \TEDIT.OPENTEXTFILE) 

before you call TEDIT, make sure that the value after EVAL is the intended file.  GETEOFPTR of the resulting stream should return 13285.

pamoroso:
The length of the file:

```
paolo@lispmachine:~/medley/medley/lispusers$ ls -l EQUATIONEXAMPLES.TEDIT 
-rw-rw-r-- 1 paolo paolo 13285 Jul 12 09:52 EQUATIONEXAMPLES.TEDIT
```

The break window:

equations

I'm not sure what EVAL you mean and how to access the stream so that I can call `GETEOFPTR`.

rmkaplan:
When you go into the break, type

EVAL

You should then see a stream, which is the value of the function.  You should be able to refer to that stream (the value of the last input) by the variable IT (all in IL:):

So after EVAL,

(GETEOFPTR IT)

Also, the filename you see in the stream should be the name of the intended file/version.

pamoroso:
Never mind, I found the cause of the issue: in my tests I opened a local copy of EQUATIONEXAMPLES.TEDIT in `~/il/`, not the one in the Medley source tree. The local copy is old and with a different length:

```
paolo@lispmachine:~/il$ ls -l EQUATIONEXAMPLES.TEDIT 
-rw-rw-r-- 1 paolo paolo 12971 Jun 22  2025 EQUATIONEXAMPLES.TEDIT
```

Here is the local copy: [EQUATIONEXAMPLES.TEDIT.zip](https://github.com/user-attachments/files/29944132/EQUATIONEXAMPLES.TEDIT.zip)

When I open the correct file I do see the right arrow in the limit:

equations

I think you no longer need the data from the break window, right?

pamoroso (approved):
It works as intended with no apparent issues.

PR #2691, opened 2026-07-13
Move Cygwin from SDL to X

fghalasz:
This PR depends on PR#549 in the Maiko repo.

SDL no longer works reliably with Cygwin.  This PR includes the changes to the medley script and the Cygwin installer that are required to move Cygwin from SDL to X. 

This PR also includes fixes to several bugs in the prior Cygwin install including
1.  CLIPBOARD did not work because xclip was not installed in Cygwin.  It now is included in medley.iss installer.
2.  The lock that prevented two medley installations with the same id from being run simultaneously did not work for Cygwin due to differences in the ps flags between Linux/MacOS and Cygwin.  This PR fixes by special casing the ps flags for Cygwin.
3.  loadup-apps-from-full.sh did not work on Cygwin because the mechanisms for locating the Notecards directory did not account for the fact that MEDLEYDIR is at the top level (i.e., /medley) in Cygwin.  This PR fixes this.


pamoroso:
It builds and runs with no issues on Linux Mint 22.1 Cinnamon.

pamoroso:
Still looking good at commit 9fb8309.

nbriggs (approved):
Seems OK.  Interlisp/maiko#549 has been merged, so this can be merged.

PR #2692, opened 2026-07-13
merge FIXMETA into LLKEY

masinter:


pamoroso:

masinter:
from @nbriggs 
> This doesn't seem right.  The left Alt key on a mac, when you're running X, is NOT the meta key -- the Command key is the meta key.  If you're running SDL, the default is for the left Alt key to be the meta key. The default mapping happens in the Maiko code.  Some people prefer meta on Command, others prefer meta on Alt.

> Can we have a design discussion, on e-mail, about what it *should* be doing rather than just putting hacks into the lisp code.

nbriggs:
With X as the display subsystem, on a Mac, the left alt key (aka option) is the Lisp key `DEF'N` or `EXPAND`.  The left command key is the Lisp key `BLANK-TOP`, `KEYBOARD`, `DBK-META`.

WIth SDL as the display subsystem, the left command key is not mapped, but the left alt key is the Lisp key  `BLANK-TOP`, `KEYBOARD`, `DBK-META`.

This should be made consistent -- if there's not reason to do otherwise, I'd make it the X11 mapping -- but some portion of the users will want to add (or maybe already have) KEYACTION calls to put the `(METADOWN . METAUP)` action onto the key *they* want rather than the default.

If the issue that `FIXMETA` was trying to address was that the meta key was assigned to some physical key other than one of these two, then we should look into that, and it may be that for Maiko systems it's the right thing to do to put an entry on the Maiko-specific keyactions table -- but lets be clear about what's going on.

Note also that the `METASHIFT` function should be looked at to ensure that what it is doing is not inconsistent (or broken by any changes that are proposed), and that there are two definitions of METASHIFT, one on LLKEY and one on VIRTUALKEYBOARDS.

rmkaplan:
If different environments naturally want meta to be assigned to keys with different keynames, does it matter if there are 2 assignments (i.e. to both keynames) if nothing else is assigned to either of those?

nbriggs:
It's not that SDL/X11 naturally want meta assigned to any key -- the person who coded the X11 mappings thought that (command) was the "correct" key for meta, and the person who coded the SDL mappings thought that (option) was the "correct" key -- pure personal preference, and they each assigned different physical keys to the *same* Lisp key name, but for SDL the other key (command) is unusable because it has no bit assigned in the key map.   If you're using SDL you *cannot* make the (command) key behave as meta.  This is just wrong.  I haven't tracked down what Lisp key number is assigned the meta function by default (without either FIXMETA or same change incorporated into LLKEY).

masinter:
s per discussion 7/26/26 meeting, FIXMETA call in LOADUP-FULL is just unnecessary and confusing. Closing this PR to add

PR #2693, opened 2026-07-13
Remove a variable conflict in the expansion of PUTMULTI in MULTI-ALIST

rmkaplan:
It was rebinding $$VAL, which made it not work right inside an iterative statement.


pamoroso:
It works as expected on Linux Mint 22.1 Cinnamon.

pamoroso:
Still looking good at commit b8765a2.

pamoroso (approved):
It works as described in my testing.

PR #2697, opened 2026-07-15
Fix Issue#2696 Weekly build fails when scheduled but works fine when dispatched manually. 

fghalasz:
Fixed by moving scheduled trigger to a new workflow that simply calls the original weekly build workflow with the default inputs specified.  This is to make up for act that Github Actions does not allow default inputs to be specified in the on:schedule section.

This PR needs to be merged before it can be tested since the new workflow will not show up on the Actions to be kicked off manually nor will it be scheduled until it is merged into the default branch.


stumbo:
Frank - would a better approach be to set default values like this?
```
  # Build Loadup
  do_release:
    uses: ./.github/workflows/buildLoadup.yml
    with:
        maiko_release: ${{ inputs.maiko_release || 'Latest' }}
        draft: ${{ inputs.draft || false }}
        force: ${{ inputs.force || false }}
    secrets: inherit
```

I think that would give you your default values for the CRON job and keep everything in one file.

pamoroso (approved):
Looks good to me.

PR #2698, opened 2026-07-15
clean up BICLOCK idle -- suspend process list on idle (IDLE.SUSPEND.PROCESS.LIST) is part of IDLE, biclock adds to that var

masinter:
processes to spsend when running idle, doesn't depend on idle.profile at all, just a globalvar called IDLE.SUSPEND.PROCESS.NAMES.


pamoroso:

nbriggs (dismissed):
I'm not sure that I understand how the title (...is a globalvar) relates to the changes in here.

I see that `BICLOCKPROCESS#2` is added as another likely name to suspend on idle - does BICLOCK usually/always generate 2 processes?

And I see that the code in the IDLER is now mapping over the process table once looking for those with a name on the IDLE.SUSPEND.PROCESS.NAMES list instead of going through the IDLE.SUSPEND.PROCESS.NAMES and trying to FIND.PROCESS for each one... but GLOBALVAR vs. IDLE.PROFILE ?  Is this just the placement of the code in the function?

masinter:
This PR doesn't solve the problem that I was trying to solve:  when the idle program is BICLOCK, and you're running (BICLOCK), when you start to IDLE, it doesn't suspend the original BICLOCK and there are two clocks, one bouncing around.

I don't know why sometimes starting (BICLOCK) starts the clock running in a process BICLOCKPROCESS#2.

The online help for Idle in the background menu has the IDLE.PROFILE option of SUSPEND.PROCESS.NAMES, but in fact the list of processes to suspend should not depend on the idle program. The logic of pulling out the SUSPEND.PROCESS.NAMES from the IDLE.PROFILE was broken anyway. This PR instead adds a IDLE.SUSPEND.PROCCESS.NAMES which are just names, so changing the logic to search (using MEMB) for any process active that is suspendable should be more effective.

But there's still something going on i don't understand -- now when I idle it suspends BICLOCK when I start to idle, but the original and now suspended process just disappears.

It may be the thing to do is go back to the VENUESYSOUT version?

nbriggs:
When the IDLE function is BICLOCK, it starts a new BICLOCK process and moves that window around the screen -- so if you already had a BICLOCK process running, the one generated by idle will be `BICLOCKPROCESS#2`.  If the user has BICLOCK loaded, but no BICLOCK running, then the IDLE generated process will be just plain `BICLOCKPROCESS` -- so that makes it hard to decide which ones need to be suspended during idle, and which don't.  I suspect you'd get mostly the right behavior if IDLE suspended all the processes that it was asked to before doing anything that would spawn an idle related process.  That way, it would be suspending nothing, or BICLOCKPROCESS, and then subsequently starting the idle BICLOCK as `...#2`.  Of course it wouldn't work right if someone had two BICLOCKs running on their display.

pamoroso:
Still no issues at commit abb571f.

pamoroso (approved):
Looks good to me.

PR #2700, opened 2026-07-17
Fix bug in TMAX update, may help HCFILES failure

rmkaplan:
Hardcopy doesn't fail when run locally, and there was a different error description from HCFILES.  But this fixes one definite problem--failing when clicking the Update menu item


pamoroso:
It works as described on Linux Mint 22.1 Cinnamon.

pamoroso (approved):
Looks good to me.

PR #2703, opened 2026-07-26
Fix the creation of Interpress fonts

rmkaplan:
This initializes INTERPRESSFONTDIRECTORIES and fixes a wrong-argument bug in the Interpress charset function.

I have been able to use the hardcopy command in the window menu to create an Interpress file, but I don't have dodo set up to try the option of sending to a printer.  I don't know if the file is correct.

One glitch that I still see on this window-menu path, compared to the Tedit meta-P, is that it doesn't print the number of pages and the name of the output file in the promptwindow when its done.


pamoroso:
On Linux Mint 22.1 Cinnamon with Dodo I can successfully print to the Interpress printer `POSEIDON:`, i.e. generate this PDF file which has a minor formatting issue: [job_019F_9FD7_9F5C.pdf](https://github.com/user-attachments/files/30392406/job_019F_9FD7_9F5C.pdf). However, even with `Meta-P` in TEdit I don't get the number of pages and the name of the output file in the prompt window:

interpress-print

pamoroso:
Never mind, I didn't notice the number of pages and the name of the output file *are* printed in the prompt window but of TEdit, not the system prompt window.

rmkaplan:
Did you get the pdf by converting from the ip file?  How does that happen?

rmkaplan:
Yes, but not when you use the window hardcopy menu item?

nbriggs:
From Dodo's readme.md:

```
    - Printing    
        (Courier program 4, version 3)    
        all protocol procedures are implemented, allowing to print from XDE, Star/Viewpoint/GVWin and
        Interlisp-D as well as to query a print job status, the printer properties and the printer status.    
        The interpress files are received and collected in the output directory,
        a PostScript file can optionally be generated for the interpress master.
        and possibly post-processed.    
        The PostScript generation provides a basic support for western-european
        character sets as well as simple graphics (vector and uncompressed bitmaps),
        giving usable but far from perfect results.
```

pamoroso:
After issuing the print command Dodo takes care of everything automatically and deposits the final PDF file in a predefined directory. Behind the scenes Dodo creates a number of intermediate files, here are the ones of the above test job:

```
job_019F_9FD7_9F5C.interpress  job_019F_9FD7_9F5C.ps
job_019F_9FD7_9F5C.ip          job_019F_9FD7_9F5C.ps.log
job_019F_9FD7_9F5C.pdf
```

This is the contents of the log file job_019F_9FD7_9F5C.ps.log:

```
Postprocessing generated PostScript:
... psFilename: ./prt-poseidon/job_019F_9FD7_9F5C.ps
... paperSize.: a4
... jobName...: 019F_9FD7_9F5C
... objectName: STREAM
... senderName: paolo
... workdir...: /home/paolo/medley/dodo-xns-services/prt-poseidon
... fnbase....: job_019F_9FD7_9F5C
```

I confirm I don't get the number of pages and file name when printing from the window hardcopy menu.

I'm aware Dodo has limitations with Interpress printing but it's interesting to take a look at the output of a document with many symbols, lispusers/EQUATIONEXAMPLES.TEDIT: [equationexamples.pdf](https://github.com/user-attachments/files/30392686/equationexamples.pdf).

rmkaplan:
The message-printing was in the meta-P calling function, I'll move it to TEDIT.TO.IMAGEFILE.

The equations file looks pretty bad, I wonder if the problem is in the IP file itself, or in the conversion from IP to PDF.  Do you have a side-by-side example of the pdf's from a file that is mostly text (with maybe some headings and pagenumbers)?

pamoroso:
The documentation of the COMPARESOURCES LispUsers module is mostly text with headings and is a few pages long. You can compare the [HCFILES output](https://files.interlisp.org/medley/lispusers/COMPARESOURCES.TEDIT.pdf) with the converted Interpress file generated by the Dodo print service: [comparesources.pdf](https://github.com/user-attachments/files/30404118/comparesources.pdf).

pamoroso:
The [documentation of the Dodo print service](https://github.com/devhawala/dodo/blob/master/printsvc-configuration.md) describes from section [Font and character set handling for PostScript generation](https://github.com/devhawala/dodo/blob/master/printsvc-configuration.md#font-and-character-set-handling-for-postscript-generation) onward how it processes Interpress and the limitations of this approach.

nbriggs:
Except for the issue with the dot in envos, my impression is that the character metrics in the Dodo version are better than the PS/PDF from HCFILES.  It would be good to compare versions of a document we haven't modified where we have the image of the printed paper as generated from a real Interpress printer and the Dodo and HCFILES version.

rmkaplan:
My first thought is that it isn't in dodo, that it is an XCCS/MCCS mismatch in Interpress for the peculiar (non-XCCS fonts).  It may be a double coercion that has to be rooted out, once the font metrics are properly imported.

So a quick experiment would be 
   (ADVISE '\TEDIT.CONVERT.XCCSTOMCCS  'BEFORE '(RETURN))
before opening the file, and then printing.  What does that get the Envos, and the equation examples?

pamoroso (approved):
It works as described.

pamoroso:
Here are the same documents after the advice `(ADVISE '\TEDIT.CONVERT.XCCSTOMCCS  'BEFORE '(RETURN))`:

* [equationexamples-advised.pdf](https://github.com/user-attachments/files/30430292/equationexamples-advised.pdf)
* [comparesources-advised.pdf](https://github.com/user-attachments/files/30430295/comparesources-advised.pdf)

rmkaplan:
Thanks, doesn't make a difference.  So more work to do.

rmkaplan:
In the badly formatted Envos line, the center-dot character is 0,267  (x387) in Classic 24 bold.

Can you create a simple test file containing that single byte, and try printing it in Classic 24 bold?  And also Classic 10 MRR?

pamoroso:
Do you mean a TEdit file? How do I type that specific character?

rmkaplan:
No, the idea is to keep tedit out of it and to see whether it is just a font issue.  

So:
(SETQ STRM (OPENSTREAM 'xxx.txt 'OUTPUT))
(BOUT STRM 267Q)
(CLOSEF STRM)

pamoroso:
Okay thanks. Should I print to the default printer or to Interpress?

rmkaplan:
Print both, and they should look the same. You should see a single center dot.

nbriggs:
A way to create the Interpress file that you can send to the Interpress printer is
```
(SETQ STRM (OPENIMAGESTREAM "{dsk}<tmp>xxx.ip" 'INTERPRESS))
(DSPFONT (FONTCREATE 'CLASSIC 12 'BRR 0 'INTERPRESS) STRM)
(PRINTOUT STRM (CHARACTER 267Q))
(CLOSEF STRM)
(SEND.FILE.TO.PRINTER "{dsk}<tmp>xxx.ip" "poseidon:")
```
(assuming you have Dodo printer poseidon set up, and networking is running etc. etc.)

rmkaplan:
Yes, that's better.  Although it should also be able to do the font setting by the simpler

(DSPFONT '(CLASSIC 24 BOLD) STRM)

Do the same for .pdf and PDF, and for (CLASSIC 10 MRR).  Send the pdf to your Unix printer, or Viewer.

pamoroso:
The file with only the centerdot character: [CENTERDOT.BIN.zip](https://github.com/user-attachments/files/30558215/CENTERDOT.BIN.zip).

When I `SEND.FILE.TO.PRINTER` to `poseidon:` nothing seems to happen and Dodo doesn't generate any PDF files in its print service directory. Also, what should I pass to `OPENIMAGESTREAM` as the second argument for the default printer?

pamoroso:
Never mind, the printer type is `PDF` (and in this case `SEND.FILE.TO.PRINTER` isn't needed as `CLOSEF` generates the PDF). The other files:

* [centerdot-classic10mrr.PDF](https://github.com/user-attachments/files/30559016/centerdot-classic10mrr.PDF)
* [centerdot-classic12brr.PDF](https://github.com/user-attachments/files/30559019/centerdot-classic12brr.PDF)
* [centerdot-classic24brr.PDF](https://github.com/user-attachments/files/30559021/centerdot-classic24brr.PDF)

rmkaplan:
Are these all from the Interpress stream then converted to pdf?  Or direct to pdf?

pamoroso:
The [PDFs I shared](https://github.com/Interlisp/medley/pull/2703#issuecomment-5135173295) are made from imagestreams of type `PDF`. I couldn't make PDFs from .ip files as Dodo does nothing when I pass the files to `SEND.FILE.TO.PRINTER` and doesn't generate any PDFs.

rmkaplan:
How did you get Dodo to give you the EQUATIONEXAMPLE pdf before, with the bad characters?  Tedit just opens an IP imagestream, prints on it, and sends it to the printer.  So it should work to take Tedit out of the loop (and hopefully looks bad).

pamoroso:
I got the EQUATIONEXAMPLES.TEDIT Interpress PDF by printing from TEdit with `Alt-P` or the `Hardcopy` menu item.

rmkaplan:
If you break SEND.FILE.TO.PRINTER when you run under Tedit, you should see how it is sending the IP file to the printer.  The call should be the same for the test case, except for the name of the file it is sending.  Does that not work?

pamoroso:
This is what I get when breaking `SEND.FILE.TO.PRINTER`, there's no Unix shell command among the arguments.

send-file-to-printer

rmkaplan:
And then, what happens if you set IMAGESOURCE to the name of your little test file, and say OK?

nbriggs:
@pamoroso it sends the file to Poseidon using the XNS Printing protocol - it is correct that there should be no Unix commands involved.  You have to have the colon in the printer name or it will assume it's a Unix printer.

nbriggs:
@pamoroso - I guess I need to get the GVWin installation and extract enough to get the XNS Printing IP/PS/PDF set up - where did you find it?

pamoroso:
I think I did provide the colon when passing the Interpress printer name to Lisp but I'll try again.

I downloaded [GlobalView 2.1 for Windows at the Internet Archive](https://archive.org/details/gvwin).

pamoroso:
I tried again from Lisp by including the colon in `poseidon:` but nothing changes, still no PDF output from an Interpress file.

Setting `IMAGESOURCE` to the name of my first Interpress file from the broken `SEND.FILE.TO.PRINTER` finally completes successfully and produces the first PDF: [centerdot-classic24brr.pdf](https://github.com/user-attachments/files/30597545/centerdot-classic24brr.pdf).

However, with the second file centerdot-classic10mrr I can't go past the break window. When I execute `OK` a TTY window pops up and the break window becomes unresponsive.

print-break

nbriggs:
You should probably make sure your Interpress output stream is named with a ".IP" extension in case the file-type detection is relying on that.

pamoroso:
You're right, the file name was missing the extension. Passing the right name does finally produce a PDF for the second Interpress file: [centerdot-classic10mrr.pdf](https://github.com/user-attachments/files/30597786/centerdot-classic10mrr.pdf).

rmkaplan:
It shouldn't need an extension if you provide the type, I'll look into that.  (And it shouldn't need the type if you provide the extension.)

But:  We finally have a result:  The Dodo pdf is bad.  It looks like 267Q, the center dot, is showing up as 264Q, the multiplication symbol.

I believe Nick looked at the file (as he created it) and saw that it did contain a byte 267Q.

So I believe the problem is in Dodo's pdf conversion.

rmkaplan:
Also, now it would be worth testing the sum and integral symbols in equationexamples.

Try the same test with (CHARCODE 46,126) and (CHARCODE  357,165).

--Ron

pamoroso:
Both PDF and Interpress? What font?

pamoroso:
Also, what values should I pass to `CHARACTER`?

nbriggs:
Exactly `(CHARACTER (CHARCODE 46,126))` and so on (that's a sigma).

pamoroso:
What font?

nbriggs:
Perhaps try Classic 24 MRR and BRR, and Modern 24 MRR.  They need to be big enough to see.

pamoroso:
Okay these are the Interpress files:

* [sigmaint-classic24brr.pdf](https://github.com/user-attachments/files/30600732/sigmaint-classic24brr.pdf)
* [sigmaint-classic24mrr.pdf](https://github.com/user-attachments/files/30600741/sigmaint-classic24mrr.pdf)
* [sigmaint-modern24mrr.pdf](https://github.com/user-attachments/files/30600745/sigmaint-modern24mrr.pdf)

I did call `(CHARACTER (CHARCODE 46,126))` and `(CHARACTER (CHARCODE 357,165)))` from `PRINTOUT` but it looks like Dodo has a thing for the center-dot character.

nbriggs:
For the original dot vs multiplication sign issue -- if you don't use the pfb files supplied with GlobalView then the Postscript and PDF generated by Dodo have the correct dot character, but of course the fonts and spacing are all wrong.  You can try it by just doing `gs job_mumble.ps` for one of the ps files in the prt-poseidon directory (assuming you have ghostscript's gs installed).

pamoroso:
Yes, I'm using the PFB files of GlobalView.

nbriggs:
If you try to print a Sigma, it generates this Interpress:
```
Header "Interpress/Xerox/2.1 "
BEGIN
  {
    Identifier "XEROX"
    Identifier "XC1-1-1"
    Identifier "TERMINAL"
    3 MAKEVEC
    FINDFONT
    282.22222900 SCALE
    MODIFYFONT
    1 FSET
  }
  {
    0.00001000 SCALE
    CONCATT
    2794 26449 SETXY
    25 0 SETCORRECTTOLERANCE
    1 SETFONT
    1 18 ISET
    2540 26449 SETXY
    Identifier "XEROX"
    Identifier "XC1-1-1"
    Identifier "CLASSIC"
    3 MAKEVEC
    FINDFONT
    846.66668701 SCALE
    MODIFYFONT
    2 FSET
    2 SETFONT
    -1701 SETYREL
    565 0 SETCORRECTMEASURE
    CORRECT
    {
      String "\xFF&V"
      SHOW
    }
    1 SETFONT
  }
END
```
I suspect that `"\xFF&V"` is not correct, but...

rmkaplan:
Is "\xFF&V" supposed to be a hex num, or somehow the bytes as characters of an XCCS code?

nbriggs:
Hmm... good point, that looks like the bytes of the NS character that I was inserting, (CHARCODE 46,126) -- `(\xFF) 255 (&)46Q (V) 126Q`

rmkaplan:
So I would say here too that we are creating the intended IP file, and the downstream mapping to PS/PDF is incorrect.

The spacing will be wrong because we are formatting with the metrics for the IP fonts on Interpress printers, but we don't know have the metrics for the PS/PDF fonts that Dodo maps them too.  If we knew the target fonts and their metrics, we could finagle our starting metric files and get Dodo pdf's to look right.  But is there any reason to pursue this?

PR #2706, opened 2026-07-29
Rework of Tedit hardcopy, other relatively minor adjustments

rmkaplan:
This is a substantial cleanup of the Tedit hardcopy layout routines, something that I had avoided messing with before.  The code is simpler, I think, and should do a better job at widow/orphan control, special locations of paragraphs, headings, and revision marks.  Should also be somewhat faster (could be speeded up quite a bit, but that's a later pass).

Please test against whatever oddball documents you know about, multiple columns as well as headings etc.

This includes the basic code for footnotes, but I think the original design was incomplete, and there was no user interface.   So I doubt that there are any documents that depend on this.

(Note: the stack overflow that I talked about in the 7/29/2026 meeting has been fixed)


rmkaplan:
This includes some other improvements, among them:

TEDIT.INCLUDE was buggy, now should work.

Setting paragraph looks after having just typed a CR to create a new paragraph should now apply the new looks to the new paragraph, that the caret blinks in front of, not the previous paragraph.

pamoroso:
I tested this on Linux Mint 22.1 Cinnamon and hardcopy works as expected, including Interpress.

The upper limit 1 of the intergal is missing from the PDF when printing lispusers/EQUATIONEXAMPLES.TEDIT: [EQUATIONEXAMPLES.pdf](https://github.com/user-attachments/files/30537390/EQUATIONEXAMPLES.pdf).

rmkaplan:
I think the upper limit is there, but it isn't as high and it maybe isn't spaced as far to the right as it should be, so it overlaps.  That may be a consequence of changing the font from sigma, maybe the code has a built-in assumption about the spacing.  Something else to look at.

But with respect to the fontclass in this pr, I think I'll do an update that has his font function return the class itself without coercing to its display fontspec.  That should also work, and it would preserve more information.

rmkaplan:
I tried to print a ps file for equationexamples in the venue sysout, to see what the original spacing might have been.

But it failed because it couldn't find a postscript font-mapping for sigma 20.

pamoroso:
Still looking good at commit a612ad9.

pamoroso (approved):
Looks good to me.

Issue #2040, opened 2025-02-20
Discussion: Permanent extension of the "Medley Character Set Standard"


rmkaplan:
Our current internal character encoding standard (MCCS) is basically XCCS with a few fiddles (dollar sign, left-arrow, which we haven't really decided on but would be handled by an explicit  XCCS to MCCS mapping).  Imporantly, MCCS codes correspond to the layout of glyphs in our fonts.  This is an initial proposal for how to extend MCCS to produce MCCS-Unicode mappings for a large number of additional Unicode characters so that we can use Unicode fonts to eliminate all of our black boxes.

First some arithmetic:

MCCS/Unicode mappings are based on the mappings for 10730 XCCS codes in 105 character sets.  Let's assume that the XCCS standard does not define any other XCCS codes that would be relevant to MCCS, i.e. we would never encounter XCCS-coded files that have standard codes outside of the mappings that we currently have.

That leaves 54805 smallp's that can be assigned to other Unicode characters.

But XCCS/MCCS does not allow 255 in any character set:
   54805 - 256 = 54549  (actually, we don't need to preserve this constraint in internal MCCS--a separate issue)

Codes in the MCCS  META and FUNCTION character sets are reserved.  
   54549 - 2*255 = 54039  (511 since we have already taken out the 255 codes)
   (We should map them to MCCS codes that are undefined in both XCCS and Unicode so that we don't have to deal with them again if we ever go to Unicode internally.)

We want to reserve some number of MCCS codes for local faking of unmapped codes, say  4 character sets = 4*255
   54039 - 4*255 = 53019

Thus we have 53019 smallp MCCS codes that can be assigned to otherwise unassigned Unicode characters.

(Side note: the maximum size of an Interlisp hash table is 32749.  So we have to divide codes into at least 2 separate hash buckets.)

The maximum number of Unicode plane 0 codes is 65535 - 6400 (reserved) = 59135 (maybe some others are not defined in the Unicode standard, I didn't check).

So we can't have smallp MCCS codes for 6116 smallp Unicodes (less whatever we want to allow from higher ups--emojis....)

A simple strategy for permanent extension to MCCS:

	Make a list of character sets (or specific characters) in Unicode that we don't care about and another list of available MCCS codes (as calculated above). 

	Then for all defined Unicode characters U from 0 to 65535 (plus others beyond that that we care about):

		Unless (UTOXCODE? U),  assign the next available MCCS code to U.

[...]

PR #2709, opened 2026-07-30
reduce \MAXCHARSET in LLCHAR to 0x10FF, actual limit in FULL Unicode

MattHeffron:
I reduced the `\MAXCHARSET` constant in `LLCHAR` to 0x10FF which is the actual limit in FULL Unicode.
This should prevent the error creating a basic `FONTDESCRIPTOR` from a too large block allocation.

In the same sysout, I tried marking all (?) `FNS`, `MACROS`, and `OPTIMIZERS` as `CHANGED` that referenced this directly, _or indirectly_, in other files, so they could be recompiled with the modified constant.
But doing `CLEANUP` or `MAKEFILE` of `LLREAD` (the first attempt) failed with an apparent parsing error of the `:PACKAGE` as `%:PACKAGE` in the `DEFINE-FILE-INFO`. I didn't try further; but I saved the virtualmem file to try later with advice from others...

I have **no idea** where the changes to `INTERPRESS*` came from.
I didn't touch them, and they didn't show as modified before the git commit.
I also don't know why this commit appears to have been associated with Larry!

The `%MAXCHARSET` in the branch name was because `\` is not allowed in a branch name.


rmkaplan (commented):
LLCHAR looks good. The changes in Interpress seem to be from the previous merged PR.  Did something else change there?

rmkaplan:
BTW, I'm moving the slugcarsetinfo into the fontdescriptor, so only adding one to the font's maxcharset (or \MAXCHARSET) to get the vector length from the last index.  A few macro changes and a recompile--seems to work.

Not sure what is supposed to happen with the 0 case you mentioned.

MattHeffron:
> Not sure what is supposed to happen with the 0 case you mentioned.

With my change in `LLCHAR`, (_and_ adding 1 not 2), the "0 case" would _successfully_ allocate 4352 entries to the `FONTCHARSETVECTOR` instead of _failing_ to allocate 65536 entries.

For me, it all went _sideways_ when I tried to forcibly recompile the files that appear to use that constant in `FNS`, `MACROS`, or `OPTIMIZERS`. The files that I identified as such are: `ACFONT`, `FILEIO`, `FONT`, `LLREAD`, `MEDLEYFONTFORMAT`. There probably are others. I'm not sure of the best way to track them all down...

MattHeffron:
**Further consideration...**
I'm not sure that changing `\MAXCHARSET` from 255, is really the right solution to holding fonts that are UNICODE encoded. 
I saw a few places in the existing code that used a hardcoded `(< CSET 256)` style test, instead of using the named constant. 
Also, with `\MAXCHARSET` > 255, wouldn't that imply that `\MAXFATCHAR` should be > 65535? (It isn't.)
The change in `\MAXCHARSET` _did_ cause `\MAXCHAR` to be redefined as `0x10FFFF` which exceeds `\MAXFATCHAR` and doesn't fit in a 16 bit word. (I didn't look for references to `\MAXCHAR` for files to recompile.)

Since **all** of the rest of the char handling code assumes the old `\MAXCHARSET` and `\MAXTHINCHAR` values of 255, either the handling of UNICODE-encoded `MEDLEYdisplayFONT` should be handled differently, or not at all. I'm not sure what "differently" would entail, but shouldn't break anything else.
(I can remove, or conditionalize, the code in `READ-BDF` that tried to create a UNICODE-encoded `MEDLEYDISPLAYFONT` file. I think that there isn't any other code that tried to do this.)

MattHeffron:
> The changes in Interpress seem to be from the previous merged PR. Did something else change there?

I'm _guessing_ that the Interpress changes were merged into master **after** I did a `git pull` from master, but before I committed my LLCHAR changes, and so show as changes from the master that was the baseline for my new branch.

rmkaplan:
If you update from master, does the Interpress go away?

rmkaplan:
In terms of recompiling, I think masterscope would give a good list of places that \MAXCHARSET is referenced, either directly or through macros.  I am looking for and recompiling all the places that calls \BUILDSLUGCHARSETINFO so that they get updated with the macro that puts the slug csinfo in the new fontdescriptor field.  Functions that allocate an extra vector cell should still be safe if they miss a recompile, just a little larger.

On the number of fat chars, I think there was a brief discussion of this when the extension idea first came up, but I can't find it now.  If we were really to go for this, the font change now is the first enabling step.  But I think the maximum fat character should remain at 65535, and that's what fat strings and atoms can contain.  But there would be a new class of characters, call them veryfat, which would not be restricted to smallps, and they would also go into their own class of strings.  The elements would be stored as FIXP pointers in a pointer block.

The problem is to fix it so that  same veryfat codes obtained (NTHCHARCODE) from 2 different veryfat strings will be EQ and not just IEQP.  A first approximation, which would solve some but not all of the cases, is to have CONCAT, RPLCHARCODE, etc hash a FIXP into a canonical representative whose pointer would be stored in the block.  Code that is testing SMALLP instead of CHARCODEP would fail.  As would code that does arithmetic on codes but expects the results to be EQ.  But maybe this would be good enough for unusual applications, like emojis.

rmkaplan:
Here are some other issues that might be relevant (I searched for emoji).

#2040 #2540

pamoroso:
Nothing unusual to report on Linux Mint 22.1 Cinnamon and Interpress printing still works with Dodo.

nbriggs:
`\BUILDSLUGCSINFO` -- the symbol is referenced in:
```
sources/ACFONT.DFASL
sources/MEDLEYFONTFORMAT.LCOM
sources/FONT.LCOM
sources/FONT
sources/LLDISPLAY.LCOM
library/tedit/TEDIT-SCREEN.LCOM
library/IMPORTFONTS.LCOM
internal/loadups/starter.sysout
internal/venuesysouts/full.venuesysout
internal/venuesysouts/lisp.venuesysout
```
`\MAXCHARSET` --
```
sources/FILEIO
sources/LLCHAR
sources/ACFONT
sources/LLCHAR.LCOM
sources/FONT
sources/LLREAD
internal/FONT-DEBUG
```
I notice that we have:
```
       (\MAXCHAR (LOGOR (LLSH \MAXCHARSET 8)
                        \MAXTHINCHAR))
```
as a constant -- does that cause issues when the result is not SMALLP?

rmkaplan:
\BUILDSLUGCSINFO is mostly called from a macro SLUGCSINFO (which is where I'm changing to the fontdescriptor field).  Masterscope knows about that but grep doesn't.  Also WHOCALLS will find it in loaded compiled code (but it won't find the constants).

I think I have found all the places, at least I've been running with the slugcsinfo stored in the font descriptor.

I don't think there is a problem with \MAXCHAR, that should work with FIXP's but perhaps not bignums.  But references to \MAXCHAR also need to be recompiled.

rmkaplan:
If the only change is the \MAXCHARSET in LLCHAR, it might be best for me to just put that with the other files in the PR I'm creating, with the slugcsinfo change and the various recompiles, and then delete your draft.

MattHeffron:
Withdrawn. @rmkaplan will make a _different_ change (restoring `\MAXCHARSET` to the _original_ value of 255.)

PR #2710, opened 2026-07-30
Upgrade EQUATION editor to recognize fontclass charlooks

rmkaplan:
It was not expecting a fontclass in the return from TEDIT.GET.LOOKS


pamoroso:

rmkaplan:
I updated so that it just passes the font or fontclass along.  Also marked the edit window as not-dirty after the initial insertion, so it doesn't ask when you try to exit without making any changes.

pamoroso:
At commit b4964fd the upper limit of the ingeral is set apart more from the symbol: [equationexamples.pdf](https://github.com/user-attachments/files/30554522/equationexamples.pdf).

pamoroso:
I found a reproducible stack overflow that leads to URAID:

1. open lispusers/EQUATIONEXAMPLES.TEDIT
2. right-click on the integral

equations-overflow

```
paolo@lispmachine:~$ medley -a -e -n -v
Using DISPLAY=:1
Using VNC_PORT=5900
"/home/paolo/bin/lde" "/home/paolo/medley/medley/loadups/apps.sysout" -id "default" -title "Medley" -g 1440x900 -sc 1440x900 -noscroll
LOGINDIR: "/home/paolo/il"
GREET FILE: "/home/paolo/medley/medley/greetfiles/APPS-INIT.LCOM"
REM.CM FILE: ""
VMEM FILE: "/home/paolo/il/vmem/lisp.virtualmem"

*Error* URAID Called:
Enter the URaid
"Stack Full -- Type LU to see stack; ^D to flush to top"

< LU
Unsupported command.

< lu
  0 :    0x11594 : IL:RAID
  1 :    0x1157e : IL:\MP.ERROR
  2 :    0x11558 : IL:\FREESTACKBLOCK
  3 :    0x1152a : IL:\PROCESS.MOVEFRAME
  4 :    0x11512 : IL:\DOMISCAPPLY1
  5 :    0x11502 : IL:\DOMISCAPPLY
>>root frame for contextsw<<


< 
```

By the way, the URAID hint `Type LU to see stack` is misleading as uppercase is apparently not supported.

rmkaplan:
To me, on my screen, it still overlaps the integral in the same way.

rmkaplan:
I have tracked this down to the fact that the buttoneventinfn on the equation image objects (EQUIO.ButtonEventIn) calls SPAWN.MOUSE before it branches on the mouse button.  For the left and middle buttons, that brings up its menus to start an edit, and it tells Tedit that the object has changed.

For the right button, it just returns NIL to tell Tedit that Tedit should handle the interaction, as if it was a click on a character.  It may be that Tedit is looping in that case, calling the image object function repeatedly.

But it may be that this is also a bug in the image object function.  If I take out the spawn.mouse altogether, the behavior for the left/middle still seems good, and right doesn't overflow.

rmkaplan:
I removed the SPAWN.MOUSE, works fine without it.  I also fixed it so that choosing Exit in the Edit window doesn't ask you to confirm.

I still don't understand why there was somewhere a loop calling the buttoneventinfn, but this separately showed that Tedit wasn't treating a NIL return from the buttoneventinfn as an indication that object itself had just been selected.  And some of the other image object functions weren't being called in all the right places.  So there is more on the Tedit side.  But I think the equation code is OK.

pamoroso:
Commit 78e8cb4 fixes the stack overflow.

However, `Edit` doesn't open an editing window, it just selects the relevent piece of equation and inserts the caret in the main document. It's been a while since I last edited an equation, so I don't remember whether TEdit is supposed to open a separate editing window such as DEdit.

nbriggs:
The documentation for SPAWN.MOUSE and ALLOW.BUTTON.EVENTS says:

> It never hurts, at least logically, to call SPAWN.MOUSE or ALLOW.BUTTON.EVENTS needlessly, as the mouse process arranges to quietly kill itself if it returns from the user's BUTTONEVENTFN and finds that another mouse process has sprung up in the meantime.  (There is, of course, some computational expense.)

so if that's true, and the call to SPAWN.MOUSE causes a problem, then it seems as though there's a problem somewhere else...

rmkaplan:
For me it opens a little Tedit window at the bottom, that allows for editing a single line for a single equation element (you can't type EOL's).  And it highlights in the main window the element that you are editing.  So my behavior is different.

(I do see another problem, in that I changed the Exit function to avoid the askuser, and that depends on a change that I made to a Tedit file.  So I have to decide whether it makes sense to put that Tedit change in this PR, or to sequence them in a different way.  But getting the window to open up is a separate question.)

pamoroso:
Nothing new to report at commit 6c2657c.

rmkaplan:
Equation edits should now be undoable.  I removed the dependency on a pending Tedit update, so this should be testable on its own.

pamoroso:
At commit 73b74ba equation edits are undoable. And, by the way, the commit brings back the editing window at the bottom of the TEdit window.

pamoroso:
Still no issues at commit 745a955.

rmkaplan:
If this looks good now, it should be approved and merged.  I don't know of anything more to do.

pamoroso:
At commit 59396d5 the edit window no longer shows up when editing an equation.

pamoroso:
If the edit window issue is unrelated to this PR I can approve it.

rmkaplan:
I think the editor is being surprisingly clever, or maybe too clever.  But unrelated to this PR.

When you click with the left mouse over the object, you expect the Select/Edit menu to come up.  And when you click Edit, you expect to see the edit window with a candidate string taken from the object (e.g. a numerator or denominator string), and that's what you edit.

You'll also see that that part of the object is highlighted in the main window, to remind you what you are editing and where your edited string will end up. 

But what if you click on a region of the object that actually isn't part of one of the components, just some white space in the lay out?  In poking around, I think if it doesn't know what part you want to edit, it doesn't bother to bring up the window.

You should play around to confirm that this is the right theory.

If you click with the middle button, it gives you the list of components to choose from.  But with the left, if it can't figure out what you want, it just bails out.

That could be fixed up, but I think in a separate round.   Maybe put a "You didn't choose a component" message in the title bar.  Or maybe have the left button behave as the middle button, put up a menu, when it can't resolve the click to a particular component.

rmkaplan:
Here's a test case:  the summation object.  If you click on the Sigma, nothing happens.  If you click on the limits, you get the edit window.

pamoroso:
I confirm the behavior: clicking on the Sigma or integral nothing happens, clicking on the limits I get the edit window.

pamoroso (approved):
Looks good to me.

PR #2711, opened 2026-07-31
Upload newly built DANCER font files

MattHeffron:
Uploaded sizes 8-14  - and replaces sizes 10 and 12.
Other sizes "available"...


pamoroso:
I'm not sure how I'm supposed to use these fonts. If I apply Dancer 14 to this text in TEdit:

```
a b c d e f g h i j k l m n o p q r s t u v w x y z
```

I get black boxes (top window):

dancer

Opening DANCEROBJ-SAMPLES.TEDIT produces unknown IMAGEOBJ errors (bottom window) and loading DANCEROBJ yields an error (break window) because DANCEROBJ is in a different branch.

rmkaplan:
Maybe approve/merge the DANCEROBJ PR first (it's harmless), and then work from that.

MattHeffron:
Duplicate of my comment from PR #2717, as it's somewhat related to both...

None of this, below, is _directly_ related to FONTSAMPLER, but it shows what the glyph mapping _should_ be and the _intended_ appearance of the glyphs (which look pretty poor at 1 pixel per point resolution).

I hand modified a .PS file of an older FONTSAMPLER output, to see what the PostScript Type 1 font generated.
(Sorry, that I used a file that reads across vs. reads down. That's the way FontForge shows it, but I didn't choose for that reason.)
Here's [DANCER12.ps.txt](https://github.com/user-attachments/files/30807867/DANCER12.ps.txt) (added `.txt` because GitHub)
And here's the generated [DANCER12.pdf](https://github.com/user-attachments/files/30807891/DANCER12.pdf)
I _did_ have to modify the original Type 1 font file:
I commented out the `serverdict begin 0 exitserver` (line 3) because `ps2pdf` (a.k.a. `ghostscript`) failed with a security issue. (I'm guessing that it's only required if **installing** the font on a PostScript printer.)
I also removed the `EOT` control char at the end.
FontForge had given a warning about the size of an EN space (1600 vs. 1024).
I noticed in the font file:
`/FontMatrix [0.000625 0 0 0.000625 0 0] def`
And that the `0.000625` is 1/1600. I replaced that line with:
`/FontMatrix [0.0009765625 0 0 0.0009765625 0 0] def`
where `0.0009765625` is 1/1024.
That scaled the renderings up a bit, maybe closer to what the old DANCER DISPLAYFONT files had. I'll try generating new BDF files and then new MEDLEYDISPLAYFONT files. (I'll wait until FONTSAMPLER works, so we can get before and after for comparisons.)
Here's the updated font file: 
[Dancer.pfa.txt](https://github.com/user-attachments/files/30808402/Dancer.pfa.txt) (added `.txt` because GitHub)
I "installed" the font by creating a symbolic link in `/usr/share/ghostscript/fonts/Dancer` linked to `/home/matt/Dancer.pfa`.
I used ghostscript directly, insead of `ps2pdf` because I was debugging the issues with the font file.
```shell
gs -dNOPAUSE -dBATCH -sDEVICE=pdfwrite -sOutputFile=DANCER12.pdf DANCER12.ps
```

MattHeffron:
> I'll try generating new BDF files and then new MEDLEYDISPLAYFONT files. (I'll wait until FONTSAMPLER works, so we can get before and after for comparisons.)

I tried this, as mentioned above. The differences were negligible, and not all positive, at that.
These display fonts just don't have the pixel resolution for the details present in the PostScript font.
(I _could_ try having `otf2bdf` use a higher pixel resolution, but a given font size in such a font would be _noticeably larger_ than _other_ DISPLAY fonts.)
I _can_ generate larger point size MEDLEYDISPLAYFONT files from the earlier BDF files, if they'd be useful. (It's pretty quick for just the DANCER fonts.)
Since we _have_ a Dancer PostScript font, I should investigate how to provide that info to the POSTSCRIPTSTREAM driver, and through to the conversion to PDF. I'll add an enhancement issue.

PR #2713, opened 2026-08-01
Alist format for default PRINTFILETYPE conversions, infer imagetype from extension

rmkaplan:
Minor adjustments to printing interface

* Default conversions on PRINTFILETYPES is now represented as an alist instead of a plist.  Simpler code and easier to read.

* OPENIMAGESTREAM now can deduce the imagetype from a file's extension, if the imagetype is not explicitly provided and the file's extension is recognizable.  (OPENIMAGESTREAM 'foo.IP) will create an Interpress imagestream on foo.ip, equivalent to (OPENIMAGESTREAM 'foo.ip 'INTERPRESS).


pamoroso:

rmkaplan:
Replaced by #2716

PR #2716, opened 2026-08-03
ALIST format for default conversions, infer imagetype from extension

rmkaplan:
(Replace PR #2713, for compatible merging)

Minor adjustments to printing interface

Default conversions on PRINTFILETYPES is now represented as an alist instead of a plist. Simpler code and easier to read.

OPENIMAGESTREAM now can deduce the imagetype from a file's extension, if the imagetype is not explicitly provided and the file's extension is recognizable. (OPENIMAGESTREAM 'foo.IP) will create an Interpress imagestream on foo.ip, equivalent to (OPENIMAGESTREAM 'foo.ip 'INTERPRESS).

Issue #2714, opened 2026-08-03
lispusers>FONTSAMPLER broken due to changes in font/medleyfont/hardcopy code


MattHeffron:
The current compiled FONTSAMPLER.LCOM fails when outputting to a PDF file.
It hasn't been updated since some of the font/medleyfont/hardcopy changes.
For example:

1. The `TitleFont` for the generated pages no longer defaults the FAMILY when targeting a non-DISPLAY device.
2. `(PRINTERTYPE output-file)` is used for defaulting the `TitleFont` `DEVICE` in the `FONTCREATE`; It doesn't return a value that is useable by `FONTCREATE`.
3. `FontSample` uses `SLUGCHARSET` to locate the `CHARSETINFO` of all slugs. This no longer exists.

So:
1. How should the font `FAMILY` be defaulted _by DEVICE_?
2. How to get the correct DEVICE for the `FONTCREATE` based on the specified output location (i.e., `Printer` parameter), which may be a printer designation or a file? (There _is_ a `StreamType` parameter, which, currently, may be `NIL` to "figure it out from the `Printer`.)
3. The location of the `CHARSETINFO` of the font is being moved to a field of the `FONTDESCRIPTOR` (@rmkaplan), changing to this should resolve this.

There may be other issues that show up once these are addressed.

rmkaplan:
In terms of defaulting the file type from the printer (point 2), I think one new way of doing that would be to replace (PRINTERTYPE Printer) with

(CAR (PRINTERPROP Printer 'CANPRINT))

This is because we are distinguishing between the various types of printers and the imagefile types that they support.  Various Interpress or PDF printers may have more specific constraints or capabilities beyond just the types of files they can interpret.  DOCUPRINTER and FAX can both do INTERPRESS, but not exactly in the same way (I presume).

Alternatively, if the TitleFont is moved to after the Stream is set in the LET*, I think the open stream can be passed as the device argument to FONTCREATE.  The CANPRINT coercion would then have to move to the OPENIMAGESTREAM.  That seems a little bit more logical to me--get the stream, then get the font for the stream.

In order for the Printer argument to be interpreted as a printer and not just a file, I think it would have to be specifiedas {LPT}... or maybe {VIEWER}.  Is that the scheme?

MattHeffron:
> In order for the Printer argument to be interpreted as a printer and not just a file, I think it would have to be specified as {LPT}... or maybe {VIEWER}. Is that the scheme?

[...]

PR #2717, opened 2026-08-04
Make FONTSAMPLER compatible with new printer interface for imagetype

rmkaplan:
This is in response to the issues raised in #2714 

1.  If a printer is provided but no streamtype, I set the streamtype be the first CANPRINT of the printer. (However, this may need a little more work to grab the more specific type from a name like {LPT}FOO.PDF).

2.  I reordered the variables so that the TitleFont and the other fonts grab their type from the image stream (no matter how its type is derived).

3.  I made it so that fonts can be specfied as FONTSPECS or FONTDESCRIPTORS.

4.  I made it so that the character sets can be specified by their names (Greek) as well as numbers.

5. I changed some numbers to constants (\MAXCHARSET PTSPERINCH)

6. I installed the SlugCharsetInfo in the FontDescriptor.

7.  I cleaned up the code a little bit.

I am starting this off as a draft so you can look at the code.  This is because the slug charset info depends on that change to FONT and the associated recompiles.  And separately, there is a bug in OPENIMAGESTREAM that is fixed in open PR #2716, but the {LPT} case won't work until that PR is merged.


pamoroso:

pamoroso:
When printing to the `POSEIDON:` Interpress printer of Dodo is this supposed to work?

```
9_ (FontSample '(CLASSIC 10) 0 "POSEIDON:")
No open function for 
NIL
```

rmkaplan:
No, but it should work if you say {LPT}POSEIDON:.  Otherwise it thinks that POSIEDON: is a file name with a colon (which maybe interpreted as a filedevice with no filename or extension).  That argument to FontSample is named "Printer", but I think it is really just the name of an output imagefile.  

(But as noted, the {LPT} behavior depends on the bug fix in the Hardcopy PR)

rmkaplan:
I took out the special CANPRINT test for StreamType.  That argument is just passed as the IMAGETYPE to OPENIMAGESTREAM, OPENIMAGESTREAM figures out the intended type, and the fonts then get that type from the image stream.

masinter:
given that both the common lisp and Interlisp execs use a readtable that uses : as the package delimiter (basically every rdtbl other than OLD-INTERLISP-T and OLD-INTERLISP-FILE), I'd suggest making all examples of file names use strings 
   "{FS1::}<medley>lispusers>example.ext"

or pathnames 
   #P"{FS1:}<medley>lispusers>exam;ple.ext"

instead of LITATOM / SYMBOL 
   FS1%:<medley>lispusers>example.ext

and avoid interning all file names in the Interlisp package when possible.

The {DSK} file system code in maiko unfortunately uses IL: symbols in many of its subrs, but perhaps we could eventually deprecate such use.

rmkaplan:
I updated FontSamplerFaked, also to let OPENIMAGESTREAM resolve the imagetype.

What's going on may be clearer if the Printer and StreamType arguments were renamed to Imagefile and ImageType (keeping the camel case), following the other imagestream conventions.  It becomes a printer only if the host is {LPT} (or NIL).  But I haven't made that change.

rmkaplan:
I went ahead and changed the argument names Printer to ImageFile, StreamType to ImageType.

Also updated the documentation.

BTW, the first 2 rows/columns for Modern and Classic, at least, look odd to me.  They have accented characters that are not in XCCS/MCCS.

nbriggs:
Hmmm... FontAsList in FontSampleFaked should probably not have been changed to a FONTSPEC -- the whole point of FontSampleFaked is that you can create a font sample for a font that we don't have, and therefore FONTCREATE will never work.  Because of the way the Interpress code works it will just stuff the name you gave it into the Interpress master and then the printer will do the right thing if it has the font installed.

rmkaplan:
But it really is a fontspec, that's the defined format for such a list.  The trick is that it is not actually passed to fontcreate, its fields (except for device)  are used to modify a font that is separately created for that imagetype

MattHeffron:
Which unmerged PRs are prerequisites for testing this?
Other than the PR #2716 you already mentioned, if any.

pamoroso:
Still looking good at commit 269a130.

rmkaplan:
Actually, if you want to make it clear that this list is unrelated to the normal way of specifying fonts (a fontspec), the family, size, and face should just be their own separate arguments to FontSampleFaked.

rmkaplan:
I eliminated a bogus free variable in FontTable, made the change mentioned above of having separate family/size/face arguments to FontSampleFaked.

pamoroso:
Still no issues at commit c0e1f04.

masinter:
would it make sense to add a PDF (or HTML) format output for font files in HCFILES? So we can convert font files as we do TEDIT and SKETCH and Lisp sourcces?

pamoroso:
Yes, PDF output for font files would be useful.

rmkaplan:
You could add FONTFILE as an entry in PRINTFILETYPES, with MEDLEYFONT.FILEP as the test function and a new function FONTFILE.TO.IMAGEFILE as the conversion function.  That would be defined as an interface to FontSample, and you would have to make sure that FONTSAMPLER.LCOM has been loaded.

But, do you really want to produce image files for all of the fonts and their character sets on every release?  Or just once, for an archive.

MattHeffron:
I just tried this new FONTSAMPLER, which I copied into a PR 2718 local repo with fresh loadups.
I compiled (`TCOMPL`) the FONTSAMPLER source.
I also copied the new DANCER MEDLEYFONTFILES (PR 2711) into this local repo.
I ran `(FontSample (FONTCREATE 'DANCER 14 'MRR 0 'DISPLAY) 0 'DISPLAY 'DISPLAY)`.
This showed a window with the DANCER glyphs, _**except**_ those in the high and low _control_ character ranges.

Trying: `(FontSample (FONTCREATE 'DANCER 14 'MRR 0 'DISPLAY) 0 "{MATT}DANCER.PDF" 'PDF)` gives a valid PDF file that has _no relation to_ the DANCER font. 
[DANCER.PDF](https://github.com/user-attachments/files/30759424/DANCER.PDF)

So, **for the "skipped" glyphs**, considering that some (_non-MCCS_?) DISPLAY fonts (e.g., DANCER) use glyphs in the ranges excluded by the `RangedCodesStreamType` checking, is it really appropriate to exclude those ranges for DISPLAY fonts?
(Or, at least, or apply this only to INTERPRESS fonts?)

**For the PDF output issue**, there are still some issues. It seems to be recreating the font specified in the image stream type; constructing `FontList`. This is **not correct**.

MattHeffron:
This file probably should have a `FILETYPE` property specifying `:TCOMPL` so `CLEANUP` will do the right thing independent of the setting of the default compiler.

rmkaplan:
I believe that DANCER 14 doesn't exist for the display, so it should have given you a font-not-found error then.   DANCER 12 exists.

When I do your first example (DISPLAY imagetype) for size 12, I also do see that the first rows are blank.  That seems to be controlled by a variable RangedCodeStreamType, that's True for DISPLAY and INTERPRESS.  There is a predicate SLUGCHARP, I think that would in general be a better per-character test.  Although, what's the harm in showing black boxes?

But changing this to NIL for DISPLAY shows more stuff in those rows, but not the same stuff that EDITFONT shows.  So that's another problem

On the other example, showing a display font in a pdf file, is the issue that the type of the font is different than the type of the stream?  Showing Interlpress fonts on a pdf printer?

nbriggs:
I think we're losing sight of the original intent -- which was to produce sample sheets of the native fonts installed on a printer, and they were almost exclusively Interpress printers, though of different models.  There was very little remapping going on.  It could also generate sample sheets for fonts which Interlisp did not have width files for (and thus could not speak the name of via the normal font machinery).  The output was similar to the sample pages in the XCCS specs, and the book of font samples that we used to keep by the Press printers.  If I were doing it for Postscript it would attempt to generate a sample based on the encoding vector the font used -- like E.6 StandardEncoding or ISOLatin1Encoding, pages 598-599 of the PSLRM 2nd edition, or the Symbol encoding vector, page 606, for the Symbol fonts.  As with the old display fonts, the characters you got for a particular character index in the font depend on the font.

MattHeffron:
> I believe that DANCER 14 doesn't exist for the display, so it should have given you a font-not-found error then. DANCER 12 exists.

But I copied the new MEDLEYDISPLAYFONT files from PR #2711, so it **IS** there.

Showing a DISPLAY font in non-DISPLAY output (i,e., not to a WINDOW) is a specific special case in `FontTable`. See `UseDisplayFontBitmaps`. This is the mechanism to make a Font Sample sheet for DISPLAY fonts into other (e.g., PS, PDF, INTERPRESS) output formats.

MattHeffron:
I think that the change in ordering of glyphs is due to the new fonts (and their source BDF files) being marked as MCCS encoding, but not actually using that encoding. The attempt is to preserve the original charcode to glyph mapping. What should be in the MEDLEYFONTFILE (and FONTDESCRIPTOR) to indicate a non-standard encoding, i.e., the glyphs are already where they're "supposed" to be?

rmkaplan:
Ahh, I wasn't running in that PR.  I copied those fonts up into my working directory (where files are not segreated by PRs), and I looked at them with EDITFONT.  The replacements for 10 and 12 are a bit smaller than what we got from John's directory.  Is that intentional?  John's current dance files would look a little different.

The other sizes look good.  So if you confirm that you wanted to change to 10 and 12, I'll approve this.

rmkaplan:
I don't think the font charencoding should make a difference, one way or the other, in this context.  There is no further MCCS-related  transformation going on here.  So I think what we are seeing with FontSample is still unexplained, at least when we are going to a window and using the display glyphs. SWe have to look more carefully at what it is doing  with control characters in the display-to-window case, for starters.

It is a separate question (touched on briefly in the meeting today) whether we want to assign currently unused MCCS codes to the special Dancer characters, and provide a mapping table for those.  Then the glyphs would move to new positions as part of either your off-Medley construction or the IMPORTFONTS procedure given your unmapped source files (the current ones, except that they would start out with DANCER as the encoding).

Either way, we would want to do the completion step of the importing, so that DANCER would have (probably by indirect reference) all of the characters that we have from some other appropriate font (Classic of the same size and face?).  And other fonts would be completed (by indirect reference) with the new Dancer stuff.  So a choreographer would be able to pick any font and just type--getting the Dancer symbols in the middle of ordinay typing would be a switch-keyboard or virtual keyboard issue.

The mapping table would be easy to construct--keep the alphabetics and any other punctuation where they are, move everything else into some unused space.  We would also interface to the routine in Tedit that notices when a file is opened that document characters in a given font also need to be translated to their MCCS positions.

MattHeffron:
> The replacements for 10 and 12 are a bit smaller than what we got from John's directory. Is that intentional? John's current dance files would look a little different.

I hadn't compared them with John's DISPLAYFONT files. As I mentioned a couple of times, I used FontForge to create a TTF, then `otf2bdf` to make BDF files in various sizes, then my READ-BDF to make MEDLEYDISPLAYFONT files.
I had `otf2bdf` use 72 pixels per inch (approx. 1 pixel per point) to make the BDF.
I could use a larger value to cause different scaling in the BDF fonts, and thus the MEDLEYDISPLAYFONT files.
I have no "emotional attachment" to the new DANCER MEDLEYDISPLAYFONT files. If the original look better at those sizes, great.
Are _his_ size 10 closer to _my_ size 12, and _his_ 12 to _my_ 14?

> So I think what we are seeing with FontSample is still unexplained, at least when we are going to a window and using the display glyphs.

The window display of the FontSample display glyphs, appears essentially equivalent to viewing the source PostScript font in FontForge, from which all of the new DANCER font files are based.

rmkaplan:
Below is what EDITFONT shows for all of the Dancer display fonts.  The bottom row is the sequence of fonts that you created.  The two on top are John's fonts of the 10 and 12 nominal sizes.  Not clear how to reconcile.

I looked at the (size height) pairs for the new fonts:
         (8 8) (9 9) (10 11) (11 11) (12 12) (14 15)
It's a little odd that 10 and 11 have the same height, but maybe that's round-off.  Here are the pairs for John's two fonts: 
     (10 13) (12 16)
Maybe he was worried about a discrepancy between hardcopy and display, the size mismatch problem we have for NS fonts?

Screenshot 2026-08-06 at 12 56 33 AM

rmkaplan:
One thought:  Based on the heights, maybe we ought to include sizes 13 and 16, either newly constructed or John's fonts renamed.  There is a variable *DANCER-FONT-SIZE* in DANCEROBJ, initially 12.  Maybe then changing that to 16 would preserve appearance of his existing files, at least on the display.

masinter:
@rmkaplan wrote https://github.com/Interlisp/medley/pull/2717#issuecomment-5196486254
You could add FONTFILE as an entry in PRINTFILETYPES

Is PRINTFILETYPES an enumeration of input formats or output formats? This doesn't seem right.


> But, do you really want to produce image files for all of the fonts and their character sets on every release?  Or just once, for an archive.

It's an option in do-hcfiles to not remake files that have already been converted.

Turned off for now when we're still debugging.

rmkaplan:
It's an enumeration of image output formats, with conversions from image sources.  I would have renamed the variable to make it clearer, but its documented

MattHeffron:
None of this, below, is _directly_ related to FONTSAMPLER, but it shows what the glyph mapping _should_ be and the _intended_ appearance of the glyphs (which look pretty poor at 1 pixel per point resolution).

I hand modified a .PS file of an older FONTSAMPLER output, to see what the PostScript Type 1 font generated.
(Sorry, that I used a file that reads across vs. reads down. That's the way FontForge shows it, but I didn't choose for that reason.)
Here's [DANCER12.ps.txt](https://github.com/user-attachments/files/30807867/DANCER12.ps.txt) (added `.txt` because GitHub)
And here's the generated [DANCER12.pdf](https://github.com/user-attachments/files/30807891/DANCER12.pdf)
I _did_ have to modify the original Type 1 font file:
I commented out the `serverdict begin 0 exitserver` (line 3) because `ps2pdf` (a.k.a. `ghostscript`) failed with a security issue. (I'm guessing that it's only required if **installing** the font on a PostScript printer.)
I also removed the `EOT` control char at the end.
FontForge had given a warning about the size of an EN space (1600 vs. 1024).
I noticed in the font file:
`/FontMatrix [0.000625 0 0 0.000625 0 0] def`
And that the `0.000625` is 1/1600. I replaced that line with:
`/FontMatrix [0.0009765625 0 0 0.0009765625 0 0] def`
where `0.0009765625` is 1/1024.
That scaled the renderings up a bit, maybe closer to what the old DANCER DISPLAYFONT files had. I'll try generating new BDF files and then new MEDLEYDISPLAYFONT files. (I'll wait until FONTSAMPLER works, so we can get before and after for comparisons.)
Here's the updated font file: 
[Dancer.pfa.txt](https://github.com/user-attachments/files/30808402/Dancer.pfa.txt) (added `.txt` because GitHub)
I "installed" the font by creating a symbolic link in `/usr/share/ghostscript/fonts/Dancer` linked to `/home/matt/Dancer.pfa`.
I used ghostscript directly, insead of `ps2pdf` because I was debugging the issues with the font file.
```shell
gs -dNOPAUSE -dBATCH -sDEVICE=pdfwrite -sOutputFile=DANCER12.pdf DANCER12.ps
```

rmkaplan:
I've lost track of what's currently wrong with FONTSAMPLER.  Is it the fact that it tries to avoid putting out black boxes for the control codes?  I don't think that's anything new.  (I mentioned SLUGCHARP as a predicate, but I think that is implemented only for the display--it would be good to have a device-independent way of recognizing slug characters.)

Or is the problem just with the Dancer fonts, or with UseDisplayBitmaps, or something else?

rmkaplan:
The FontList fonts now retain their original device, not coerced to the imagestream type.  That seems to make the bitmap work for display fonts.  Was that it?

pamoroso:
Still looking good at commit fa119a5.

lispusers/FONTSAMPLER
```diff
@@ -1,110 +1,112 @@
...
+                                                  (* ;; "Forcibly install ALL CharacterSets.")
+
+                                                  (for CS from 0 to \MAXCHARSET
+                                                     when (\INSURECHARSETINFO Font CS) collect CS))
+                                            (:INTERESTING (for CS in (CHARSET.DECODE 
+                                                                            *INTERESTING-CHARSETS*)
+                                                             when (\INSURECHARSETINFO Font CS)
+                                                             collect CS))
+                                            ((T :INCORE) 
+                                                 (for CS from 0 to \MAXCHCLEARSET
```
MattHeffron:
```suggestion
                                                 (for CS from 0 to \MAXCHARSET
```
was `\MAXCHCLEARSET` a typo, or is it defined in another PR that's needed?

MattHeffron (commented):
typo?

rmkaplan:
Sorry, fixed

MattHeffron:
This now works as I expect for a DISPLAY font to PDF.
If it has no other issues (INTERPRESS, network, etc.), remove DRAFT status and I'll approve it.

pamoroso:
Still no issues at commit 7c36b92.

rmkaplan:
OK, now ready

MattHeffron (approved):
Approved.

PR #2718, opened 2026-08-05
SLUGCSINFO is a field of FONTDESCRIPTOR

rmkaplan:
I made the switch in the SLUGCSINFO macro, and recompiled all of the affected files (those that contained functions calling \BUILDSLUGCSINFO, also scanned).

This included HARDCOPY, which then got mixed up with the versions of HARDCOPY and IMAGEIO in #2716, so I included IMAGEIO here too.  #2716 will be deleted.

The new version of FontSampler should run in this environment.


pamoroso:
I've found no issues so far on Linux Mint 22.1 Cinnamon.

MattHeffron (approved):
Assuming that this has _caught_ all of the files that needed to be recompiled, it works and I approve this.

MattHeffron:
Aside: This demonstrates why we _ought_ to be able to recompile everything...

Issue #2727, opened 2026-08-14
Lost non-0 rotation on display streams.  Font issue?


nbriggs:
In the Venue FULL sysout if you do:
```
(SETQ G90 (FONTCREATE 'GACHA 10 'MRR 90 'DISPLAY))
(SETQ W (CREATEW '(100 100 200 200)))
(SETQ STRM (WINDOWPROP W 'DSP))
(DSPFONT G90 STRM)
(DSPXPOSITION 50 STRM)
(DSPYPOSITION 50 STRM)
(PRIN1 MAKESYSDATE STRM)
```
you get:

<img width="226" height="223" alt="Image" src="https://github.com/user-attachments/assets/66594539-c30a-479d-8ea5-4de0d7126fac" />

however if you do this in a newly built Medley full sysout you get:

<img width="224" height="224" alt="Image" src="https://github.com/user-attachments/assets/1df7a712-c1ac-4443-8299-156cd3fa855e" />

This is a serious regression.

rmkaplan:
The problem stems from the fact that rotation is not a property of font-file names, so the new, generic lookup for the font file was ignoring the rotation and coming up with the unrotated file.

But I had noticed quite awhile ago that the rotation coercion was only allowed for 90 and 270, but not for 180, not clear why.  The Lisp display code seems to do fine with 180, it inverts the bitmap as you would expect.

But that seems to fail in \MAIKO.BLTCHR, which produces an error in \SLOWBLTCHAR.  Is that the expected result?  If so, I'll keep the error earlier when an upside-down font is requested on the Lisp side.

rmkaplan:
As noted, the PR  #2735 fixes the 90 degree case, 270 fails at a deeper level

rmkaplan:
Actually, I don't understand how this works.  The only change to the font is the rotation of the bitmap.  The widths, offsets, imagewidths are the same for the 0 and 90 fonts, and the ywidths are NIL for both.  So, is Maiko branching on the fact that the ROTATION field is 0 or 90?

EDITFONT doesn't appear to work on a 90 rotated font, I haven't tried fontsampler.

pamoroso:
On Linux Mint 22.1 Cinnamon it's working as expected.

nbriggs:
[...]

PR #2735, opened 2026-08-19
Restore display font rotation for 90 degrees

rmkaplan:
This addresses issue #2727. 

It rotates the bitmaps for display fonts that are specified as 90.

It gives an error inside the commonlisp function ROTATE-BITMAP when given a 270 as the rotation.  I apparently don't know how to debug a commonlisp function, even when run interpreted the variables are visible on the stack only with string names, not atoms.

And the error may be inside the Pilot BBT machinery, which I know nothing about.

So I'm submitting this as at least a partial improvement--I think the font-level lisp interface is straightened out.


pamoroso:
On Linux Mint 22.1 Cinnamon it's working as expected.

pamoroso:
Still looking good at commit e85b1d1.

rmkaplan:
Further adjustments for font rotation:

*  The rotation specified is coerced by (IMOD xxx 360).  (e.g. -90 → 270)
*  Non-zero rotation is shown in the Font defprint
*  \SLOWBLTCHAR code is cleaned up.  Could be extended to do 180

The 270 case works better, but still some problems with getting the right offset adjustment for slugs and when the widths and imagewidths disagree.

masinter:
<lispusers>STORAGE is an example of use of rotated fonts, if you are looking for a real case.  SHOWSTORAGE will attempt to FONTCREATE a rotated Gacha if it can. See STORAGE.TEDIT to see.

nbriggs:
The documentation for FONTCREATE says:
> If no acceptable font is found, the action of FONTCREATE is determined by NOERRORFLG.  If NOERRORFLG is NIL, it generates a FONT NOT FOUND error with the offending font specification; otherwise, FONTCREATE returns NIL.

The new code is consistent with the Venue code for unsupported rotations and faces reporting an illegal-argument error, but I'm not sure whether it should do that or report a font-not-found error as the documentation says.

Since the original Venue code did it, I guess I'd be OK leaving it, but we should note it somewhere.

rmkaplan:
The font code gives an illegal arg error if the argument doesn't make sense:  family specified as a number, or a face like (FAT SLANTED).  That error is not controlled by the NOERRORFLG.

It produces the font not found error if the arguments are of the appropriate type, but the font doesn't exist.  E.g. size 79 or family FOO. 

If the rotation is not a fixp (e.g. rotation  XXX), it should cause the type error for sure.  BUt should we consider an unimplemented rotation (87) as a bad-argument error or a font-missing error?

nbriggs:
I'm not sure about the bad rotation response.  I think leaving it as is at the moment is OK, since it's what it has done since Venue's time.

But... on a different topic:  the values returned by `(CHARWIDTH charcode font)` and `(CHARWIDTHY charcode font)` for rotated fonts do not follow the documented results:

>(CHARWIDTH CHARCODE FONT)  	[Function]
CHARCODE is an integer that represents a valid character (as returned by CHCON1).  Returns the amount by which an image stream's X-position will be incremented when the character is printed.

>(CHARWIDTHY CHARCODE FONT)  	[Function]
Like CHARWIDTH, but returns the Y component of the character's width, the amount by which an image stream's Y-position will be incremented when the character is printed.  This will be zero for most characters in normal portrait fonts, but may be non-zero for landscape fonts or for vector-drawing fonts.

```
(SETQ F90 (FONTCREATE 'GACHA 12 'MRR 90 'DISPLAY))
(SETQ W (CREATEW '(200 200 200 200)))
(DSPFONT F90 W)
(DSPXPOSITION 100 W)
(DSPYPOSITION 100 W)
(PRINTOUT W "*")
(EQP (CHARWIDTH (CHCON1 "*") F90) (IDIFFERENCE (DSPXPOSITION NIL W) 100))
(EQP (CHARWIDTHY (CHCON1 "*" F90) (IDIFFERENCE (DSPYPOSITION NIL W) 100))
```
Both EQPs should return T, as they do when the font is rotation 0.

The stream X and Y positions after printing are correct (and in the landscape case the text advances up the window as it should).  The width values should have been interchanged for the 90 degree rotation, but were not.  Not sure why the output to the stream gets the positioning correct when the x,y escapements are incorrect.
I haven't checked a 270 degree rotation - I'd guess it has the same problem.

rmkaplan:
As I mentioned in the meeting, I don't think any of the Lisp-level font-related functions, other than \SLOWBLTCHAR, do the right thing for rotated fonts.  Even in the Venue sysout.

The rotation code just rotates the bitmap and fixes the offsets (in the 270 case), but otherwise doesn't fiddle any of the other metrics.  The WIDTH and YWIDTH vectors, for example, are not touched, even though you might think they should be swapped. CHARWIDTHY continues to return 0.

I think it's only \SLOWBLTCHAR that branches on the rotation so as to behave as if those vectors had been properly updated.  These rotated fonts can be used for printing, but otherwise they are not really real.

I have spent a little time looking at GETCHARBITMAP, seeing what it would be like to do the same rotation-dependent faking there.  I think that would enable EDITFONT and maybe FontSample to work.

One change is to calculate the height of the character bitmap as the ascent+descent rather than taking the (now very large and irrelevant) height of the full rotated bitmap.

A more general solution would be for the rotating code to update the widths and ywidths vectors.  But for the 270 case that runs into the problem that the ywidths would be negative in a 16-bit unsigned field (\SLOWBLTCHAR does the IMINUS).

A side note:  there is also an undocumented operation DSPROTATE, authored by hdj in 1985, that presumably would simply rotate the coordinates of the stream.  Unfortunately, it doesn't appear to have a display implementation.

rmkaplan:
SHOWSTORAGE works.

rmkaplan:
I updated the rotated display fonts so that the YWIDTHS and IMAGEWIDTHS are more accurate (although the negative values are not represented).  GETCHARBITMAP and \SLOWBLTCHAR  should work, and maybe a few other standard font functions (CHARWIDTHY?).  Others may still not provide the expected results, but probably not worth tracking down.

I think this is ready for further testing and hopefully merging.

nbriggs:
I checked against the Venue sysout and it has the same problem with the CHARWIDTH and CHARWIDTHY, so I guess we should go ahead and merge this.  I verified that the Maiko BLTCHAR code appears to punt on all calls to place a rotated character.  In the unrotated case the Maiko code is responsible for advancing the display data's DDXPOSITION for which is uses the C equivalent of (from \SLOWBLTCHAR)
```
          (freplace (\DISPLAYDATA DDXPOSITION) of DISPLAYDATA with
                      (IPLUS CURX (\DSPGETCHARWIDTH CHAR8CODE  DISPLAYDATA)))
```

nbriggs (dismissed):
FIxes the reported problem.  Now operates the same as the older Venue sysouts.

nbriggs:
Hmmm... actually not sure about this -- it leaves CHARWIDTH of the 90 degree rotated font non-zero, but the definition of CHARWIDTH is that it is the amount by which the DSPXPOSITION is advanced when the character is displayed, and the X position does NOT advance when a 90 degree rotated font is displayed.

rmkaplan:
The width could also be changed so that it returns (IPLUS ASCENT DESCENT) on every character.  But if the width is supposed to represent the advance, it would have to be negative for one of the rotations.  And we currently can't represent that.

So that's where I stopped.  More things work than before, but some things still don't.

rmkaplan:
Actually, my last message was confused.  The CHARWIDTH could be made to return 0 on every character, that would be correct--the XPOS doesn't move.  The YWIDTHS is what would have to be negative in the 90 degree case.

nbriggs:
Wouldn't the Y widths be positive in the 90 degree case?  The coordinate system for an image stream (that is a window) is (0,0) in the lower left, increasing to the top right.

rmkaplan:
Yes, 90 has always been the easy case.

Latest commit zeros out the WIDTHS vector.

rmkaplan:
If the widths as well as the ywidths could be interpreted as signed integers, we could also run Hebrew backwards.  Is 32K large enough for the widths of hardcopy font sizes?

nbriggs:
I don't know what's required for the hardcopy font widths.  I do know that Modern 36, which has the largest point size of any Interpress fonts I can see, has a CHARWIDTH for "_" of 1270 -- so +/- 32K would likely be fine.  Do you know of any likely larger fonts?  I was a bit surprised that `(FONTSAVAILABLE '* '* 'MRR 0 'POSTSCRIPT T)` returned NIL.

pamoroso:
Nothing unusual to report at commit c0696a1.

nbriggs:
Found another bug.  Here's the test code I'm using --
```
[DEFINEQ (DRAWALL (LAMBDA (W F0 TEXT)
       (WINDOWPROP W 'TITLE (FONTPROP F0 'SPEC))
       (CLEARW W)
       (DSPOPERATION 'PAINT W)
       (DRAWIT F0 W TEXT)
       (DRAWIT (FONTCREATE F0 NIL NIL 90)
              W TEXT)
       (DRAWIT (FONTCREATE F0 NIL NIL 270)
              W TEXT)
       NIL)
]

[DEFINEQ (DRAWIT (LAMBDA (F W TEXT)
       (LET ((CX (IQUOTIENT (WINDOWPROP W 'WIDTH) 2))
             (CY (IQUOTIENT (WINDOWPROP W 'HEIGHT) 2))
             (TEXT (OR TEXT "TESTING")))
            (DSPFONT F W)
            (DSPXPOSITION CX W)
            (DSPYPOSITION CY W)
            (PRINTOUT W TEXT)
            (LET ((X (DSPXPOSITION NIL W))
                  (Y (DSPYPOSITION NIL W)))
                 (PRINTOUT W .I4 X .I4 Y))
            (BITMAPBIT W CX CY 1)))
]
```
I'm using a window, `(SETQ W (CREATEW '(200 200 508 508)))`, and doing, for example, 
        `(DRAWALL W (FONTCREATE 'GACHA 12))`
but what I discovered was that while it worked OK for GACHA 12 and CLASSIC 12, the characters that are draw in non-0 rotation for larger sizes such as CLASSIC 24 have extra bits... I've highlighted the ones that are obvious to me in red here (and I now see I missed some between the E and S in the 90 degree case):
Screenshot 2026-08-24 at 14 47 03

nbriggs:
It's also pretty obviously buggy if you look at `(EDITFONT (FONTCREATE 'CLASSIC 24 'MRR 90))`

nbriggs:
Each character has on its top (right, before rotation) a few bits of the character originally on its left.
image

rmkaplan:
It needed the symmetry to be completed by saving the original IMAGEWIDTHS as a YIMAGEWIDTHS vector, and using that as the IMAGEHEIGHT if it exists (and using ASCENT+DESCENT if the vector doesn't exist--unrotated).

Rather than adding another field to the charsetinfo, I put it on the charset's property list. So, a little bit slower.

With this change I don't see the extra pixels you saw for Classic 24.  The only glitch is that there seems to be a single pixel turned on at the center of the 3 different output strings.  Is that from the BITMAPBIT at the end of your DRAWIT ?

nbriggs:
Yes, the pixel in the center is the BITMAPBIT so that I could tell where the origin was.
Does the data need to be copied, or does the code just need to know whether it's a rotated font or not and use the entry from the existing vector in the right field of the bitblt/pilotbbt?  If the field didn't previously exist then no other code is relying on it.

rmkaplan:
For the rotated fonts, \SLOWBLTCHAR calls BITBLT directly, not part of the finagling it does in the slow case for unrotated fonts.  So I don't think it matters whether the vector is copied or not.

rmkaplan:
I extended it to support 180 degree font rotations.  To test it, in your DRAWALL, put in a call to DRAWIT for rotation 180.

pamoroso:
At commit e96cabc 180° rotation seems to work.

font-rotation

pamoroso:
At commit 396c495 the `NOERRORFLG` argument seems to work as expected.

nbriggs:
Looks good.  One small concern I have about how the widths are handled is that if we wanted to extend the fast case code in the Maiko BLTCHAR (and friends)  implementation, it depends on the fields in the DISPLAYDATA structure and the CHARSETINFO structure being correct for the font/character it's being asked to BLT -- in particular it knows about these in DISPLAYDATA:
- ddcharheightdelta
- ddcharimagewidths
- ddcharset
- ddcharsetascent
- ddclippingbottom
- ddclippingleft
- ddclippingright
- ddclippingtop
- dddestination
- ddfont
- ddleftmargin
- ddoffsetscache
- ddpilotbbt
- ddrightmargin
- ddslowprintingcase
- ddwidthscache
- ddxoffset
- ddxposition
- ddyoffset
- ddyposition

and these in the CHARSETINFO:
- charsetascent
- charsetbitmap
- charsetdescent
- widths
- offsets
- imagewidths

nbriggs (approved):
Looks good now.  If we decide to add cases to the fast path for BLTCHAR then we can revisit the Lisp side of the code at that time.

rmkaplan:
I tried swapping the ASCENT/DESCENT fields in the 90 degree case, to get a little closer.  But it shifted more than I expected and I didn't think it was worth the time to understand it.

The other issues include making sure that the vectors are treated as signed ints, having the font-changing code move the YIMAGEWIDTHS into a new field in the DISPLAYDATA structure, and understand whether the ADD1 can be be moved into the data.

I agree it would be cleaner to push this through, but for now at least we have something that works.

nbriggs:
I haven't looked at the code for STRINGREGION, but it's not right for anything other than an unrotated font.  I updated DRAWIT to show the failure - it should be drawing a gray background for each piece of text and then painting the text onto that:
```
[DEFINEQ (DRAWIT (LAMBDA (F W TEXT)
       (LET ((CX (IQUOTIENT (WINDOWPROP W 'WIDTH) 2))
             (CY (IQUOTIENT (WINDOWPROP W 'HEIGHT) 2))
             (TEXT (OR TEXT "TESTING")))
            (DSPFONT F W)
            (DSPXPOSITION CX W)
            (DSPYPOSITION CY W)
            (LET ((SREG (STRINGREGION TEXT W)))
                 (BLTSHADE GRAYSHADE W
                        (fetch (REGION LEFT) of SREG)
                        (fetch (REGION BOTTOM) of SREG)
                        (fetch (REGION WIDTH) of SREG)
                        (fetch (REGION HEIGHT) of SREG)
                        'REPLACE))
            (PRINTOUT W TEXT)
            (LET ((X (DSPXPOSITION NIL W))
                  (Y (DSPYPOSITION NIL W)))
                 (PRINTOUT W .I4 X .I4 Y))
            (BITMAPBIT W CX CY 1)))
]
```

nbriggs:
p.s., drawing just the 0 and 90 degree case in the Venue full.sysout produces (for Classic 12):
image
so it produced the same region for the rotated font as for the unrotated one, and therefore blotted out the unrotated text instead of laying down the background for the rotated text.

rmkaplan:
I'm sure that STRINGREGION never worked.  As noted before, rotated display fonts were incompletely implemented, really just a hack for displaying.

If we want to take this on as a project, we would have to add other Y-direction functions that parallel (correctly) the X functions and make sure they work for rotated/unrotated fonts.

For example, we would make sure that CHARWIDTHY (which maybe should be an alias of CHARHEIGHT) uses the proper data, and then we would define STRINGHEIGHT in terms of CHARHEIGHT (CHARWIDTH does open coding for speed, not necessary).   The generic STRINGREGION would call STRINGWIDTH and STRINGHEIGHT to get the region (assuming the ascent and descent have been switched to get the left/bottom of the region).

There may be other functions and macros that would have to be updated or generalized.  Conceptually not very difficult, but as I said, a project.

PR #2739, opened 2026-08-21
Fix typo causing undefined function error

### LINKED ISSUE #2736: Undefined function error when looking up string in LOOKUPINFILES window
Looking up a string in a window created with the LispUsers module LOOKUPINFILES yields this error:

```
UNDEFINED-FUNCTION-IN-APPLY
In OLDFAULT1:
OPENSTREAMSTRING is an undefined function.
```

Image

To reproduce:

1. launch Medley
2. load LOOKUPINFILES with `(FILESLOAD LOOKUPINFILES)`
3. evaluate `(MakeLookupWindow)`
4. middle-click on the resulting icon
5. type `foo` and press `ENTER`

A stack trace is not necessary as the error is caused by a typo in LOOKUPINFILES. Function [`ShowLookUpString`](https://github.com/Interlisp/medley/blob/634d092ed802314ada447878418686fbfeeb3048/lispusers/LOOKUPINFILES#L478) contains a [call to the nonexistent function](https://github.com/Interlisp/medley/blob/634d092ed802314ada447878418686fbfeeb3048/lispusers/LOOKUPINFILES#L507) `OPENSTREAMSTRING`. Renaming `OPENSTREAMSTRING` to `OPENSTRINGSTREAM` fixes the issue.

I'm using Medley rmk222--Fix-displayfont-rotation/48a57af and Maiko master/c5d40e5 on Linux Mint 22.1 Cinnamon but the issue occurs also in current Medley Online.

pamoroso:
This change resolves #2736 by fixing a typo that cuases an undefined function error when looking up a string in the LispUsers module LOOKUPINFILES.


rmkaplan (approved):
I think this addresses the typo problem and the Tedit-internals issue, so I'll approve.

But I'll also note that there is a function TEDITSTRING that might be used in ShowLookUpString instead of the spelled out steps here to open a Tedit on a text provided as a string.   Up to you, whether it is worth making the change.

pamoroso:
Thanks, for the time being I'll leave the code to open TEdit.

PR #2742, opened 2026-08-26
Improves  tracking in Tedit windows, Tedit follows WHENOPERATEDONFN documentation

rmkaplan:
This PR primarily has to do with image objects in Tedit, and in particular the image objects in the Tedit windows.

As the mouse tracks overs the menu with the left/middle button held down, it now moves from object to object, showing what the next state of a new object will be and restoring the existing state of the prior object in the tracking.  (With the current exception that if the mouse goes over the label of a type-in field, the tracking to future objects requires another click.)  Essentially, the behavior of the menu objects has been moved from the WHENOPERATEDON-SELETED protocol to the BUTTONEVENTINFN.

The WHENOPERATEDON-SELECTED has been changed to honor the documentation.  Previously, it would only apply the selected function if the selection contained just a single object.  Now, if an extended selection contains several objects, the function of each object is applied.  Thus the selection can be rejected if any object disallows it.


rmkaplan:
Actually, I should have emphasized Tedit *menu* windows, that's where the testing focus should be.

pamoroso:
My initial testing on Linux Mint 22.1 Cinnamon shows no issues.

pamoroso:
Still looking good at commit 48a3775.

pamoroso:
Nothing unusual to report at commit 02b20b4.

fghalasz:
Tested this PR relative to Issue#2734 (Notecards Document Card creation fails).  This PR fixes that issue.  Closing unmerged PR#2743, which had a similar fix to Issue#2734 in \TEDIT.INSTALL.PIECE.

pamoroso (approved):
Looks good to me.

PR #2743, opened 2026-09-04
Fix Issue 2734 (Notecards Document card create fails) by fixing logic error in \TEDIT.INSTALL.PIECE

### LINKED ISSUE #2734: Not an array error when scrolling Document card of DEMO.NOTEFILE in NoteCards
When a Document card of DEMO.NOTEFILE in NoteCards is scrolling automatically after being created I get this error:

```
SIMPLE-ERROR
In %%ARRAY-BASE:
Not an array: #<IO Nodircore Stream/776, 143400>
```

Image

To reproduce:

1. launch Medley
2. activate NoteCards
3. open DEMO.NOTEFILE
4. from the NoteCards badge, middle-click `New Cards` and select `Document`
5. shift-click on the title bar of the Table of Contents filebox
6. accept all the document parameter defaults but set `Build Back Links` to `Cards and Boxes`
7. click `DONE`
8. wait for the Document card to scroll automatically

The stack trace: [DOC-CARD.TXT](https://github.com/user-attachments/files/31173196/DOC-CARD.TXT)

I'm using Medley X11/VNC on Linux Mint 22.1 Cinnamon with loadup 16-Aug-2026 11:55:05 (Maiko: master/c5d40e5; Medley: master/634d092).

fghalasz:
This PR fixes Issue 2734 (Notecards Document card create fails).  The proximal cause of 2734 is a logic error in \TEDIT.INSTALL.PIECE.  Specifically,  \TEDIT.INSTALL.PIECE takes a piece as an argument. PC is set to this argument.  It then sets PCONTENTS to (PCONTENTS PC).    Under certain conditions, it then resets PC to (\TEDIT.FORMATLINE.UPDATELOOKS TSTREAM PC).  And later it operates on PCONTENTS differently based on (PTYPE PC).  But if the PTYPE of the passed in PC is different from the PTYPE of the reset PC, then the operation fails as it tries to operate on the PCONTENTS (which matches the PTYPE of the passed in PC rather than the reset (i.e., currently set) PC.

Fix is to reset PCONTENTS iff PC gets reset by \TEDIT.FORMATLINE.UPDATELOOK.

@rmkaplan.  This fixes the Notecards Document card issue.  But I have no idea how to test whether it has any negative effects on the rest of TEdit.  I'll leave it to you to check out this fix and judge whether it breaks anything (non-obvious) in TEdit.


pamoroso:
I tested this on Linux Mint 22.1 Cinnamon and it seems to resolve the issue.

On one occasion I scrolled the document window while the document was still being generated and got an invalid argument error. However, I can no longer reproduce it.

rmkaplan:
I think you are correct that the contents (and also PLEN) should be reset if the piece is changed by the subfunction call.  But my understanding is that the only time the PC returned from the subfunction would be different than the PC passed in is if there is a sequence of invisible pieces.  I haven't seen any instance of invisibility, although I have tried to preserve the code for it.  Does Notecards make use of that feature?

This fix is to the file on the master branch, but there are other changes to this file in PR #2742.  So I'll make the equivalent change on that branch, to avoid merging hassles.

rmkaplan:
I moved your fix, with some other adjustments, to the TEDIT-STREAM file in #2742 .  If that looks good, this can be closed.

fghalasz:
@rmkaplan  Notecards does indeed use INVISIBLE character looks (see function NCHACKS.MakeTEditCharInvisible) although I have yet to find where this is actually used.

In this particular case, however, the PIECE in question does not have CLINVISIBLE set.  Instead it is a piece with PLEN of 0.   VISIBLEPIECEP "returns" (its a MACRO) NIL when PLEN is 0.  Not sure how/where the PIECE of PLEN=0 originates.  I probably won't track this any further unless it would be really informative to you to know where a PLEN=0 piece gets created.

rmkaplan:
A PLEN of 0 is a little scary.  I can imagine situations where the algorithms might get confused--landing on difference pieces depending on which direction you are coming from or how the btree is laid out.  What if it begins or starts a paragraph, maybe it creates an empty paragraph?

If they are being read from a legacy file, I think I would fix the file reader to prune them out.  But it would be helpful, but not worth a lot of effort, to know if it is being created through some current operations.  Otherwise, we can wait until something else breaks.

fghalasz:
Closing this PR in favor of PR#2742, which includes this fix as well as other changes to \TEDIT.INSTALL.PIECES.   Tested PR#2742 w.r.t. Issue #2734.  Works just same as this PR.

Issue #2741, opened 2026-08-26
FONTSAVAILABLE and FONTCREATE don't agree on whether a font is available


nbriggs:
With `(SETQ DISPLAYFONTEXTENSIONS (CONS 'DISPLAYFONT DISPLAYFONTEXTENSIONS))`, and DISPLAYFONTDIRECTORIES set to a list including a directory which has (amongst other files)
```
/Users/briggs/fonts/TrilliumLogo24-c0.displayfont	/Users/briggs/fonts/TrilliumLogo30-c0.displayfont
/Users/briggs/fonts/TrilliumLogo24-mrr-c0.displayfont	/Users/briggs/fonts/TrilliumLogo30-mrr-c0.displayfont
```
doing `(FONTSAVAILABLE 'TRILLIUMLOGO '* 'MRR 0 NIL T)` reports that these fonts are available:
```
((TRILLIUMLOGO 24 (MEDIUM REGULAR REGULAR) 0 DISPLAY) (TRILLIUMLOGO 30 (MEDIUM REGULAR REGULAR) 0 DISPLAY))
```
However doing `(FONTCREATE 'TRILLIUMLOGO 24 'MRR)` results in a "FONT NOT FOUND" error.
One or other of FONTSAVAILABLE and FONTCREATE are producing an incorrect result.

The trace is attached.

[fontsavailable.txt](https://github.com/user-attachments/files/31443662/fontsavailable.txt)

rmkaplan:
It seems that fontcreate is looking for the file with -C0 in its name only in the c0> subdirectory.  But you have it in the top-level directory, so it isn't found.

But, you also have the files with and without the MRR.  Which of those do you expect it to find (if they aren't the same)?  Should the one without the face component be ignored?

rmkaplan:
There are a lot of old/new pathways in the code for mapping from fontspec+charset to the set of possible file names, and that can lead to a lot of searching.

One way of cleaning up and simplifying would be to associate a sequence of filename templates which each of the font formats on the [device]CHARSETFNS.  So the names to be considered  when looking for strike files could be different than AC files than MEDLEYFONT files.

For strike files we might have 
((c CHARSET > FAMILY SIZE - FACE -C CHARSET)
  (FAMILY SIZE - FACE -C CHARSET))

Medleyfont would just have (FAMILY SIZE - FACE).

After substituting and packing, the alternative directories/extensions would be added to produce a sequence of explicit names to look for.

[...]

PR #2745, opened 2026-09-04
Rmk225  Importing postcript fonts and other font improvements

rmkaplan:
This was instigated by the mismatch in FONTCREATE and FONTSAVAILABLE behavior for postscript and pdf fonts #2741 .  That led to a refactoring of the way that postscript pscfonts are imported into Medley,  and further work on when and where coercions (including the double coercion of pdf fonts) and scaling take place.

The fontspec-to-filename mismatches are now driven by templates that are associated with the various font-formats.  Medleyfonts don't have separate files for separate character sets, so their template is uniformly just (FAMILY SIZE - FACE).  Legacy fonts include character sets in different ways, for example (c CHARSET > FAMILY SIZE - FACE -C CHARSET).  The appropriate templates have been added to the display CHARSETFNS for AC and STRIKE files.  A number of spec-to-file functions have been removed.

With respect to postscript, the unscaled postscript PSCFONTs were created from Adobe metrics some time int he distant past (@MattHeffron ).  Previously, they were converted to fontdescriptors and scaled as needed.  On this pass the conversion to fontdescriptors is done separately and offline, in the IMPORTFONTS framework, and the resulting unscaled fonts are saved as medleyfont files.  They are read, scaled, and cached as needed by FONTCREATE.  

This PR also includes some changes to get rid of some funky or confusing names, e.g. the fields \SFAscent, \SFDescent, \SFHeight of the fontdescriptor have been renamed FONTASCENT FONTDESCENT FONTHEIGHT, followiing the pattern of most of the other field names. The other names are still there as synonyms--there are a few files that I haven't yet updated.

There were quite a few tangles that had to be worked out, particularly in the logic of font coercions.  So I'm marking this as draft until it gets exercised in a broader way.


rmkaplan:
I forgot to include the new unscaled fonts, will do that tomorrow

pamoroso:
On Linux Mint 22.1 Cinnamon, when opening lispusers/EQUATIONEXAMPLES.TEDIT I get the error:

```
INTERLISP-ERROR
In ERROR:
FONT NOT FOUND
(MODERN 6 (MEDIUM REGULAR REGULAR) 0 DISPLAY)
```

font-error

The backtrace:

```
6_: BTV
   MESS1 "FONT NOT FOUND"
   MESS2 (MODERN 6 (MEDIUM REGULAR REGULAR) 0 DISPLAY)
   NOBREAK NIL
ERROR
FONTCREATE
   FILE 
#<Input Stream on {MEDLEY}<lispusers>EQUATIONEXAMPLES.TEDIT;1/130,31700>
   TEXTOBJ {TEXTOBJ}#121,141600
   LOOKS {CL101/4954:Gacha10}
   FILEPOS 11122
   LOOKSLEN 47
   FONT NIL
   NAME MODERN
   SIZE 6
   SUPER 0
   STYLESTR NIL
   BOLD NIL
   ITALIC NIL
   BITS 0
   PROPS NIL
   PROPS ((COLOR . BLACK))
\TEDIT.GET.SINGLE.CHARLOOKS
   FILE 
#<Input Stream on {MEDLEY}<lispusers>EQUATIONEXAMPLES.TEDIT;1/130,31700>
   TEXTOBJ {TEXTOBJ}#121,141600
   I 7
\TEDIT.GET.CHARLOOKS.LIST
   TEXT 
#<Input Stream on {MEDLEY}<lispusers>EQUATIONEXAMPLES.TEDIT;1/130,31700>
   TSTREAM #<IO Tedit Stream/130,31000>
   PCCOUNT 187
   CURTEXTBYTE# 0
   END 13285
   PCNO 1
   TEXTOBJ {TEXTOBJ}#121,141600
   ORIGBYTE# 0
   PC NIL
   BYTELEN 0
   PREVPC {PIECE}#121,135710
   FIRSTPC {PIECE}#121,135710
   PARALOOKSMAP {ARRAYP}#137,63006
   CHARLOOKSMAP NIL
   DEFAULTCHARLOOKS {CL101/5062:Gacha10}
   OLDPARALOOKS {PL81/49096:LE-0-0}
\TEDIT.GET.PIECES3
   TEXT 
#<Input Stream on {MEDLEY}<lispusers>EQUATIONEXAMPLES.TEDIT;1/130,31700>
   TSTREAM #<IO Tedit Stream/130,31000>
   START 0
   END 13285
   PROPS NIL
   TEXTOBJ {TEXTOBJ}#121,141600
   TRAILER (8863 52 3 187 1813772093 (&))
   PCCOUNT 187
   IDATE NIL
   PROPS NIL
   PC NIL
\TEDIT.GET.FORMATTED.FILE
   TEXTOBJ {TEXTOBJ}#121,141600
   PWINDOW NIL
   READONLY NIL
SI::*UNWIND-PROTECT*
   TEXT 
#<Input Stream on {MEDLEY}<lispusers>EQUATIONEXAMPLES.TEDIT;1/130,31700>
   TSTREAM #<IO Tedit Stream/130,31000>
   START 0
   END 13285
   PROPS NIL
   LISPXHIST ((&) (4 "" . "_ ") "<not yet evaluated>" 
NIL)
   SI::*RESETFORMS* NIL
   RESETSTATE NIL
\TEDIT.OPENTEXTSTREAM.PIECES
   TSTREAM #<IO Tedit Stream/130,31000>
   TEXTOBJ {TEXTOBJ}#121,141600
   TEDIT.GET.FINISHEDFORMS NIL
   PRIMPANE NIL
   START NIL
SI::*UNWIND-PROTECT*
   TEXT 
#<Input Stream on {MEDLEY}<lispusers>EQUATIONEXAMPLES.TEDIT;1/130,31700>
   WINDOW Tedit
   START/PROPS NIL
   END NIL
   PROPS NIL
   LISPXHIST ((&) (4 "" . "_ ") "<not yet evaluated>" 
NIL)
   SI::*RESETFORMS* NIL
   RESETSTATE NIL
OPENTEXTSTREAM
   TEXT {MEDLEY}lispusers>EQUATIONEXAMPLES.TEDIT
   WINDOW NIL
   DONTSPAWN NIL
   PROPS NIL
   TSTREAM NIL
   PROC NIL
TEDIT
   TEXT {MEDLEY}lispusers>EQUATIONEXAMPLES.TEDIT
   WINDOW NIL
   DONTSPAWN NIL
   PROPS NIL
TEDIT
   *FORM* (TEDIT (QUOTE 
{MEDLEY}lispusers>EQUATIONEXAMPLES.TEDIT))
   *ARGVAL* NIL
   *TAIL* NIL
   *FN* TEDIT
\EVALFORM
FAULTEVAL
   *FORM* (UNDOABLY (TEDIT &))
\EVALFORM
   \INTERNAL NIL
EVAL
EVAL-INPUT
   RETRYFLAG NIL
   HELPCLOCK 911
DO-EVENT
   SI::*DUMMY-FOR-CATCH* T
   SI::*CATCH-RETURN-FROM* (&)
   LISPXHIST ((&) (4 "" . "_ ") "<not yet evaluated>" 
NIL)
   HELPCLOCK 0
XCL::EXECA0001A0002
   *CURRENT-EVENT* ((&) (4 "" . "_ ") 
"<not yet evaluated>" NIL)
   SI::NLSETQ-VALUE NIL
   *PROCEED-CASES* (&)
   SI::*NLSETQFLAG* NIL
XCL::EXECA0001
\PROGV
   XCL::TOP-LEVEL-P T
   XCL::WINDOW {WINDOW}#161,156664
   XCL::TITLE-SUPPLIED NIL
   XCL::TITLE NIL
   *THIS-EXEC-COMMANDS* (#<Hash-Table @ 165,74632>)
   XCL::ENVIRONMENT NIL
   XCL::PROMPT NIL
   XCL::FN EVAL-INPUT
   XCL::PROFILE "XCL"
   *EXEC-ID* ""
   XCL::PROFILE-CACHE (XCL::*PROFILE-NAME* "IL" 
XCL:*EVAL-FUNCTION* EVAL *PACKAGE* #<Package INTERLISP> 
*READTABLE* #<ReadTable INTERLISP/172,64714> 
XCL:*EXEC-PROMPT* "_ " --)
EXEC
\PROC.REPEATEDLYEVALQT
   *FORM* (\PROC.REPEATEDLYEVALQT)
   *ARGVAL* NIL
   *TAIL* NIL
   *FN* \PROC.REPEATEDLYEVALQT
\EVALFORM
   %#FORM# (\PROC.REPEATEDLYEVALQT)
   *CURRENT-PROCESS* #<Process EXEC/172,22204>
   HELPFLAG BREAK!
   \CURRENTDISPLAYLINE 0
   \#DISPLAYLINES 22
   \LINEBUF.OFD #<IO Linebuffer Stream/171,141000>
   *READTABLE* #<ReadTable INTERLISP/172,64714>
   \PRIMTERMTABLE {TERMTABLEP}#172,57740
   \PRIMTERMSA {CHARTABLE}#172,60000
   TtyDisplayStream #<Output Display Stream/137,106100>
   SI::*RESETFORMS* NIL
   \INTERRUPTABLE T
   \TTYWINDOW NIL
   READBUF NIL
   \TERM.OFD #<Output Display Stream/166,120200>
   *STANDARD-OUTPUT* #<Output Display Stream/166,120200>
   *STANDARD-INPUT* #<IO Linebuffer Stream/171,141000>
\MAKE.PROCESS0
T
```

rmkaplan:
I have to upload the new fonts

nbriggs:
It might be a good idea, since you're converting to medleyfont format for Postscript fonts, and you're now converting to/from MCCS, to take a new look at "CONVERT-AFM-FILES" in POSTSCRIPTSTREAM.  It ignores many characters that are present in the font that are not part of the standard encoding.  There's `*POSTSCRIPT-EXTRA-CHARACTERS*` to pull some specials in, but that variable is NOT defined in the file, or anywhere else I can find.

You can find the Adobe Font Metric specs online pretty easily, and there are quite a few versions of the .AFM files out there (beware, while they may say version 2 of the AFM file format specification, there seem to be many versions of the metrics themselves -- e.g., Helvetica.afm has font versions 001.002 and 001.006 and they're quite different).  I have no idea what versions of the fonts the existing .PSCFONT files were created from -- they don't seem to maintain references back to the AFM source they were created from -- and neither of the above 2 versions of Helvetica.afm recreate the existing PSCFONT file for Helvetica.

pamoroso:
At commit 7426263 I still get the font not found error when opening TEdit files.

pamoroso:
Still font not found errors at commit 7c3179e.

rmkaplan:
For a particular font?  From a top-level call to FONTCREATE, or embedded somewhere?

Is it just postscript/pdf?  Are display fonts OK?

pamoroso:
I get the error for all display fonts, for example when just opening DInfo or a TEdit file.

pamoroso:
For context this is the font related code of my greetfile:

```
           (SETQ DISPLAYFONTDIRECTORIES (APPEND '(NIL "{DSK}<home>paolo>il>fonts>Commodore>" 
                                                      "{DSK}<home>paolo>il>fonts>ComicSans>" 
                                                      "{DSK}<home>paolo>il>fonts>")
                                               DISPLAYFONTDIRECTORIES))
```

rmkaplan:
Is there a difference between

(FONTCREATE '(CLASSIC 24 MRR 0 DISPLAY))

and

(FONTCREATE '(CLASSIC 24 MRR 0 POSTSCRIPT))

rmkaplan:
What if you take off the NIL ?

pamoroso:
The calls to `FONTCREATE`:

```
4_ (FONTCREATE '(CLASSIC 24 MRR 0 DISPLAY))
In ERROR:
FONT NOT FOUND
(CLASSIC 24 (MEDIUM REGULAR REGULAR) 0 DISPLAY)

5_: 
6_ (FONTCREATE '(CLASSIC 24 MRR 0 POSTSCRIPT))
{CLASSIC24-MRR/136,121314}
```

What do you mean by taking off the `NIL`?

nbriggs:
> What do you mean by taking off the NIL

Ron is referring to your DISPLAYFONTDIRECTORIES -- 
```
  (SETQ DISPLAYFONTDIRECTORIES (APPEND '(NIL "{DSK}<home>paolo>il>fonts>Commodore>" 
          "{DSK}<home>paolo>il>fonts>ComicSans>" 
          "{DSK}<home>paolo>il>fonts>")
        DISPLAYFONTDIRECTORIES))
```
He's suggesting that you remove the NIL that you're putting on the front of DISPLAYFONTDIRECTORIES and see if that's causing the font file search failure.

nbriggs:
I just noticed in the IRM, relevant to the way Postscript fonts are stored, this note:

> Warning:  One must be careful when using the function FONTSAVAILABLE to determine what Press font files are available.  For Press font families/faces, the font widths for different sizes are consistently scaled versions of the smallest font in the family/face.  Therefore, instead of storing data about all of the sizes in the FONTS.WIDTHS file, only the widths for the font of SIZE=1 are stored, and the other widths are calculated by scaling these widths up.  This is signified in the FONTS.WIDTHS file by a font with SIZE=0.  Therefore, if FONTSAVAILABLE is called with CHECKFILESTOO?=T, and it finds such a "relative" font, it returns a font spec list with size of 0.  For example, 
> _(FONTSAVAILABLE 'GACHA '* '* 0 'PRESS T)
> ((GACHA 0 (BOLD ITALIC REGULAR) 0 PRESS)
>  (GACHA 0 (BOLD REGULAR REGULAR) 0 PRESS)
>  (GACHA 0 (MEDIUM ITALIC REGULAR) 0 PRESS)
>  (GACHA 0 (MEDIUM REGULAR REGULAR) 0 PRESS))
> This indicates that Press files can be created with GACHA files of any size with faces BIR, BRR, MIR, and MRR.  Of course, this doesn't guarantee that these fonts are available in all sizes on your printer.
>

rmkaplan:
Yes, the code may be stumbling over NIL in this context, which I will look at if that's the problem.

But, what directory(ies) do you expect it to search over, when it sees NIL?

pamoroso:
Got it, here are the calls to `FONTCREATE` with `NIL` taken off:

```
5_ (FONTCREATE '(CLASSIC 24 MRR 0 DISPLAY))
In ERROR:
FONT NOT FOUND
(CLASSIC 24 (MEDIUM REGULAR REGULAR) 0 DISPLAY)

6_: 
7_ (FONTCREATE '(CLASSIC 24 MRR 0 POSTSCRIPT))
{CLASSIC24-MRR/136,121314}
```

rmkaplan:
On this revision I have made the change that the unscaled sources for scaling appear in the directory (and fontdescriptor) as size 0, not 1 (or the smallest as described here).  This now only applies to postscript/pdf, since we no longer support press, but the same logic would apply to scale fonts for other devices.

rmkaplan:
What happens in the display case if you revert to the original value of DISPLAYFONTDIRECTORIES, removing the APPEND?

nbriggs:
@rmkaplan - in a Venue full.sysout, NIL on DISPLAYFONTDIRECTORIES is treated as LOGINHOST/DIR -- though I don't know that that is a documented feature, and one might equally as well expect it to be the connected directory (though I think having it be LOGINHOST/DIR would lead to more reasonable and deterministic behavior)

pamoroso:
@rmkaplan Do you mean something like this?

```
(SETQ DISPLAYFONTDIRECTORIES '(NIL "{DSK}<home>paolo>il>fonts>Commodore>" 
                                   "{DSK}<home>paolo>il>fonts>ComicSans>" 
                                   "{DSK}<home>paolo>il>fonts>"))
```

pamoroso:
I went ahead and removed `(SETQ DISPLAYFONTDIRECTORIES ...)` altogether from the greetfile as it was only for testing early versions of READ-BDF:

```
5_ (FONTCREATE '(CLASSIC 24 MRR 0 DISPLAY))
{CLASSIC24-MRR/136,121314}
6_ (FONTCREATE '(CLASSIC 24 MRR 0 POSTSCRIPT))
{CLASSIC24-MRR/136,121252}
```

rmkaplan:
For filepkg operations NIL usually maps to DIRECTORIES.  But (DIRECTORYNAME NIL) is defined as the login host/dir and (DIRECTORYNAME T) is the connected directory.

I'll make sure that NIL is interpreted as the login--not sure it ever was in this context.

rmkaplan:
So, now, when it is clean, it works?

Do the Commodore and ComicSans subdirectories actually exist?  Maybe I'm missing a DIRECTORYNAMEP.

pamoroso:
Yes the subdirectories do exist.

I did some more testing with the default `DISPLAYFONTDIRECTORIES` and applications such as DInfo seem to work. However, printing lispusers/EQUATIONEXAMPLES.TEDIT to a Postscript file yields the error:

```
INTERLISP-ERROR
In ERROR:
FONT NOT FOUND
(MATH 8 (MEDIUM REGULAR REGULAR) 0 POSTSCRIPT)
```

rmkaplan:
Try the display again.  

I also discovered the MATH/SYMBOL problem, I have to look closer at the postscript DSPFONT function.  It may be looking in the wrong place now.

pamoroso:
At commit 604594f, after restoring the old `(SETQ DISPLAYFONTDIRECTORIES ...)` in the greetfile the call `(FONTCREATE '(CLASSIC 24 MRR 0 DISPLAY))` succeeds.

rmkaplan:
On SYMBOL, the problem was that it was asking for SYMBOL 10, but the coercion table didn't have an entry to map SYMBOL 10 to SYMBOL 0.  So I added (* (* 0)) to pick up everything.

There was no mapping for MATH for postscript, I'm not sure how that ever worked.  But I also added an entry (MATH CLASSIC).  EQUATION EXAMPLES now goes through for me, with a little difference in the spacing of the integration limits.

The remaining glitch (that I know of) is that the entry that I put into the MCCS to PS translation table doesn't work for the Circumflex. It shows up as up arrow.  But the leftarrow, uparrow, and underscore seem to work.

pamoroso:
At commit 028c81c printing EQUATIONEXAMPLES.TEDIT to a Postscript file completes successfully.

rmkaplan:
By the way,  PDF now works because PDFFONTCOERCIONS has a single entry (* (* * * * POSTSCRIPT)), which takes MATH-8-PDF to MATH-8-POSTSCRIPT, and POSTSCRIPTFONTCOERCIONS then takes it to CLASSIC-8-POSTSCRIPT which takes it to (NEWCENTURYSCHLBK-0-POSTSCRIPT), which is then scaled.   The fonts are relabeled on the way back, so it shows up as MATH-8-PDF with the right widths.

rmkaplan:
I uploaded the new Interpress Medleyformat fonts.  I've done some breath-of-life testing, and it doesn't break on EQUATIONEXAMPLES, but I don't know if it is doing the proper rendering.

pamoroso:
At commit 064fcd4 printing to Dodo's `POSEIDON:` Interpress printer does succeed but the summation, integral, and limit symbols are missing from the PDF rendering: [EQUATIONEXAMPLESip.pdf](https://github.com/user-attachments/files/31869334/EQUATIONEXAMPLESip.pdf)

rmkaplan:
Some of the layout is off too.  Was the layout off before?  Can you send me an image of the previous rendering?

rmkaplan:
Also, see what it looks like if you first (pop INTERPRESSFONTDIRECTORIES) to get rid of the MEDLEYFONT pathway.

pamoroso:
This is EQUATIONEXAMPLES.TEDIT in master which looks like a jumbled mess: [EQUATIONEXAMPLESip-master.pdf](https://github.com/user-attachments/files/31869595/EQUATIONEXAMPLESip-master.pdf)

And this is the file in rmk225--Postcript-fonts-and-other-font-improvements after popping `INTERPRESSFONTDIRECTORIES`: [EQUATIONEXAMPLESip-pop-INTERPRESSFONTDIRECTORIES.pdf](https://github.com/user-attachments/files/31869660/EQUATIONEXAMPLESip-pop-INTERPRESSFONTDIRECTORIES.pdf)

Back to the earlier rendering of EQUATIONEXAMPLES.TEDIT with math symbols replaced by a center dot, we saw [something similar](https://github.com/Interlisp/medley/pull/2703#issuecomment-5095697643) in PR #2703.

rmkaplan:
So, it seems we are doing better than master.

It may be that the fonts are correct in terms of metrics, but it also may be that they are marked with the wrong FONTCHARENCODING (maybe Altofont instead of MCCS).  Which would cause Tedit to try to compensate by on-the-fly recoding.

nbriggs:
I believe that the Dodo Postscript printing uses the centered dot as the slug character.  There's a comment somewhere in the Dodo printing that talks about which symbol/equation fonts are missing.  In the printer resource directory I see these PFB files that look as though they're for equation brackets:
```
EQBRAC12.PFB
EQBRAC20.PFB
EQBRAC28.PFB
EQBRAC36.PFB
EQLARGER.PFB
EQLARGE_.PFB
```
so sizes 12, 20, 28, 36 and ...

rmkaplan:
I vaguely recall that there was an earlier discussion, that the problem may be in Dodo, not in the interpress master we are creating.  Or, do you still think there is something to do on the Lisp side?

rmkaplan:
I wonder what Interpress shows for a character that doesn't exist (slug).  Does it show a black box, some slug character, or nothing?  A simple test:

(SETQ TSTREAM (TEXTSTREAM (TEDIT))
(PRINTCCODE 65 TSTREAM)
(PRINTCCODE 1000 TSTREAM)
(PRINTCCODE 66 TSTREAM)

and then send it to the Interpress printer.  What shows up between the A and B?

pamoroso:
The TEdit window after running your test code:

tedit-test

The Interpress output via Dodo's `POSEIDON:` printer: [ip-slug.pdf](https://github.com/user-attachments/files/31876844/ip-slug.pdf)

pamoroso:
As a reminder here are the [limitations of Dodo's Interpress printing](https://github.com/devhawala/dodo/blob/master/printsvc-configuration.md#postscript-generation).

rmkaplan:
This is consistent with the fact that the widths vector has zero for all of the missing/slug characters.  If the center dot is the way it represents the slugs, then the widths vector should have the width of that character instead of 0.    

Or it may be that a width of 0 is the way that slugs are signaled in the wd files (or as read by the wd reader), and that the outcharfn is doing the character substitution now but without getting the right width. 

We have added the FONTSLUGWIDTH as a field of the fontdescriptor, we should put the width of that character there, and make sure that the outcharfn knows to look there.

Does it make sense for any real character to have a width of 0, for Interpress or any other device? If not, then we could use width=0 as a generic indicator of the slug character.  The display would be changed to test that instead of checking the offset value.  

There would be no need for a separate slugchar bit vector, as I mentioned the other day.  But, what about diacritics and other non-spacing characters?  Do they naturally have 0 widths?  Do Interpress/postscript distinguish the imagewidth from the advance-width, as does the display?

rmkaplan:
I attempted to do the same slug-printing test on PDF/postscript, and it got an error.  Maybe a bug I introduced, or slugs were something that tedit/postscript never dealt.

rmkaplan:
I fixed the new postscript so it doesn't cause an error when it attempts to build a slug charset info.  But the behavior is still not reasonable.  Nothing shows up for a character in a character set with no characters, various different characters are substituted for an undefined character in an existing charset.  No consistent black box, or circle or...

pamoroso:
Nothing unusual to report at commit 9add89b.

rmkaplan:
With respect to NS/Interpress printing on Dodo, apart from the slug issue, we could get better text spacing if we made the Interpress Medley fonts not from the .WD files but from the widths of the postscript fonts that Dodo maps the XCCS characters into.

But is there any reason to spend more time on this?  It's been good way of validating that we can still get Interpress to work, and a good test of the new font interfaces, but do we really care about the output when we can go straight to pdf/postscript?

nbriggs:
The Postscript fonts that Dodo maps the Interpress fonts into *are* fonts provided by Xerox which should match the widths of Classic and Modern, assuming you have those fonts installed in the resources directory so that they get used.

I'd also point out that WE don't have correct widths for the Postscript fonts that we map onto (the AFM files we converted to PSCFONT files long ago are older versions of the fonts) and not only that, but if you use Ghostscript to convert from PS to PDF you don't even have the real Adobe fonts that we think we have the widths for-- Ghostscript doesn't ship with even the base 14 fonts, let alone the standard 35 (37?) Type 1 fonts that printers used to ship with.  Regular Adobe Postscript generating apps no longer (2023) support type 1 fonts - they've switched to TTF (TrueType) and/or OTF (OpenType) fonts.

rmkaplan:
If we don't have the correct widths for the PS fonts, and if pd2df uses a different set of fonts, isn't it surprising that our pdf files look as good as they do?

rmkaplan:
This is now a very big update, since it includes new versions of all the medley display, postscript, and interpress fonts, in addition to the code changes.  These fonts now systematically record the slug-status of every character, with some generic functions to set things up and interpret that status.   The postscript outcharfn now puts up a black box for the slug, the Interpress outcharfn still needs some work.  Maybe it should also do what postscript does.

pamoroso:
Still looking good at commit f1198be.

pamoroso:
Still looking good at commit 17c82bb.

rmkaplan:
@MattHeffron raised the question whether the justification would be off in postscript/pdf streams for lines that contain slug characters.  I did the test, the justification still works, no problem.

pamoroso:
Still looking good at commit 0a93310.

pamoroso:
Nothing unusual to report at commit c009f10.

rmkaplan:
Interpress wasn't showing slugs.  Without getting into BLTSHADE or choosing a specific character for the slug, I put in a hack that I think should just space over the intended width of the slug, showing white space.

@pamoroso , can you see what Dodo shows if you recreate the Tedit file that has a slug (e.g. charcode 1000) between an A and a B?  I think the B was right next to the A before, I hope now it will be spaced over a bit.

(To get the 1000, do (PRINTCCODE 1000 (TEXTSTREAM(WHICHW))) with the mouse over the Tedit window.)

nbriggs:
The *printer* puts in the slugs if the character is not present in its loaded fonts - you probably shouldn't second-guess what it's capable of doing.  That's how the fake font sampler works -- it's creating Interpress for a font whose name was changed from an existing font, but all the characters imaged get passed through even though we don't have the widths for the characters.

rmkaplan:
Do we know the width of the slug that the printer puts in?  That's what any formatter in Lisp needs to know?

pamoroso (approved):
Looks good to me.

PR #2746, opened 2026-09-05
Extend FILENAMEFIELD and PACKFILENAME with HOST-DIR composite field name

rmkaplan:
This adds a new composite field name HOST-DIR to FILENAMEFIELD and PACKFILENAME that makes it convenient to deal with the host/directory as a single entity.

(FILENAMEFIELD '{DSK}/foo/fum.xxx HOST-DIR) returns {DSK}/foo/, no need to retrieve HOST and DIRECTORY and  pack them together.

I've this hanging around for quite awhile, now I want to use it in the fontfile code.


pamoroso:
On Linux Mint 22.1 Cinnamon the change seems to work as expected.

pamoroso:
Nothing unusual to report at commit b01db6d.

masinter:
This change seems to interfere with the goal of moving file name packing into compatibility with Common Lisp pathname parsing and using the PATHNAME datatype holding the parsed components. 
Is that really no longer a goal?

rmkaplan:
I think it was more of a nice-to-have than a significant goal.  Nobody has signed up to do it, and it isn't so clear to me that the disruption would be worth the benefit.

But even if someone wanted to pursue it, I don't think it would change what FILENAMEFIELD returns.  It would only mean that there would be more cases where UNPACKFILENAME could return a result by the path branch that is currently there, without doing the more cumbersome string processing.

If you wanted to move forward on this, you could change PACKFILENAME to produce a path instead of a string or atom...and then see what breaks.

But I'm not sure it's worth the effort.

pamoroso (approved):
I tested it and it works.

PR #2751, opened 2026-09-10
Add inspector for 1-bit blocks

rmkaplan:
This adds an item to the pop-up menu for blocks that offers to inspect as a vector of single bits, not just 8, 16, ...

This allows, for example, the inspection of the 1-bit vectors that code the slug characters in a charset info.


pamoroso:
Medley builds and runs fine on Linux Mint 22.1 Cinnamon.

pamoroso:
How do I get a charset object I can inspect to access a slug? Calls such as `(CHARSET)` and `(CHARSET (OUTPUT (WHICHW)))` in an Exec return `NIL`.

nbriggs:
Probably the easiest way is to create a font object and then look at it with the inspector and walk the structure from there.

rmkaplan:
If you have EXPORTS.ALL loaded (or (LOADCOMP 'FONT)), then 

(\GETCHARSETINFO FONT CHARSET) will return the current CHARSETINFO for CHARSET (a number)).  NIL if that charset hasn't been instantiated.

(\INSURECHARSETINFO FONT CHARSET)  will make sure it is instantiated if it isn't already.

Or you can inspect the font's FONTCHARSETVECTOR, as Nick suggests.

But, hold off on this, I'm working on the construction of new font files and macros that make use of this.

The PR for INSPECT allows you to inspect the bit vector.  That should be easy to approve and merge.

pamoroso:
Okay thanks, I can now inspect `FONTCHARSETVECTOR`s and `CHARSETINFO`s.

pamoroso:
Nothing unusual to report at commit 551847c.

pamoroso (approved):
Looks good to me.

PR #2752, opened 2026-09-15
Rmk229  fix tedit heading keep

rmkaplan:
By Tedit standards, this is a small PR.

The main purpose is to fix the implementation of the heading-keep algorithm, and to remove a redundant and confusing field in the way the keep property was represented.

Included here is a new version of BITMAP-GALLERY.TEDIT, the file that first showed the problem.  Not only was there a problem in the implementation, but the heading-keep properties in this file were not as obviously intended.  The bitmap examples should be kept with the legend underneath, not the other way around.

This also includes a minor fix to HARDCOPY DISPLAY mode:  the code was not taking account of the line-select region of the window.


pamoroso:

pamoroso:
Still looking good at commit dcb0ead.

pamoroso (approved):
Looks good to me.

PR #2755, opened 2026-09-19
Rmk230  delete medley/fonts sources fonts that have moved to resources/fonts

rmkaplan:
These font source files have been copied to fonts/ in the Resources repository, now deleting them from medley/fonts/


pamoroso:
It works and runs with no issues on Linux Mint 22.1 Cinnamon.

rmkaplan:
I included an update to GITFNS whereby it creates a project and pseudohost for RESOURCES, if it finds a resources/ directory as a sibling of an ancestor of MEDLEYDIR.   So e.g. display font sources can be referenced by {RESOURCES}/fonts/display/.

(An alternative would be to create the pseudohost as part of the MEDLEYDIR set up, maybe based on a separate UNIX-GETENV.)

pamoroso:
`{RESOURCES}` works as expected at commit 36b184c.

masinter:
There might be a problem with having a pseudohost when the target has changed.
If you create the PSUEDOHOST in a system where the resources directory is there, and then save the sysout, and then restart the sysout when the resources directory isn't to be found, will it update the pseudohost? After every event? or even just modifying the file system in another process...

pamoroso (approved):
Looks good to me.

rmkaplan:
With respect to the pseudohost, it was decided at the meeting (9-21) to create the pseudohost for RESOURCES also at the time that the MEDLEY host is created, independent of GITFNS.  GITFNS would still make sure, as in this code, that the pseudohost is defined, as a consequence of defining RESOURCES as one of the default projects (like LOOPS, MAIKO, NOTECARDS).

Issue #2756, opened 2026-09-19
MAKEFILE-ENVIRONMENT property is outdated


rmkaplan:
There are bunch of files whose COMS explicitly establish a MAKEFILE-ENVIRONMENT property.  I think this was the convention from an earlier time, before DEFINE-FILE-INFO was set up to store the property in the file hashtable.

There is a danger of confusion if these two specifications get out of step (as happened when I tried to convert POSTSCRIPTSTREAM to UTF-8).

It should be an error to make a file that defines this property--the property should be removed.

PR #2757, opened 2026-09-19
Detect outdated MAKEFILE-ENVIRONMENT property

rmkaplan:
This puts a check in PRETTYDEF to cause an error if there is both a MAKEFILE-ENVIRONMENT property and an entry in the file environment hashtable (presumably from DEFINE-FILE-INFO).  CONVERT-TO-UTF8 also has a check.

This addresses issue #2756


pamoroso:
On Linux Mint 22.1 Cinnamon I copied POSTSCRIPTSTREAM from the library to my `~/il/`, called `(CONVERT-TO-UTF8 'POSTSCRIPTSTREAM)` from Medley, and got the expected error:

```
Help! Please remove the outdated MAKEFILE-ENVIRONMENT property
```

